PR #2, opened 2021-04-13
Add design doc for panics and errors.

paddycarver:
Add a design doc that proposes a set of conventions and trade offs
involved in deciding whether to use a panic or an error in the
framework. The goal of this document is to help us achieve consistency
on this, and get the most out of Go's tools around exceptional behavior.

This proposal can be summarised as "use whichever is most appropriate
for the situation you're in", but outlines a number of different
guidelines to use and things to consider when determining which is most
appropriate.

### DIFF
--- a/docs/design/panic-error.md
+++ b/docs/design/panic-error.md
@@ -215,7 +215,7 @@ with the `Type` specified. In the vast majority of circumstances, `v` will be
 of a type known at compile time, and therefore panics will show up if the code
 is executed even once in testing. But there are edge cases in which an API may
 return a value with an interface type, in which case, the type of `v` may not
-be known at compile time. In those edge cases, providers can call
+be known at compile time. In those edge cases, providers should call
 `tftypes.ValidateValue` before calling `tftypes.NewValue`, secure in the
 knowledge that if no error was returned from `tftypes.ValidateValue` they are
 safe from panics.
--- a/docs/design/panic-error.md
+++ b/docs/design/panic-error.md
@@ -123,7 +123,10 @@ provider.
 
 While it's easy to assert that provider developers should just check the error,
 the API should be designed such that it's hard to use incorrectly, and
-incorrect usage should be loud and yield noticeable feedback.
+incorrect usage should be loud and yield noticeable feedback. Linters can help
+with this, and we can encourage their use, but we should try to design APIs
+that don't lean too much on external tooling we can't assume provider
+developers will have or use.
 
 Panics are more conspicuous than errors, as they cannot be accidentally
 ignored. They default to being noticeable.
--- a/docs/design/panic-error.md
+++ b/docs/design/panic-error.md
@@ -157,6 +157,20 @@ tend to only convey the information available at the level of abstraction that
 prompted them. This suggests that panics lower in the abstraction should be
 considered more expensive, as they convey less context to the developer.
 
+## Interruptability
+
+Related to the ability to recontextualise is the consideration around halting.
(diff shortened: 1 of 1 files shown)


docs/design/panic-error.md
```diff
@@ -0,0 +1,231 @@
...
+invalid, if the value is an invalid type, or if the type of the value doesn't
+match the type expected at that path. This function is ubiquitous in provider
+development. Unfortunately, possibly because it's so ubiquitous, provider
+developers often forget to check the error value returned, leading to the
+exception not being surfaced or addressed, and unexpected behavior from the
+provider.
+
+While it's easy to assert that provider developers should just check the error,
+the API should be designed such that it's hard to use incorrectly, and
+incorrect usage should be loud and yield noticeable feedback.
```
kmoe, line 126: We can also strongly recommend (and include in terraform-provider-scaffolding) the use of a linter to ensure errors are always checked. It's not the perfect solution, but it's another option.

kmoe (commented):
This is great and I agree with almost everything here.

Two considerations:

#### Continuing execution versus interrupting execution

Panics halt execution, while errors allow cleanup by functions higher up the call stack. Terraform plugins are particularly concerned with graceful failure (if recovery is impossible), as they may be engaged in non-atomic API operations (https://github.com/hashicorp/terraform/issues/28271 is a recent example of a panic leading to a corrupted file, although in this case the panic was in Core). If an exception is likely to occur in a context in which cleanup will need to occur, it probably shouldn't be a panic.

A panic in the middle of a non-atomic API operation is different from a panic in the middle of constructing a gRPC response. Core can now give more informative errors about which gRPC response was expected when a plugin process crashed (https://github.com/hashicorp/terraform/pull/26694). Still, the practitioner and provider developer experience is poor due to lack of contextual information.

#### Naming

 There is a convention in Go to use `Must` to denote a version of a function that panics instead of erroring: for example, `regexp.MustCompile`, the [`Must`](https://golang.org/pkg/text/template/#Must) wrapper in `text/template`, and the [`Must`](https://pkg.go.dev/github.com/hashicorp/go-version#Must) helper in `hashicorp/go-version`. I think this is well-known enough to consider recommending here, and avoids further overloading the term "validation".

docs/design/panic-error.md
paddycarver: Think I addressed this in the text.

paddycarver:
I updated the text to add a section on halting program execution, which I think addresses your first point. On your second point, I think we should discuss specifics of the naming on each implementation of the pattern, as I think it depends on the context for what we do. For example, if we want the default behavior most people do to be returning errors and checking them, I agree with you that the `Must` pattern is one we should lean on. But if we want the default behavior most people do to be panicking, and checking the error is only really relevant for a small portion of the users, I think it makes sense to make the "unmodified" name (the name the user would be most likely to reach for first) the one that panics, with the one that returns an error getting some modifier (doesn't need to be Validate, I think we have several options we could pick from there.)

These are all really good points though, and I like that we're having this discussion.

Any more thoughts?

kmoe (approved):
The section on interruptability looks good. The sort of situation I was referring to with "non-atomic API operations" is one where the provider is making multiple API calls in between manipulating values, and may need to make further API calls within the same resource CRUD function depending on the errors it gets. The vsphere provider has some nice examples (e.g. in `resource_vsphere_vmfs_datastore.go`) of complex logic like this.

I like your distinction as regards the naming: `Must` is used when the default is _not_ to panic. If it's possible for a function to panic, this should always be mentioned in the GoDoc regardless of naming.

PR #3, opened 2021-04-13
create structs/interfaces decision doc

kmoe:
This document summarises the tradeoffs in the decision between representing Terraform providers, resources, and data sources as structs, or interfaces.


docs/decision/structs-interfaces.md
```diff
@@ -0,0 +1,231 @@
...
+
+On the other hand, framework developers can add additional methods to an existing struct and document the old ones as deprecated, without forcing provider developers to use a specific combination of old and new methods. 
+
+> This was the approach taken in `helper/schema` when two new sets of CRUD functions were added: `CreateContext` etc ([#276](https://github.com/hashicorp/terraform-plugin-sdk/pull/276)), and `CreateWithoutTimeout` etc ([#723](https://github.com/hashicorp/terraform-plugin-sdk/pull/723)). 
+ 
+As is evident from the `helper/schema` case, such new methods must have unique and therefore rather verbose names within the existing `Resource` struct, putting the struct approach at a **slight disadvantage** relative to interfaces.
+
+#### Interfaces
+
+Compile-time assertions can be made in provider code that a struct instance satisfies a particular interface. For example, the following resource satisfies `ResourceFactory`:
```
paddycarver, line 199: I'm not seeing `ResourceFactory` defined anywhere, and it looks like it's not being asserted against, which I think is what you're talking about here? Having trouble connecting this example to the description.

Oh, you mean that it will give a compile-time error if `computeInstanceResource` doesn't fulfill `framework.Resource`? Should we call that out explicitly? A lot of code going on there.

paddycarver (commented):
This looks awesome, thanks for writing it up. I left some comments inline.

I think the only other things I can think of that come into play here are:

* ability to store SDK-controlled information (TerraformVersion, etc.)
* mutability (Structs can technically be mutated at runtime; interfaces cannot.)
* Compiler support for typos (structs will error about Craete, interfaces will just silently ignore that method, possibly erroring at runtime if that method is required, silently ignoring it if that method is optional)
* what deprecation looks like in these cases (we'll need to maintain this package for a while, we should take into account whether it helps or hinders us in communicating with users)

docs/decision/structs-interfaces.md
```diff
@@ -0,0 +1,231 @@
...
+
+## Tradeoffs
+
+### State
+
+#### Structs
+
+If providers, resources, and data sources are represented as structs in framework code, provider code will instantiate  values of those structs. It is not possible to add fields to these structs in provider code for the purposes of storing arbitrary strongly-typed private data that could serve as, for example, the internal state of a resource. This lack of flexibility is a **disadvantage**, although it likely also prevents some anti-patterns.
+
+Certain struct fields could be populated by the framework: suggestions for fields on `Resource` have included `Config` for the resource configuration, and possibly even `State`. This approach seems strictly worse than including config and state values in CRUD function request and response objects (e.g. `ResourceCreateRequest`), since it would be unclear when and how the framework modified those values.
```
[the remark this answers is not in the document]
kmoe: Would `ProviderMeta` need to have type `interface{}`?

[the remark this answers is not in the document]
kmoe: `TerraformVersion` also seems to come under SDK-controlled information. Does this still all fit into a "runtime state" section?

docs/decision/structs-interfaces.md
kmoe: Thanks, I'd accidentally mixed two examples!

kmoe:
Thanks for the review. I've expanded the doc with your suggestions. I didn't include a discussion of mutability at first as I wanted to focus on practical consequences for provider developers. Let me know if you think we should take that section in another direction.

docs/decision/structs-interfaces.md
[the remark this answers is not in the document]
paddycarver: ProviderMeta currently has a cty.Value type, and is analogous to provider configuration values, but is specified in resource RPC calls, not in a separate RPC call.

Runtime state may be a good description of it, yeah.

paddycarver:
I think in the rename, you may have lost the actual name of the file here, so this no longer really renders as Markdown.

paddycarver:
Renamed just to fix rendering so review can move forward.

docs/decision/structs-interfaces.md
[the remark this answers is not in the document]
kmoe: I'd like to deal more with `ProviderMeta` and other `meta` concepts in the upcoming runtime information design doc, which will be cross-referenced to this one.

kmoe:
Thanks, @paddycarver - squashed everything together now for easy review.

docs/design/structs-interfaces.md
```diff
@@ -0,0 +1,251 @@
...
+Certain struct fields could be populated by the framework: for `Resource`, this could include `ProviderMeta` and `TerraformVersion`. Provider code cannot add additional struct fields to store strongly-typed data. This lack of flexibility is a **disadvantage**, although it likely also prevents some anti-patterns.
+
+An abandoned idea involved storing `Config` and `State` on the resource struct. This approach seems strictly worse than including config and state values in CRUD function request and response objects (e.g. `ResourceCreateRequest`), since it would be unclear when and how the framework modified those values.
+
+
+#### Interfaces
+
+In fulfilling an interface by creating instances of their own provider, resource, and data source struct types, provider developers have the **advantage** of more flexibility in their choice of fields within those structs, and whether they are used to store their own state.
+
+Allowing provider developers to define their own resource types, and construct them with a factory pattern, means that the type assertion on any `ProviderMeta` parameter need be done only once, in the factory itself. This is an **advantage**.
```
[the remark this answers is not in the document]
paddycarver: Similarly, it's mentioned in the Structs section above. I think this is likely a result of confusing/inaccurate feedback I gave in a previous review or in a conversation.

docs/design/structs-interfaces.md
```diff
@@ -0,0 +1,251 @@
...
+#### Structs
+
+If providers, resources, and data sources are represented as structs in framework code, provider code will instantiate values of those structs. 
+
+Certain struct fields could be populated by the framework: for `Resource`, this could include `ProviderMeta` and `TerraformVersion`. Provider code cannot add additional struct fields to store strongly-typed data. This lack of flexibility is a **disadvantage**, although it likely also prevents some anti-patterns.
+
+An abandoned idea involved storing `Config` and `State` on the resource struct. This approach seems strictly worse than including config and state values in CRUD function request and response objects (e.g. `ResourceCreateRequest`), since it would be unclear when and how the framework modified those values.
+
+
+#### Interfaces
```
paddycarver, line 121: One bit I think is missing from here is an acknowledgement of the disadvantage: data is strongly-typed, but that comes at the cost of needing to be _explicit_. Every bit of runtime state the framework wants to track needs to be explicitly communicated to these types, for example, `TerraformVersion`.

This can be set when the struct is created, by passing arguments to the factory:

```go
func NewComputeInstance(data framework.RuntimeState) Resource {
  return computeInstance{
    terraformVersion: data.TerraformVersion,
    client: data.ConfigureProviderData.(*Client),
  }
}
```

or it can be set with explicit methods:

```go
type computeInstanceResource struct {
  terraformVersion string
  client *Client
}

func (c *computeInstanceResource) SetTerraformVersion(version string) {
  c.terraformVersion = version
}

func (c *computeInstanceResource) SetConfigureProviderData(data framework.ConfigureProviderData) {
  c.client = data.(*Client)
}
```

It's also worth noting that helpers controlled by the framework may want access to this information (for example, tfhttp is the main audience of the TerraformVersion information, to include it in the User-Agent). This implies the necessity of a mechanism for the framework to retrieve this information. On framework-controlled structs, this is easy--it just accesses the property. On provider-controlled implementations of an interface, it needs to be accounted for with something like a getter:

```go
type computeInstanceResource struct {
  terraformVersion string
}

func (c computeInstanceResource) GetTerraformVersion() string {
  return c.terraformVersion
}
```

It may be worth calling out, though, that we could provide a framework-controlled struct providers are required to embed in their interface implementations (we can enforce this at compile time by requiring an unexported method to be implemented) that would hide this from provider developers entirely:

```go
type computeInstanceResource
  framework.RuntimeData
}
```

that would be implemented in the framework as

```go
type RuntimeData struct {
  terraformVersion string
  ConfigureProviderData interface{}
}

func (r ResourceData) getRuntimeData() framework.RuntimeData {
  return r
}
```

That would let us get access to `RuntimeData` from an `Resource` interface fulfilled by `computeInstanceResource` by just calling `getRuntimeData()`, without the provider developer needing to know anything about it, implement any methods, or do anything other than embedding the `RuntimeData` struct in their type. It would also let us hide data we wouldn't want to expose to provider developers while maintaining access to it for the framework, which might include TerraformVersion.

docs/design/structs-interfaces.md
```diff
@@ -0,0 +1,251 @@
...
+On the other hand, a compile-time error will occur if a provider developer has made a typo and defined, for example, a `Craete` method instead of `Create`. This is an **advantage**.
+
+#### Interfaces
+
+Compile-time assertions can be made in provider code that a struct instance satisfies a particular interface. For example, the following resource must satisfy `framework.Resource` and the compiler will error if any method name has a typo:
+
+```go
+type computeInstanceResource struct {
+  framework.Resource
+}
```
paddycarver, line 200: This would actually embed a `framework.Resource` type inside of `computeInstanceResource`. It would never cause a compilation error, as long as `framework.Resource` is a type that exists.

```suggestion
type computeInstanceResource struct {
}

var _ framework.Resource = computeInstanceResource{}
```

paddycarver (commented):
This is looking awesome! I left some comments and some suggestions for things we may want to call out. The general crux of the feedback is that the runtime data stuff can get confusing (really sorry, I'm like 95% sure at least some of this is me just being casually wrong, plausibly in previous review passes of this very PR) and there are some additional considerations to interfaces and runtime data that we should probably acknowledge.

docs/design/structs-interfaces.md
kmoe: Yes, it does make sense to have this included here as well as in the runtime state doc - we can remove any excessive duplication later.

paddycarver: My mental model is that this document is about the tradeoff of structs and interfaces as providing structure to the framework, and the runtime data document is about the different ways we can inject that runtime data and what runtime data we want to inject, and what the tradeoffs of those options are.

paddycarver: I think there's a lot of overlap here, but I think there's some context that can be teased out on either end of this, potentially with some overlap.

kmoe: It seems a bit bizarre to have `TerraformVersion` be the concern of a `Resource`, since it doesn't vary from resource to resource within the same provider. Is there any other runtime state the resource could be concerned with which the framework will need to access?

I think the `TerraformVersion` example, but with `Provider` instead of `Resource`, can be instructive here in showing what would need to happen in provider code if the framework needed access to runtime data stored on a provider-defined `Provider` struct (i.e. something fulfilling a framework-defined `Provider` interface... speaking of confusing naming). I'll write this up. 

In your final code example, what is `ResourceData`? Do you mean `Resource`?

docs/design/structs-interfaces.md
```diff
@@ -0,0 +1,251 @@
...
+```
+
+## Tradeoffs
+
+### Runtime state
+
+This section concerns stateful data generated during the lifecycle of the provider, not Terraform state.
+
+For example, consider the `meta` parameter in the CRUD functions of [`helper/schema#Resource`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema#Resource), which is conventionally used by providers to store a configured API client. Since it is sent to the provider via the `ConfigureProvider` RPC call without a way for the provider to specify a schema for this data, it cannot be strongly typed and is represented as `interface{}`. Its type must be asserted at the top of each CRUD method where it is used.
```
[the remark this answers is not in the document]
kmoe: It seems that every way we've worded this so far has been confusing! I've tried again to resolve it with your very helpful _framework-owned vs provider-owned_ distinction.

docs/design/structs-interfaces.md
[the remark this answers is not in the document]
kmoe: Thanks!

docs/design/structs-interfaces.md
paddycarver: > It seems a bit bizarre to have `TerraformVersion` be the concern of a Resource, since it doesn't vary from resource to resource (instance) within the same provider. Is there any other runtime state the resource could be concerned with which the framework will need to access?

Agreed, but this is necessary because it [comes to us in the ConfigureProvider request](https://github.com/hashicorp/terraform-plugin-go/blob/e0e351efc90e60fa583afa6a2a2e3d8231f25e24/tfprotov6/provider.go#L140) and isn't anywhere in the [ApplyResourceChange request](https://github.com/hashicorp/terraform-plugin-go/blob/e0e351efc90e60fa583afa6a2a2e3d8231f25e24/tfprotov6/resource.go#L308-L359) meaning that if we want to include it in the User-Agent (which is why it exists), the framework is going to have to keep track of that state between the ConfigureProvider and ApplyResourceChange RPCs.

I think at the moment, that's the only framework-owned state. The API client produced by the ConfigureProvider RPC is the canonical example, but that's of a provider-defined type. But I think over time we'll continue to see information added that's framework-owned state (for example, I have proposals in mind to surface the provider address and version through ConfigureProvider RPC, etc.).

Whether that data gets embedded in a user-defined Provider type or a user-defined Resource type I think is an implementation detail that should be explored in the runtime data design doc. But the implementation doesn't materially change the way it interacts with the structs/interfaces decision being presented here.

> In your final code example, what is `ResourceData`? Do you mean `Resource`?

This is what I get for typing fast in GitHub comments. :D I mean `RuntimeData`, and have fixed it.

paddycarver: I think the crux of what I'm talking about here is "how does this choice impact how we smuggle data from the ConfigureProvider RPC into the CRUD functions?" and "how does this choice impact how we can automatically enrich helpers with information from the ConfigureProvider RPC, even when the helper is called within CRUD functions?"

kmoe: Thanks, this is an extremely helpful framing and I've aimed to summarise the core points in the expanded runtime state section.

docs/design/structs-interfaces.md
```diff
@@ -0,0 +1,304 @@
...
+```
+
+Provider code:
+```go
+type provider {
+  framework.RuntimeData
+}
+```
+
+`ConfigureProviderData` here represents other runtime data supplied by the `ConfigureProvider` request. Please see the document on Runtime State for further discussion of what runtime data may be injected and the tradeoffs involved.
```
paddycarver, line 176: (I don't think this needs to be changed, but I just want to be clear about what I was intending: `ConfigureProviderData` in my example is a 1:1 mapping to `meta` in SDKv2. I think this comment should only result in a change if it changes your understanding of this paragraph. I think the paragraph is fine and accurate as-is, however.)

docs/design/structs-interfaces.md
```diff
@@ -0,0 +1,304 @@
...
+
+This section concerns stateful data generated during the lifecycle of the provider, not Terraform state.
+
+For example, consider the `meta` parameter in the CRUD functions of [`helper/schema#Resource`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema#Resource), which is conventionally used by providers to store a configured API client. Since this data lives in a framework-owned type and not a provider-owned type, its type cannot be known at compile time and it is represented as `interface{}`. Its type must be asserted at the top of each CRUD method where it is used.
+
+#### Structs
+
+If providers, resources, and data sources are represented as structs in framework code, provider code will instantiate values of those structs. 
+
+Certain struct fields could be populated by the framework: for `Resource`, this could include `ProviderMeta` and `TerraformVersion`. Provider code cannot add additional struct fields to store strongly-typed data. This lack of flexibility is a **disadvantage**, although it likely also prevents some anti-patterns.
```
paddycarver, line 114:
```suggestion
Certain struct fields could be populated by the framework: for `Resource`, this could include the equivalent of SDKv2's `meta` value and `TerraformVersion`. Provider code cannot add additional struct fields to store strongly-typed data. This lack of flexibility is a **disadvantage**, although it likely also prevents some anti-patterns.
```

Just a suggestion. I think talking about `ProviderMeta` was a mistake on my part and I needlessly complicated everything, sorry.

paddycarver (approved):
This is really solid, comprehensive, and compelling. Really awesome work. Thanks for the patience through all the back and forth, and sorry for confusing matters.

PR #4, opened 2021-04-21
create provider-level runtime state doc

kmoe:


docs/design/runtime-state.md
```diff
@@ -0,0 +1,157 @@
...
+An extra `RuntimeData` parameter in resource CRUD functions may make them harder to unit test.
+
+### Option 1b: Resource or Provider type
+
+`RuntimeData` could be included in the `Resource` or `Provider` type, so that it could be accessed from within provider `Resource` methods (assuming the resource instance contains a reference to the provider instance if data is stored there). 
+
+As detailed in the [Structs and Interfaces](./structs-interfaces.md) design document, this approach would require additional tradeoffs if resources/providers are defined as interfaces in the framework: provider developers may need to implement getters on their resource/provider struct types, or at the very least embed `framework.RuntimeData` in their structs, with the following provider code:
+
+```go
+type provider {
```
paddycarver, line 114:
```suggestion
type provider struct {
```

docs/design/runtime-state.md
```diff
@@ -0,0 +1,157 @@
...
+```go
+type provider struct {
+  Client *github.Client
+}
+```
+
+The provider code would need to set this field in the configure func:
+
+```go
+func (p Provider) Configure(ctx context.Context, req *tfprotov5.ConfigureProviderRequest, res *tfprotov5.ConfigureProviderResponse) {
```
paddycarver, line 134:
```suggestion
func (p *Provider) Configure(ctx context.Context, req *tfprotov5.ConfigureProviderRequest, res *tfprotov5.ConfigureProviderResponse) {
```

docs/design/runtime-state.md
```diff
@@ -0,0 +1,157 @@
...
+func (r myResource) Create(ctx context.Context, req framework.ResourceCreateRequest, resp framework.ResourceCreateResponse) {
+  // determine correct parameters "params" for API request from config and state
+  
+  r.Provider.Client.CreateResource(params)
+}
+```
+
+The signatures of `Configure` and `Create` functions here are illustrative examples.
+
+This approach has the clear advantage, over Option 1, that the provider API client is now strongly typed. No type assertions are necessary, and it is clear at compile time prior to any resource implementations whether `provider.Client` is of the correct type.
```
paddycarver, line 155: The only trade-off perhaps worth mentioning for this is that it may (depending on how `Provider` is surfaced to resources) require providers to use mutexes to write concurrency-safe code, or for us to do some mutex trickery ourselves (e.g., creating a RWMutex and locking it before ConfigureProvider, unlocking it after, and read-locking/unlocking it before/after every CRUD call).

kmoe: Agree, great call.

kmoe:
Addressed all review comments, squashed. Ready for re-review.

paddycarver (approved):
This looks great. Awesome job on breaking the problem down and bringing clarity.

PR #5, opened 2021-04-22
Add design doc about the factory problem.

paddycarver:
We want to instantiate new instances of resources, data sources, and
providers so that we can have some isolation at runtime, but either the
provider defined the type of the resource/data source/provider and we
don't know how to create new instance of it, or the framework defined
the type and the provider needs to supply values for it, which the
framework needs to get from the provider when creating a new instance of
it.

This document explores different ways to achieve that.

### DIFF
--- a/docs/design/factories.md
+++ b/docs/design/factories.md
@@ -134,7 +134,7 @@ type ResourceFactory interface {
 ```
 
 Which can then be used instead of the type the factory will return when asking
-users for resources, data sources, and providers:
+consumers for resources, data sources, and providers:
 
 ```go
 type Provider struct {
--- a/docs/design/factories.md
+++ b/docs/design/factories.md
@@ -155,7 +155,7 @@ would return an interface) or by the framework.
 
 ### Documentation
 
-[sdkv2-provider-func][https://github.com/hashicorp/terraform-plugin-sdk/blob/893e7238350e1980eb2cce3303689ba59ae47490/plugin/serve.go#L28]
-[sdkv2-resource-func-call][https://github.com/hashicorp/terraform-provider-scaffolding/blob/243ba4948171e3902003f678c7c43ec3fafcdc20/internal/provider/provider.go#L33]
-[sdkv2-resource-schema][https://github.com/hashicorp/terraform-provider-scaffolding/blob/243ba4948171e3902003f678c7c43ec3fafcdc20/internal/provider/resource_scaffolding.go#L10-L29]
-[structs-interfaces][https://github.com/hashicorp/terraform-plugin-framework/blob/main/docs/design/structs-interfaces.md]
+[sdkv2-provider-func]: https://github.com/hashicorp/terraform-plugin-sdk/blob/893e7238350e1980eb2cce3303689ba59ae47490/plugin/serve.go#L28
+[sdkv2-resource-func-call]: https://github.com/hashicorp/terraform-provider-scaffolding/blob/243ba4948171e3902003f678c7c43ec3fafcdc20/internal/provider/provider.go#L33
+[sdkv2-resource-schema]: https://github.com/hashicorp/terraform-provider-scaffolding/blob/243ba4948171e3902003f678c7c43ec3fafcdc20/internal/provider/resource_scaffolding.go#L10-L29
+[structs-interfaces]: https://github.com/hashicorp/terraform-plugin-framework/blob/main/docs/design/structs-interfaces.md
--- a/docs/design/factories.md
+++ b/docs/design/factories.md
@@ -125,7 +125,7 @@ provider would need to use reflection to achieve this outcome.
 ### Factory types
 
 Instead of using functions, the framework can define an interface for a factory that
-providers define:
+can be implemented by a provider-defined type:
 
 ```go
 type ResourceFactory interface {
--- a/docs/design/factories.md
+++ b/docs/design/factories.md
@@ -144,7 +144,7 @@ type Provider struct {
 }
 ```
 
-This works whether the type is defined by the provider (the `NewResource()`
+This works whether the `Resource` type is defined by the provider (the `NewResource()`
 method would return an interface) or by the framework.
 
 ### Separating types and values
--- a/docs/design/factories.md
+++ b/docs/design/factories.md
@@ -42,7 +42,7 @@ would return an interface) or by the framework.
 
 ### Named functions
 
-Similarly to anonymous functions, we can ask for named functions:
+Similarly to anonymous functions, we can ask for functions with a named type:
 
 ```go
 type ResourceFactory func() framework.Resource
--- a/docs/design/factories.md
+++ b/docs/design/factories.md
@@ -284,6 +284,17 @@ Reflection, manual copying, and anonymous function types can only have their
 intent inferred by the name of the property they're set on, which is the
 hardest to build automation around.
 
+### Schema Access
+
+Anonymous functions, named functions, factory types, reflection, and manually
+copying all let the provider developer access the resource's schema inside the
+CRUD functions of that resource, by just calling the method.
+
+If resource types and resource instances are separated into two Go types,
+however, this avenue will no longer be available, and we'll either need to
+surface the resource type or its schema in the CRUD functions somehow, or
+decide that provider developers don't need access to it.
+
 [sdkv2-provider-func]: https://github.com/hashicorp/terraform-plugin-sdk/blob/893e7238350e1980eb2cce3303689ba59ae47490/plugin/serve.go#L28
 [sdkv2-resource-func-call]: https://github.com/hashicorp/terraform-provider-scaffolding/blob/243ba4948171e3902003f678c7c43ec3fafcdc20/internal/provider/provider.go#L33
 [sdkv2-resource-schema]: https://github.com/hashicorp/terraform-provider-scaffolding/blob/243ba4948171e3902003f678c7c43ec3fafcdc20/internal/provider/resource_scaffolding.go#L10-L29
--- a/docs/design/factories.md
+++ b/docs/design/factories.md
@@ -144,8 +144,8 @@ type Provider struct {
 }
 ```
 
-This works whether the `Resource` type is defined by the provider (the `NewResource()`
-method would return an interface) or by the framework.
+This works whether the `Resource` type is defined by the provider (the
+`NewResource()` method would return an interface) or by the framework.
 
 ### Separating types and values
 
--- a/docs/design/factories.md
+++ b/docs/design/factories.md
@@ -5,10 +5,11 @@ and resources that are scoped to the RPC server; you can see this in the SDKv2
 [`ProviderFunc`][sdkv2-provider-func] type, which is used to instantiate a
 provider when the server starts up. But you can also see it in the practice of
 [using a function][sdkv2-resource-func-call] to [register resource
-schemas][sdkv2-resource-schema], even though there's no enforced requirement
-that providers do this. This keeps the variables representing providers, data
-sources, and resources scoped to the gRPC server, which is very helpful when
-running multiple servers at the same time, as in the acceptance test drivers.
+schemas][sdkv2-resource-schema-usage], even though there's no enforced
+requirement that providers do this. This keeps the variables representing
+providers, data sources, and resources scoped to the gRPC server, which is very
+helpful when running multiple servers at the same time, as in the acceptance
+test drivers.
 
 There are two main goals when instantiating new instances of a type: to provide
 isolation from other instances of that type that may be running (otherwise, a
@@ -149,21 +150,76 @@ This works whether the `Resource` type is defined by the provider (the
 
 ### Separating resource types and resource instances
 
-The SDK currently conflates two separate ideas: the type of a resource--like
-"random_pet", the resource's name and schema, not the Go type--and a specific
-instance of that resource type--like "random_pet.my_resource", a set of
-concrete values filled into the resource type, a single state entry.
+Resources are currently treated as a single logical concept: they have a
+schema, and they have CRUD functions, and the same type is used to implement
+both of these.
 
-Rather than conflating resource types and instances, we can separate them out
-into two different Go implementations. The resource type can then serve as a
-factory that contains information common to all instances, and instances can
-surface the implementations that are used to operate only on instances of a
-resource type:
+There are, however, two underlying concepts that are surfaced as "resources"
+right now: resource types and resource instances.
+
+Resource types are the resource in abstract form. `random_pet` is a resource
+type. It has a schema, but no config, state, or plan. It doesn't show up in a
+practitioner's configuration files at all. It has no lifecycle.
+
+Resource instances are the resource in concrete form. `random_pet.my_pet` is a
+resource instance. It has a schema, but also has a config, state, and plan. It
+shows up in the pracitioner's configuration files. It has a lifecycle.
+
+At the moment, both of these concepts are surfaced as a single `Resource` type.
+This leads to two problems:
+
+First, `helper/schema` uses [a single instance][sdkv2-resource-registration] of
+the `Resource` type for all RPC calls. _If_ we use a provider-defined type for
+resources, this may lead providers to try and store information generated
+during RPC calls in their `framework.Resource` implementation:
+
+```go
+type myResource struct {
+  readResult tftypes.Value
+}
+
+func (m *myResource) Read(ctx context.Context, req framework.ReadResourceRequest, resp framework.ReadResourceResponse) {
+  // fetch state from the API here
+  // this next line assumes the state from the API is "hello, world"
+  // this is unlikely, but sufficient to illustrate the point
+  m.readResult = tftypes.NewValue(tftypes.String, "hello, world")
+}
+
+func (m *myResource) Create(ctx context.Context, req framework.CreateResourceRequest, resp framework.CreateResourceResponse) {
+  var readResult string
+  err := m.readResult.As(&readResult)
+  if err != nil {
+    panic(err)
+  }
+  // make an API call here using readResult
+}
+```
+
+This code could _sometimes_ work if we're not careful about always generating a
+new `myResource` for each RPC call we handle. But if the inner workings of the
+SDK or Terraform's graph change in any way, it's likely to break this code,
+which may not be obvious to provider developers.
+
+Second, there exists a certain kind of state that it's very reasonable for
+providers to want to have available to all RPC calls for every instance of
+their resources. This state usually is _used_ by RPC calls, not _created_ by
+it. An example of this we see a lot in the wild is a mutex that constrains the
+number of requests that can be made in parallel, to not provoke API rate
+limiting. Currently, the only way to keep this state is to register it at as
+global mutable state. This is problematic in testing scenarios, as all provider
+servers will need to share that same state; it's not just one server's resource
+instances, it's all the resource instances for all the servers created by any
+concurrently-running tests.
+
+We have the option of surfacing this distinction explicitly, allowing providers
+to store this state that should be shared among all instances of a resource
+type. We could define the resource type and the resource instance as separate
+Go types:
 
 ```go
 type ResourceType interface {
   GetSchema() *tfprotov5.Schema
-  NewValue() Resource
+  NewResource() Resource
 }
 
 type Resource interface {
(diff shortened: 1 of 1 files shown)


docs/design/factories.md
```diff
@@ -0,0 +1,234 @@
...
+newResource.Read = resource.Read
+newResource.Update = resource.Update
+newResource.Delete = resource.Delete
+```
+
+we can use this in a helper, to do something like:
+
+```go
+type Provider struct {
+	Resources map[string]Resource
```
kmoe, line 116:
```suggestion
  Resources map[string]Resource
```

docs/design/factories.md
```diff
@@ -0,0 +1,234 @@
...
+its own instances of that type, and would be able to copy over the data used to
+populate them:
+
+```go
+resource := framework.Resource{
+	Schema: map[string]*Schema{},
+	Create: createFunc,
+	Read: readFunc,
+	Update: updateFunc,
+	Delete: deleteFunc,
```
kmoe, line 101:
```suggestion
  Schema: map[string]*Schema{},
  Create: createFunc,
  Read: readFunc,
  Update: updateFunc,
  Delete: deleteFunc,
```

docs/design/factories.md
```diff
@@ -0,0 +1,234 @@
...
+  Resources map[string]ResourceFactory
+}
+```
+
+This works whether the type is defined by the provider (the `NewResource()`
+method would return an interface) or by the framework.
+
+### Separating types and values
+
+The SDK currently conflates two separate ideas: the type of a resource--i.e.,
```
kmoe, line 150: I'm not sure I understand the conflation in question here. The SDK currently defines a `schema.Resource` type, which is a struct:
```go
type Resource struct {
  Schema map[string]*Schema
  // all our familiar resource fields
}
```

This is a type. 

In provider code, instances of this type are created:
```go
myResource := &schema.Resource{
  Schema : ...
  // etc.
}
```

The SDK has no trouble distinguishing these two things, because one is a type and one is a value, just as the `http.Client` type is different from `&http.Client{}`.

When the SDK needs to talk about _a single resource definition in state_, it has its own types for this: `terraform.ResourceState` and `terraform.InstanceState`. As far as I can tell, provider code is not concerned with these, because the job of the provider server is to:
1. Receive an RPC specifying a certain `type_name` of resource (e.g. `compute_instance`) and the config and state for that resource instance;
2. Perform API requests as necessary to make sure the state of the world matches the Terraform state;
3. Respond to the RPC with the updated state values.

I haven't found any examples of idiomatic Go that look like the `ResourceType` and `ResourceValue` constructs below, and it's unclear to me how they map to the problem of separating _the general idea of a resource_ from _a single resource definition in state_. A resource's `Create` method is always the same regardless of which instance of the resource we're talking about - that information comes in via the _parameters_ of the `Create` function.

docs/design/factories.md
```diff
@@ -0,0 +1,234 @@
...
+There are a few ways these values can be instantiated, and some of the details
+are dependent on whether the types in question are [framework-controlled or
+provider-controlled][structs-interfaces].
+
+This document is meant to catalogue our available options and explore their
+benefits and trade-offs.
+
+## Options
+
+### Anonymous functions
```
kmoe, line 27: It might be clearer if instead of "anonymous functions" and "named functions" (which to me sounds like a distinction between ways of declaring a function - whether we bind an identifier to the function itself), we used the terms "anonymous function types" and "named function types".

docs/design/factories.md
```diff
@@ -0,0 +1,234 @@
...
+users for resources, data sources, and providers:
+
+```go
+type Provider struct {
+  Resources map[string]ResourceFactory
+}
+```
+
+This works whether the type is defined by the provider (the `NewResource()`
+method would return an interface) or by the framework.
```
kmoe, line 146:
```suggestion
method would return an `interface{}`) or by the framework.
```

I think... Or did you mean `framework.Resource`, which would be an interface type?

docs/design/factories.md
```diff
@@ -0,0 +1,234 @@
...
+```
+
+and do the instantiation ourselves behind the scenes.
+
+This works only if the type is defined by the framework. Types defined by the
+provider would need to use reflection to achieve this outcome.
+
+### Factory types
+
+Instead of using functions, we can have an interface for a factory type that
```
kmoe, line 127:
```suggestion
Instead of using functions, the framework can define an interface for a factory that
```

docs/design/factories.md
```diff
@@ -0,0 +1,234 @@
...
+providers define:
```
kmoe, line 128:
```suggestion
can be implemented by a provider-defined type:
```

We might prefer "fulfilled" to "implemented" here - I just noticed that the Go language spec never uses "fulfil".

docs/design/factories.md
```diff
@@ -0,0 +1,234 @@
...
+
+```go
+type ResourceFactory interface {
+  NewResource() Resource
+}
+```
+
+Which can then be used instead of the type the factory will return when asking
+users for resources, data sources, and providers:
```
kmoe, line 137:
```suggestion
consumers for resources, data sources, and providers:
```

docs/design/factories.md
```diff
@@ -0,0 +1,234 @@
...
+
+```go
+type Provider struct {
+  Resources map[string]ResourceFactory
+}
+```
+
+This works whether the type is defined by the provider (the `NewResource()`
```
kmoe, line 145:
```suggestion
This works whether the `Resource` type is defined by the provider (the `NewResource()`
```

docs/design/factories.md
```diff
@@ -0,0 +1,234 @@
...
+	Resources map[string]Resource
+}
+```
+
+and do the instantiation ourselves behind the scenes.
+
+This works only if the type is defined by the framework. Types defined by the
+provider would need to use reflection to achieve this outcome.
+
+### Factory types
```
kmoe, line 125: I've tried to disambiguate this section but may have gone in the wrong direction, so these are especially tentative suggestions.

docs/design/factories.md
```diff
@@ -0,0 +1,234 @@
...
+```
+
+This works whether the type is defined by the provider (the `NewValue()` method
+would return an interface) or by the framework.
+
+## Trade-offs
+
+### Mutability of data
+
+As a general rule of thumb, resources, data sources, and providers should not change at runtime. It's hard to imagine a scenario where modifying the schema, validation, or CRUD implementations while the server is running is a good idea, and we should consider that scenario out of scope for this design.
```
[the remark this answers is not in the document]
paddycarver: I think, without work in Terraform core, we should strongly advise providers away from this. @alexsomesan may have more information, but I think Terraform will be _very_ unhappy if you change the schema of anything at runtime.

Also, GetSchema RPCs currently happen before ConfigureProvider RPCs, I believe.

docs/design/factories.md
paddycarver: I think Go uses tabs [by convention](https://golang.org/doc/effective_go#formatting) but I'm not invested in that choice and willing to bow to readability here.

docs/design/factories.md
paddycarver: I meant `framework.Resource` would be an interface type.

I'm trying very hard to make sure nothing returns an `interface{}`.

docs/design/factories.md
paddycarver: I think I may have accidentally confused the matter by (:sadtrombone:) using the words "type" and "value", so I'm gonna take a stab at clarifying that.

Essentially, I think resources are used in two different ways, and the best way to talk about them is from the perspective of the framework.

The first way the framework sees resources is as a source of schemas. Technically, this doesn't even _need_ to be part of the Go type implementing the resource (it could just be all defined inline in the provider GetSchema method), but I think conceptually it is part of a resource. This can be not-accurately-but-not-entirely-inaccurately as a call that will never happen in parallel to other GetSchema calls. I think, therefore, it makes sense to think of whatever is implementing the GetSchema functionality can be thought of as a singleton-per-resource-per-provider-instance. You're getting the schema of what I'm calling the "resource type" (not to be confused with the framework.Resource type), and there's no real reason to isolate these.

The second way the protocol sees resources is as a group of methods that will be called at runtime. These _will_ be run in parallel, likely, and are concerned with a single _instance_ of the resource. Yes, the methods will stay the same, but the context they're called in will be specific to the instance of a resource, where as the context of GetSchema is called in the context of all instances of the resource. I think, therefore, it makes sense to think about implementing these as "values" or "instances" of that resource "type" and providing isolation between them.

I think, in code, what I'm talking about could be described like this:

```go
// inside the framework module
type ResourceType struct {
  Schema map[string]Schema
  NewResource func() Resource
}

type Resource interface {
  Create(ctx context.Context) // etc
}
```

```go
// inside the provider

// the resource "type" is just an instance of a framework.ResourceType type. All resource types share the same
// implementation/functionality, the only data that changes is the schema, the shape of the resource.
var computeInstanceType = framework.Resource{
  Schema: map[string]framework.Schema{
    "name": framework.Schema{},
  },
  NewResource: func() framework.Resource{
    return computeInstanceResource{}
  }
}

// the CRUD methods are defined on a separate type, because they aren't shared among all resource types, they're shared
// among all instances of a resource type.
type computeInstanceResource struct{}

func (c computeInstanceResource) Create(ctx context.Context) {}
```

Basically, I think if we're thinking of Go types as a collection of common data structures and behaviors, "things that return schemas" have more in common with each other than they do with "things that operate on data described by their schema", given how they're used.

This is related to the factory problem because I discovered this while trying out things like this:

```go
type Resource interface{
  GetSchema()
  NewResource() Resource
  Create()
  Read()
  Update()
  Delete()
}
```

Basically, combining what I'm calling "ResourceType" and "Resource"/"ResourceValue" into a single type. In that situation, we're stuck in a loop--we need an instance of `Resource` to get an instance of `Resource`. This led to factories:

```go
type ResourceFactory interface {
  NewResource() Resource
}

type Resource interface{
  GetSchema()
  Create()
  Read()
  Update()
  Delete()
}
```

But having a separate type that doesn't hold any data doesn't make sense. And then I realised there was no reason it _couldn't_ hold the schema, and that draws a nice separation between the Schema as something Terraform retrieves at the beginning of the run and the CRUD calls as something that happens on instances of that Schema.

I'm really struggling to articulate this, sorry :(

paddycarver:
Think I've addressed all the feedback. :)

docs/design/factories.md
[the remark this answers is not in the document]
kmoe: Yes, good point. If we see a provider that calculates the response to `GetSchema` based on (say) an API response as "changing the schema at runtime" or "while the server is running", then I think it is a plausible design, but since, as you rightly point out, this must happen before `ConfigureProvider`, I think it's sufficiently far from our intended use case as to be considered out of scope for the framework design.

docs/design/factories.md
kmoe: Yes, let's at least rephrase this so we're not using the overloaded terms "type" and "value" - I still think the distinction between `ResourceType` and `Resource`/`ResourceValue` is confusing and doesn't do much to help provider developers.


> Yes, the methods will stay the same, but the context they're called in will be specific to the instance of a resource, where as the context of GetSchema is called in the context of all instances of the resource. 

This is what I meant by "A resource's Create method is always the same regardless of which instance of the resource we're talking about - that information comes in via the parameters of the Create function." - this distinction is already incorporated in the parameters of `GetSchema` vs those of `Create`. Providers do not need to have the concept of resource instance exposed to them via the framework, because they can't do anything with it: modulo [runtime state](https://github.com/hashicorp/terraform-plugin-framework/blob/main/docs/design/runtime-state.md#data), Create is a pure function of prior/planned resource state and config, to new state. 

Regarding the following:
```go
type Resource interface{
  GetSchema()
  NewResource() Resource
  Create()
  Read()
  Update()
  Delete()
}
```

It's already an anti-pattern trying to make an object's factory or constructor be a method on that object. I don't think we need another reason why `NewResource` needs to be separated out.

I like the [mix of structs and interfaces](https://github.com/hashicorp/terraform-plugin-framework/blob/main/docs/design/structs-interfaces.md#alternative-3-mix-of-structs-and-interfaces) approach, for reasons possibly different to the ones you go into here. The result might be the same, though:

```go
type ResourceFactory interface {
  NewResource(Provider) (Resource, []*tfprotov5.Diagnostic)
}

type Resource struct {
  Schema                *tfprotov5.Schema
  Implementation        func(p ProviderState) ResourceImplementation
}

type ResourceImplementation interface {
  Create(ctx context.Context, *CreateResourceRequest, *CreateResourceResponse)
  Read(ctx context.Context, *ReadResourceRequest, *ReadResourceResponse)
  Update(ctx context.Context, *UpdateResourceRequest, *UpdateResourceResponse)
  Delete(ctx context.Context, *DeleteResourceRequest, *DeleteResourceResponse)
}
```

The crux of my response is: What benefit would provider developers get from separating `ResourceType` and `ResourceValue`, except what they already get from separating a resource struct type from a resource interface from a resource factory (interface)?

paddycarver: I think the `ResourceType`/`Resource` distinction is basically the mix of structs/interfaces approach you linked to just with different names and different contexts. If you think of `Resource` in the example you linked as `ResourceType`, and `ResourceImplementation` as `Resource`, you basically have the same exact types I documented above; you just don't need the `ResourceFactory` type for anything, arguably.

I think why I'm calling it out here is:

1. There's no guarantee we're going to choose to use that approach, so I want to address the trade-offs in this document, too. That way, if we decide to go with another approach, we can still talk about the trade-offs of separating or combining these concepts when we get to this decision and are considering it.
2. I think the struct and interface doc does a great job of examining how that choice impacts provider devs through its lens (the Go type system we expose), but I think it's out of scope for that document to talk about what I'm trying to address here, which is conceptual distinctions.

I'm gonna try to summarize the conceptual distinction I'm trying to draw, because I think I've been having difficulty articulating it, so I'm gonna take another stab at it. If you think you understand it and just disagree, please feel free to chime in with that and I'll stop trying to explain it. I just feel like I'm circling around what I'm trying to say and haven't quite found the center of the idea yet.

I think the core distinction I'm trying to make here is between an instance of a resource and the type of a resource. The protocol itself makes this distinction: the RPC requests define a [TypeName](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-go/tfprotov6#ApplyResourceChangeRequest.TypeName) property identify the resource type they're operating on. The schema is returned using [a separate RPC](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-go/tfprotov6#GetProviderSchema) which is asking about the resource type, not any specific instance of the resource. Core makes the distinction between a resource type and a resource instance. I think the question I'm asking here is: do we want to make that distinction, too? I don't know what my answer is, but I think it's a reasonable question to explore.

You raise that providers don't need to have the concept of resource instances exposed to them, because the function is pure and has all the data it needs (besides, presumably, the [provider-owned runtime data](https://github.com/hashicorp/terraform-plugin-framework/blob/main/docs/design/runtime-data.md)) passed via the arguments into it. I think this is true, but maybe not _obviously_ true; the type the `Create` method is defined on, arguably, is state that provider developers can store things in during Create or Read, and if we don't create a clear distinction or explicitly say it's not supported, they may see code that _sometimes_ works (subject to race conditions and graph ordering) and build a reliance on that.

I think one of the trade-offs I didn't list in this section, that I'll add when revamping it, is that if we make that distinction, we need to figure out how we're going to give provider developers access to that schema information in their CRUD methods (if we can't prove they don't need it).

paddycarver: Also, I guess it's worth calling out, since we're making the decision about mixing structs/interfaces separately from this, if we decide to do that, and then need to consider this decision, I'd like this option explicitly called out so there's no chance we'll end up going with an option that doesn't appear in this document.

kmoe: Incomplete feedback to get started:

>I think the ResourceType/Resource distinction is basically the mix of structs/interfaces approach you linked to just with different names and different contexts.
Yes, this is what I meant by "I like the mix of structs and interfaces approach, for reasons possibly different to the ones you go into here. The result might be the same, though". Perhaps that doc would be a better place for the discussion of the conceptual distinction being drawn here, except for how it impacts factories in particular.

>the type the Create method is defined on, arguably, is state that provider developers can store things in during Create or Read, and if we don't create a clear distinction or explicitly say it's not supported, they may see code that sometimes works (subject to race conditions and graph ordering) and build a reliance on that.
The doc should be clear on the claims I think are being made here, that:
1. The provider-defined Resource struct type, which implements the framework.Resource interface, maps to the concept of a resource instance, and
2. This would sometimes work and sometimes not (this should be expanded on with examples of what "working" means).

kmoe (approved):
Excellent, thank you for the clarifying examples in the "Separating resource types and resource instances", which are very instructive.

bflad:
For my first read through here, it would be super helpful to annotate each of the code examples with `// in the framework` or `// in the provider` to reduce any confusion of who is implementing what. 👍

bflad (approved):
Overall this feels like a nice summary of this topic.

Some helpful enhancements might be:

* Annotating code examples to distinguish the implementor (framework vs provider code)
* Whether including a `sync.Locker` for resource types can alleviate the mutex problem (or potentially offering it as its own common separate type for separate of concerns and reducing provider implementation burden with custom types)
* Discussing more about resource instance state reading/writing of each option as these are critical to the data conversion to/from remote system operations -- I guess I would just be looking for quick code examples of each, if possible
* Discussing how resource type schema versions (in the tfprotov5 vs tfprotov6 sense) will be handled
* Discussing how resource instance schema versions (in the state upgrade/migration sense) will be handled, if necessary

If these are discussed elsewhere, it may be helpful to link out as appropriate. 👍

paddycarver:
> Annotating code examples to distinguish the implementor (framework vs provider code)

Done, thanks!

> Whether including a sync.Locker for resource types can alleviate the mutex problem (or potentially offering it as its own common separate type for separate of concerns and reducing provider implementation burden with custom types)

I think this may be a separate topic at that point and may be out of scope. I think the goal of that section is establishing that there is a certain set of data--whatever shape it takes--that is shared amongst all instances of a resource. That _can_ be a sync.Mutex, but it could also be, for example, a data type that helps batch Get requests.

> Discussing more about resource instance state reading/writing of each option as these are critical to the data conversion to/from remote system operations -- I guess I would just be looking for quick code examples of each, if possible

Are you asking about config/state/plan reading? I think that's out of scope here, as the choice we make here shouldn't impact that at all--I think we've largely consolidated on passing that all (bundled up in a type) into CRUD functions as an argument, no matter which of these approaches we take.

> Discussing how resource type schema versions (in the tfprotov5 vs tfprotov6 sense) will be handled

Do you mean how we'd handle different protocol versions? Not sure this is the place for that conversation, again, because I don't know that that impacts any of the options available to us.

> Discussing how resource instance schema versions (in the state upgrade/migration sense) will be handled, if necessary

Likewise, I think all these options handle this in the same way, and it's therefore probably an orthogonal design decision to be made separately.

I think we're trying to keep a fairly narrow focus on these documents: given a design question, what nuance needs explored to help us differentiate between the possible approaches available to us? Everything outside that gets split out into its own doc to be made as a separate decision, in an attempt to break this down into reasonably digestible chunks for discussion.

docs/design/factories.md
```diff
@@ -0,0 +1,439 @@
...
+### Anonymous function types
+
+Wherever we need instances of provider, resource, and data source types from
+provider developers, we can ask for them as the results of anonymous function
+types.  For example:
+
+```go
+// in the framework
+type Provider struct {
+  Resources map[string]func() framework.Resource
```
bflad, line 46: Since this is in the framework:

```suggestion
  Resources map[string]func() Resource
```

docs/design/factories.md
```diff
@@ -0,0 +1,439 @@
...
+would return an interface) or by the framework.
+
+### Named function types
+
+Similarly to anonymous function types, we can ask for functions with a named
+function type:
+
+```go
+// in the framework
+type ResourceFactory func() framework.Resource
```
bflad, line 62: Since this is in the framework:

```suggestion
type ResourceFactory func() Resource
```

docs/design/factories.md
```diff
@@ -0,0 +1,439 @@
...
+We have the option of surfacing this distinction explicitly, allowing providers
+to store this state that should be shared among all instances of a resource
+type. We could define the resource type and the resource instance as separate
+Go types:
+
+```go
+// in the framework
+type ResourceType interface {
+  GetSchema() *tfprotov5.Schema
+  NewResource(p framework.Provider) Resource
```
bflad, line 248: Since this is in the framework:

```suggestion
  NewResource(p Provider) Resource
```

docs/design/factories.md
```diff
@@ -0,0 +1,439 @@
...
+    // hard-code schema here
+  }
+}
+
+type myResource struct {
+  reqMutex sync.Mutex
+  client *Client
+}
+
+func (m *myResource) Create(ctx, req, resp) {
```
bflad, line 286: Since this is in the provider:

```suggestion
func (m *myResource) Create(ctx context.Context, req *framework.CreateResourceRequest, resp *framework.CreateResourceResponse) {
```

bflad (commented):
I'll PR these separately as clarifications

PR #6, opened 2021-04-22
create schemas design doc

kmoe:


docs/design/schemas.md
```diff
@@ -0,0 +1,180 @@
...
+```
+
+or
+```go
+type Resource struct {
+  GetSchema func(context.Context) (*tfprotov6.Schema, []*tfprotov6.Diagnostic)
+}
+```
+
+If `framework.Resource` is an interface, the only option is a method (`framework.Resource.GetSchema()`), since the actual resource struct will be a provider-defined type:
```
paddycarver, line 27: I think this isn't _strictly_ true. For example, we could do something like this:

```go
type Resource interface {
  // whatever whatever whatever
  GetSchema(ctx context.Context) (*tfprotov6.Schema, []*tfprotov6.Diagnostic)
}

type Schema map[string]*tfprotov6.Schema

func (s Schema) GetSchema(_ context.Context) (*tfprotov6.Schema, []*tfprotov6.Diagnostic) {
  // implement a conversion function here that converts from s to *tfprotov6.Schema
}
```

Then users could do something like this:

```go
type myResource struct {
  framework.Schema
}

func NewMyResource() framework.Resource {
  return myResource{
    "foo": *tfprotov6.Schema{},
    "bar": *tfprotov6.Schema{},
  }
}
```

Which lets them use a method _or_ a struct implementation. Similarly, we could make `GetSchema` an unexported method in the interface (`getSchema`) which would then _require_ them to use our struct implementation.

Whether any of that is a good idea or not I think is open for discussion, but I do think it's an option that exists for us.

docs/design/schemas.md
```diff
@@ -0,0 +1,180 @@
...
+						},
+					},
+				},
+			},
+		},
+	},
+}
+```
+
+### `framework.Schema`
```
paddycarver, line 114: I think I have some nitpicking over what the `framework.Schema` type would look like, but I think the place to do that nitpicking and decide what complexity we could gloss over and what we need to surface could happen on the PR implementing that choice. I think my only ask here is that we just note that this example is illustrative and isn't necessarily the design that we'd ship, that we're just capturing the pattern of abstracting away some of the complexity. I only worry about it because it's such a detailed design. :)

docs/design/schemas.md
```diff
@@ -0,0 +1,180 @@
...
+
+### Method `framework.Resource.GetSchema()`
+
+This approach is also fairly simple and discoverable, and offers the option of returning an error or diag if the schema is not available. This may be a desirable feature for providers written using generated code, where generating the schema could throw an error.
+
+### Both/either
+
+A `Resource` struct could have a `Schema` field, and a `GetSchema` field with func type added later if needed. The framework could use whichever is set, either erroring if both are set or using a documented fallback mechanism to prefer one over the other.
+
+## Types
```
paddycarver, line 48: The only extra thing I'd hope for here that I didn't see, but don't have guidance on how to achieve, is I hope our design can surface the distinction  between the following types and their capabilities:

* Block
* Attribute
* tftypes/cty value

The reason I'm focused on that is I feel like helper/schema conflates all those things, and it has misled a lot of provider developers about the impacts of their choices, given they don't see whether Optional, Computed, or Sensitive is available for fields depending on what type they end up being.

I don't know how to surface that in this, and a fair response to this comment may be "this feels out of scope for this document", but I think I'm arguing we should consider the distinction between those three things "necessary complexity" that we can't abstract away from the user. (Though I think limiting which of them a user can use is fair game, still.)

docs/design/schemas.md
kmoe: Indeed, this was the intention behind "A minimally verbose option could look something like" - I'll make this clearer.

docs/design/schemas.md
kmoe: Yes, great point. Added this option in.

docs/design/schemas.md
kmoe: I added a rather weak closing paragraph naming this tradeoff, but leaving the details to the implementation. Not sure if we should go into more detail here - what do you think?

paddycarver (approved):
Great work, I think this captures everything really well.

PR #7, opened 2021-04-26
Add design doc on resource registration.

paddycarver:
How do we tell a provider about which resources and data sources exist,
so it can list their schemas and route requests to them?


kmoe (approved):
This is all great, really clear. I think there might be another option but we can add it later.

paddycarver:
I'd love to hear more about this other option!

kmoe:
Nothing concrete yet, so I think this is good to merge now, and I'll PR something later if necessary.

PR #11, opened 2021-05-05
Start defining the Schema type.

paddycarver:
Start building out the schema type, leaving some TODOs for things I'm
still mulling over and thinking about.

This also brings in the StringKind type, which is a pattern I'm not
thrilled about and we should probably think about it some more.

I also included the AttributeType and AttributeValue types, which I'll
add a package with some builtins for in a future commit, to show off how
they're used.


paddycarver:
Alright, I'm calling this "ready for review", but I don't think the _code_ is what we'll end up merging. It still needs tests, it needs more basic, builtin types added, and it needs better docstrings on a lot of different things. I mostly just wanted to get enough down to give the basic idea.

I'd like to get some reviews on this direction before I put too much time or energy into polishing it. Is this something y'all are comfortable with? Do you like the way this is shaping up?

As an example, here's a schema I think a provider may write:

```go
mySchema := tf.Schema{
  Attributes: map[string]tf.Attribute{
    "my_string": {
      Type: types.String{},
      Required: true,
      Sensitive: true,
    },
    "my_list_of_strings": {
       Type: types.List{ElemType: types.String{}},
       Optional: true,
       Computed: true,
       DeprecationMessage: "my_list_of_strings is deprecated and will be removed in the next major version. Please use my_nested_attributes instead.",
    },
    "my_nested_attributes": {
      Attributes: map[string]tf.Attribute{
        "my_nested_string": {
          Type: types.String{},
          Optional: true,
          Computed: true,
        },
      },
      AttributesNestingMode: tf.NestingModeList,
      Optional: true,
      Computed: true,
    },
  }
}
```

The `types` package has some ideas about how we could present typed information to users. The SDK would get the `tftypes.Value` off the wire during `ApplyResourceChange` (or whatever), call `ValueFromTerraform` with it on the type defined in the resource's schema, and give the provider the strongly-typed value in their CRUD methods. The provider developer can then modify the value or construct their own and pass it onto the KatyState, and the framework can call the `ToTerraformValue` method on whatever's returned so it can be sent down the wire.

This overlaps a bit with the work that @kmoe is doing, so I'd love early feedback on it. My goal was mostly "how can I support user-defined types like Timestamp or ARN or CIDR or whatever?" and I kinda fell backwards into "how can we expose data to users in a non-frustrating way.

I think there is a downside here in that we would probably still need users to assert the type of the `AttributeValue` that's coming back to get at the value of it, but the benefit it has over `helper/schema` is that we can support multiple types for the same field; using a flag on the `AttributeType` can control the type of `AttributeValue` that gets generated. We could also use separate properties on the `AttributeValue`, each with a distinct type, and set both. This is mostly prompted by our struggle with `json.Number` and `int64` for `schema.TypeInt`, and how much of a mess it was to change out the representation when we needed a more accurate representation. Still thinking and tinkering on this, but interested in y'all's thoughts.

Finally, this PR makes the aggressive choice of `tf` as a package name. Other package names I toyed with:

* `tfsdk`
* `framework`
* `sdk`
* `terraform`

Interested in y'all's thoughts on that, too.

kmoe:
I like this direction. Some initial partial feedback:

1. Love the complete lack of blocks. If we're doing everything with attributes and nesting modes, we have an opportunity to make the nesting mode concept more user-friendly. For example, in the schemas design doc I attempt to make a pseudo-type of the _nested block_ (`framework.NestedBlockType`), the subtext being that the framework type can contain all combinations of protocol schema attribute type and nesting mode, without exposing the nesting mode concept to the framework user at all. Is this possible, and is it desirable?


2. We can avoid exposing users to the additional cognitive load (and import) of `tftypes` by allowing them to use `AttributeType`s for `List.ElemType`s.
 

3. In your example, you'd actually need to have:
```go
			"my_string": {
				Type:      types.StringType{},
				Required:  true,
				Sensitive: true,
			},
```

(instead of `Type: types.String{}`). It would be nice not to require users to sprinkle `{}` about the place - how about making the primitive attribute types ints, like the nesting modes (or even using the `iota` enum pattern)?
For composite types, structs seem correct, and preferable to `helper/schema`'s separate `Schema.Elem` field, but we still have that option if we want to make the attribute type types more consistent. (Helper) funcs are another option but seem strictly worse than ints or structs, being both more verbose and less flexible.


4. This tradeoff is well framed:
>I think there is a downside here in that we would probably still need users to assert the type of the AttributeValue that's coming back to get at the value of it, but the benefit it has over helper/schema is that we can support multiple types for the same field; using a flag on the AttributeType can control the type of AttributeValue that gets generated. We could also use separate properties on the AttributeValue, each with a distinct type, and set both. This is mostly prompted by our struggle with json.Number and int64 for schema.TypeInt, and how much of a mess it was to change out the representation when we needed a more accurate representation.

The `json.Number` debacle was unfortunate, and we certainly need to be sure we can make backwards-compatible changes to types in this framework. I'd hope that if we were to build in some way to change the type generated by primitive `AttributeTypes` (more int precision issues? string encodings? hopefully nothing to do with booleans?), WAGNI.

If the user is calling `ValueFromTerraform()`, they must make a type assertion to convert the `AttributeValue` to the appropriate type, or at least pass it in as the `target` argument to `KatyState.Get()`. I don't at present see a way around this while still allowing users to define custom `AttributeType`s, but I wonder if KatyState can do something clever here and match up the schema type automatically. This is a question for KatyState, which isn't realistically going to be available until Monday, but may build on the work in this PR anyway - this is the point at which I would propose larger modifications to the code, which looks like a great direction for now.

kmoe:
Also, regarding package name, I'm in favour of `terraform`, but all the alternatives you listed are fine, except `sdk` which might conflict with other SDK packages imported by providers.

kmoe:
Currently the root package `tf` is imported by the `types` package, which leads to import cycles when writing tests for the root package that use both packages. I think we can just extract the `tf` identifiers needed by `types` into a third package, and am doing so in my katystate prototype that extends from this PR, but there may be a better solution here.

schema.go
```diff
@@ -0,0 +1,149 @@
...
+	// a value for this attribute or not. Optional and Required cannot both
+	// be true.
+	Optional bool
+
+	// Computed indicates whether the provider may return its own value for
+	// this attribute or not. Required and Computed cannot both be true. If
+	// Required and Optional are both false, Computed must be true, and the
+	// attribute will be considered "read only" for the practitioner, with
+	// only the provider able to set its value.
+	Computed bool
```
bflad, line 86: Passing thought: Is now a good time to introduce this as a single enumeration type instead of allowing plugin developers potentially introduce invalid combinations and allowing for easier static analysis and unit testing (this one field being required)?

Quick sketch being:

```go
const (
    AttributeConfigurationRequired AttributeConfiguration = 0
    AttributeConfigurationOptional AttributeConfiguration = 1
    AttributeConfigurationReadOnly AttributeConfiguration = 2
)

type AttributeConfiguration uint8

type Attribute struct {
    // ... other fields ...
    Configuration AttributeConfiguration
}
```

schema.go
```diff
@@ -0,0 +1,149 @@
...
+	// If Attributes is set, Type cannot be.
+	//
+	// TODO: do we need MaxItems/MinItems? Can we just make those weird
+	// validation helpers?
+	Attributes map[string]Attribute
+
+	// AttributesNestingMode controls the various ways these sub-groups of
+	// attributes can behave. It can only be used with Attributes, and must
+	// not be set if Type is set.
+	AttributesNestingMode NestingMode
```
bflad, line 57: My initial feeling when seeing these three conflicting options was that this feels a little confusing for provider developers and hard to setup for static analysis. Is there a reason to shy away from always requiring `Type`  (embedding the nesting mode if necessary into those implementations) and treating `Attributes` like the old `Elem`?

bflad (commented):
Some initial thoughts/comments -- happy to chat about them!

schema.go
```diff
@@ -0,0 +1,149 @@
...
+	// Attributes are the fields inside the resource, provider, or data
+	// source that the schema is defining. The map key should be the name
+	// of the attribute, and the body defines how it behaves. Names must
+	// only contain lowercase letters, numbers, and underscores.
+	Attributes map[string]Attribute
+
+	// Version indicates the current version of the schema. Schemas are
+	// versioned to help with automatic upgrade process. Whenever you have
+	// a change in the schema you'd like to provide a manual migration for,
+	// you should increment that schema's version by one.
```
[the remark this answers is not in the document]
paddycarver: My concerns about putting too much inline:

1. It'll become out of date if best practices change.
2. I've seen provider developers skim the comment blocks because they were given too much info, and miss the important info they were actually looking for. I worry that if we include too much text, we're going to encourage skimming.

Because of that, I'm tempted to say this belongs more in a Learn guide or a best practices document on the website, where we have the space to tease out the nuance and talk about schema versioning in depth.

schema.go
paddycarver: Now's a great time if it's something we want to introduce! I'm just not convinced it's a thing we want to introduce. I think it was @apparentlymart who attempted to combine required/optional/computed (maybe not in the way you're suggesting) and determined he needed all three booleans to support all the nuance.

I'm interested in hearing more how this impacts static analysis or unit testing.

My concern with the example you shared:

* The distance between where the options are enumerated and their effect explained, and where they're used
* The verbosity that hosting them to global scope forces on us (I don't think anyone actually wants to type `AttributeConfigurationOptionalWithMissingDriftDetection` instead of `Optional: true`.)
* I think this framing is reductive about what `Computed` does. `AttributeConfigurationOptionalWithoutMissingDriftDetection` is (assuming that's Optional + Computed) actually also going to be used for, e.g., providing a default and may have nothing to do with drift detection.

I like the idea of trying to enforce only valid configurations with the compiler, but I'm not sure what we lose in this scenario is worth the benefit. Thought hearing more about the static analysis and unit testing impacts may change my mind.

schema.go
paddycarver: I originally wanted to combine everything into one attribute, but I eventually gave up on it when I realized I was leading provider developers into confusion.

For example, I considered making `Attributes` fill the `Type` interface.  But that would mean that provider developers could try and set Optional/Computed/Sensitive on elements in a list, which isn't supported. This ends places like hashicorp/terraform-plugin-sdk#201, and I think one of the necessary complexities I'm interested in exposing provider developers to with the framework is the difference between a nested attribute and a complex type, and how those decisions have consequences on what capabilities are available to them.

That's why I thought `Type` and `Attributes` should be separate and mutually exclusive--you need to opt into either this is a cty value, where the hole thing shares one set of flags, or this is a set of nested attributes, where each can have its own individual flags. I think there are benefits to each of those, and they're appropriate in different situations, and so I think the provider developer has to decide which set of trade-offs makes sense for their specific situation.

(`AttributeNestingMode` we can definitely get rid of, I just gotta think my way through how to do that properly.)

types/string.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+type StringType struct{}
+
+// TerraformType returns the tftypes.Type that should be used to
+// represent this type. This constrains what user input will be
+// accepted and what kind of data can be set in state. The framework
+// will use this to translate the AttributeType to something Terraform
+// can understand.
+func (s StringType) TerraformType(ctx context.Context) tftypes.Type {
```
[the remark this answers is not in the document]
paddycarver: That's a good question, and I tried to put it where I thought we'd need it for logging, but I think my answer is just going to be "yes, until proven otherwise", with the exception of the `Equal` method, which needs a specific type signature if `go-cmp` is going to be able to use it.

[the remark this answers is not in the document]
paddycarver: Oh, sorry, you were asking the other way, of "can we get rid of it". I think my preference is to keep it here, as it's necessary to get to a logger instance, so anywhere a provider may want to have access to a logger, I want to make sure a context is available.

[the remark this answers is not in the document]
kmoe: Yes, good point regarding loggers. Context everywhere then

schema.go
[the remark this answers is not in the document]
kmoe: Agree with putting this info in the Godoc. If best practices change, we can update it just as easily, if not more easily, than any external documentation.

schema.go
kmoe: I think bflad's suggestion might be the same as mine [above](https://github.com/hashicorp/terraform-plugin-framework/pull/11#issuecomment-834296696), which is not that `Attributes` and `Type` should be combined, but that `Type` and `NestingMode` should be combined. Then `Type` would always be required, and some values of `Type` (like `NestedBlock`, say), would also require `Attributes` to be set.

kmoe: Sorry, I didn't read the whole comment and just saw you acknowledged in the last paragraph that we might be able to get rid of nesting modes.

kmoe:
obj implementation was slightly broken - should also cherry-pick https://github.com/hashicorp/terraform-plugin-framework/pull/26/commits/0c41edf585e32a9f3f7f2cbbbcd03cd4c00654c8

schema.go
paddycarver: I'm thinking more and more about this, and I think it's a useful and valid direction for us to pursue, but I'm not sold on it just yet. I think my hot take is that we should create an issue for exploring this, and schedule that work for the v0.2.0 release.

bflad: I have created https://github.com/hashicorp/terraform-plugin-framework/issues/31 to track this investigation longer term. 👍

paddycarver: Thank you!

paddycarver:
Closing this out in favor of #28, #29, #30, and a soon-to-come complex type PR that will break out the support for Lists, Maps, etc. into their own PR. This just got Too Big.

PR #20, opened 2021-05-10
Define State type

kmoe:
WIP, also contains req/resp types for now

extends from #11 

Summary:
```go
type State struct {
  Raw tfprotov6.DynamicValue
  Schema Schema
}

func (s *State) Get(ctx context.Context, attributePath *tftypes.AttributePath, attrType attribute.AttributeType) (attribute.AttributeValue, error) {}

func (s *State) Set(ctx context.Context, attributePath *tftypes.AttributePath, val interface{}) error {}
```

There's also an example of a type-specific `Get` helper, `GetString`, which returns a `types.String`. We could have one of these for each primitive type, but it's not possible for structural types due to the need to specify `ElemType`.

Puzzle:

It would be nice if users didn't have to make type assertions against values they get from state. I couldn't figure out how to make a version of `Get` which took a `target *attribute.AttributeValue` or `target interface{}` (pointer) and just marshalled the value into that, since `AttributeValue` is an interface.

Still thinking:
 - AttributePath helpers
   - I expect the vast majority of `.Get()` and `.Set()` calls will concern top-level attributes, so we should either have:
      - a quick helper `Attr(attributeName string) tftypes.AttributePath` to plug into state methods, or
      - versions of the state methods which take a string instead of `tftypes.AttributePath`, e.g. `Get(context.Context, string, *tftypes.AttributePath) (attribute.AttributeValue, error)`
 - Nested attributes
   - doesn't work yet


kmoe:
This is not in any way ready to merge, but I'm setting it to ready to review now because I don't want it to block anything while I'm away. Feel free to add/change commits on this branch.

I think the best API for `Get()` would involve the provider initialising an `AttributeValue` variable, passing the pointer, and having the framework set its value.

We should also have a State function that returns a `map[string]AttributeValue` using the types from the schema.

With these two in place, we probably don't need `GetString` and friends. Type assertions in provider code are still needed, but I think this is the best we can do.

server.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"github.com/hashicorp/terraform-plugin-go/tfprotov6"
+	// "github.com/hashicorp/terraform-plugin-go/tfprotov6/server"
+)
+
+// make a tfprotov6.ProviderServer
+// and ResourceServer
+
+type ResourceServer map[string]tfprotov6.ResourceServer
```
paddycarver, line 11: I'm curious what prompted including this in this PR.

state.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		return ret, fmt.Errorf("error walking attribute path in state; %v still remains in the path: %s", remaining, err)
+	}
+
+	typ := types.StringType{}
+	attrValue, err := typ.ValueFromTerraform(ctx, attr.(tftypes.Value))
+	if err != nil {
+		return ret, fmt.Errorf("error converting from tftypes.Value: %s", err)
+	}
+
```
paddycarver, line 55: Could this all be replaced with a call to `Get`?
```suggestion
attrValue, err := s.Get(ctx, attributePath, types.StringType{})
if err != nil {
  return ret, err
}
```

state.go
```diff
@@ -0,0 +1,89 @@
...
+	attrValue, err := attrType.ValueFromTerraform(ctx, attr.(tftypes.Value))
+	if err != nil {
+		return nil, fmt.Errorf("error converting from tftypes.Value: %s", err)
+	}
+
+	return attrValue, nil
+}
+
+func (s *State) Set(ctx context.Context, attributePath *tftypes.AttributePath, val interface{}) error {
```
paddycarver, line 86: What if we expect an `AttributeValue` here?

```suggestion
func (s *State) Set(ctx context.Context, attributePath *tftypes.AttributePath, val AttributeValue) error {
```

state_test.go
```diff
@@ -0,0 +1,98 @@
+package tf
```
paddycarver, line 1: We can use `tf_test` as the package name to avoid import loops, I think.

state_test.go
```diff
@@ -0,0 +1,98 @@
...
+func TestState(t *testing.T) {
+	raw, err := tfprotov6.NewDynamicValue(tftypes.Object{
+		AttributeTypes: map[string]tftypes.Type{
+			"my_string":          tftypes.String,
+			"my_list_of_strings": tftypes.List{ElementType: tftypes.String},
+		},
+	}, tftypes.NewValue(tftypes.Object{
+		AttributeTypes: map[string]tftypes.Type{
+			"my_string":          tftypes.String,
+			"id":                 tftypes.String,
```
paddycarver, line 22: I believe `id` would need to be part of the object type, too.

state_test.go
```diff
@@ -0,0 +1,98 @@
...
+	actualStr, err := state.GetString(context.Background(), tftypes.NewAttributePath().WithAttributeName("my_string"))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	expectedStr := types.String{
+		Value: "katy",
+	}
+
+	if actualStr != expectedStr {
```
paddycarver, line 77: Think you want
```suggestion
	if !actualStr.Equal(expectedStr) {
```

paddycarver:
> I think the best API for Get() would involve the provider initialising an AttributeValue variable, passing the pointer, and having the framework set its value.

I also think that's an interesting direction. #25 has the beginnings of a stab at that approach, built on #12. Specifically commit 2085e26b7dda583912c6c6913b4a6532e0354ab8 implements the approach in `get.go`. It's very rough and was basically a prototype to see how that might work. I think a `Set` (maybe differently named?) could be used to wholesale replace the state or plan when writing for the response. I don't know.

> We should also have a State function that returns a map[string]AttributeValue using the types from the schema.

I'm not sure I disagree, but just know that we can't have the map values be types from the schema, they would need to be `AttributeValue`s, and the user would need to type assert them. But I think that may be a valid approach, too. I'm wondering if we just give users all of these approaches and see what they use, a cull the ones that don't make sense at a later date...

In theory, we could have something like....

```go
type State struct {
  // ....
}

// Get populates the struct passed as `target` with the entire state. No type assertion necessary.
func (s *State) Get(ctx context.Context, target interface{}) error {}

// GetAttribute retrieves the attribute found at `path` and returns it as an attr.Value,
// which provider developers need to assert the type of
func (s *State) GetAttribute(ctx context.Context, path attr.Path) (attr.Value, error) {}

// Set replaces the entire state with the values of the struct passed as `value`.
func (s *State) Set(ctx context.Context, value attr.Value) error {}

// SetAttribute sets the attribute found at `path` with the value passed as `value`
func (s *State) SetAttribute(ctx context.Context, path attr.Path, value attr.Value) error {}
```

And then if users want to translate the state to provider-defined Go types at the boundaries and work with only compiler-checked types internally and not have any chance of a typo not getting caught and causing a bug, they can use `Get` and `Set`. If they don't want to define the extra types are want to be more surgical about what they're retrieving or setting, they can use `GetAttribute` and `SetAttribute`. We could even use a `GetAttributes` and `SetAttributes` that returns/accepts a `map[string]attr.Value` for use cases where people just want to get a map of the data and not have to care about what all the keys are. (Though I still think those use cases are usually.... misguided.)

kmoe:
State work is superseded by #25/#26 (work ongoing), and request/response work is split out into https://github.com/hashicorp/terraform-plugin-framework/pull/27.

server.go
kmoe: Hubris (thinking it would be quick to scaffold everything needed to build a provider) followed by accident (committing this file)

PR #28, opened 2021-05-31
Add support for defining schemas.

### LINKED ISSUE #12: Define Schema type

paddycarver:
This is another attempt at allowing providers to define schemas. It's
born out of the ashes of #11, which gradually grew from baking an apple
pie from scratch into creating the universe. Or however that saying
goes.

This is intentionally limited in scope to just setting up the types for
declaring schemas and the types required by that, namely our attribute
interfaces. Unlike #11, it makes no attempt to use these types for
anything or prove they're the right types; the work done with #11 gives
me confidence that they're a worthwhile direction to pursue.

I'm submitting this as a separate PR to make review easier and to
optimize for mergeability, letting us get some shared types established
while still taking an appropriate amount of time to review the
reflection code that is in our future.

Fixes #12.


paddycarver:
I believe this is ready for review and merge, and it supersedes #11, which has grown in size and scope to a point where reviewing it properly does not seem feasible.

kmoe (approved):
LGTM. Excellent documentation.

PR #29, opened 2021-05-31
Add primitive types.

paddycarver:
This adds to #28 by creating implementations of the attr.Type interface for our primitive types: bools, numbers, and strings.

This is likewise a split of code originally written for #11 that is being split out into its own PR for easier review. (Or, at least, review will be easier when #28 is merged, which should happen before this is merged.)

### DIFF
--- a/types/primitive.go
+++ b/types/primitive.go
@@ -1,0 +1,90 @@
+package types
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+)
+
+type primitive uint8
+
+const (
+	// StringType represents a UTF-8 string type.
+	StringType primitive = iota
+
+	// NumberType represents a number type, either an integer or a float.
+	NumberType
+
+	// BoolType represents a boolean type.
+	BoolType
+)
+
+var (
+	_ attr.Type = StringType
+	_ attr.Type = NumberType
+	_ attr.Type = BoolType
+)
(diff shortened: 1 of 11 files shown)


paddycarver:
These tests rely on hashicorp/terraform-plugin-go#85, which fixes unmarshaling into uninstantiated *big.Floats.

types/primitive.go
```diff
@@ -0,0 +1,90 @@
...
+	case BoolType:
+		return boolValueFromTerraform(ctx, in)
+	default:
+		panic(fmt.Sprintf("unknown primitive %d", p))
+	}
+}
+
+// Equal returns true if `o` is also a primitive, and is the same type of
+// primitive as `p`.
+func (p primitive) Equal(o attr.Type) bool {
```
kmoe, line 78: What's the thinking behind this? Where will this be used?

paddycarver: Largely in test functions, it's an interface to make `go-cmp` work.

types/primitive_test.go
```diff
@@ -0,0 +1,191 @@
...
+	t.Run(BoolType.String(), func(t *testing.T) {
+		t.Parallel()
+
+		testBoolValueFromTerraform(t, false)
+	})
+}
+
+// testAttributeType is a dummy attribute type to compare against with Equal to
+// make sure we can handle unexpected types being passed in.
+type testAttributeType struct{}
```
kmoe, line 56: v nice :+1:

kmoe (approved):
delightful, especially the tests

PR #30, opened 2021-06-01
Add internal/reflect package

paddycarver:
Add a package that will let us build reflect.Values from any Go type and
populate them with the data from any tftypes.Value. This culminates in
the reflect.Into function, which lets us have json.Unmarshal-esque
behavior but for tftypes.Values.

This will be used later for making aggregate types easier to work with.

### DIFF
--- a/internal/reflect/reflection.go
+++ b/internal/reflect/reflection.go
@@ -1,0 +1,141 @@
+package reflect
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"math/big"
+	"reflect"
+
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+)
+
+// Into uses the data in `val` to populate `target`, using the reflection
+// package to recursively reflect into structs and slices. If `target` is an
+// AttributeValue, its assignment method will be used instead of reflecting. If
+// `target` is a tftypes.ValueConverter, the FromTerraformValue method will be
+// used instead of using reflection. Primitives are set using the val.As
+// method. Structs use reflection: each exported struct field must have a
+// "tfsdk" tag with the name of the field in the tftypes.Value, and all fields
+// in the tftypes.Value must have a corresponding property in the struct. Into
+// will be called for each struct field. Slices will have Into called for each
+// element.
+func Into(ctx context.Context, val tftypes.Value, target interface{}, opts Options) error {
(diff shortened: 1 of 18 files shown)


paddycarver:
This is based on #29 and dependent on that merging first.

paddycarver:
I am generally unhappy with the test situation here; things are tested and we have good coverage, and I'm happy with e.g. primitives, helpers, and numbers. But for larger things like structs, slices, maps, etc. I'm not sure how to test things while keeping the number of tests reasonable, because it's conceivable that e.g. setting a slice directly would work, but setting a slice as a property of an object wouldn't. Or vice-versa.

I think we may want to improve on that over time, though; I don't know that it's a problem we're able to fully solve in the moment. We can do _better_; I have a TODO for standalone testing of pointers and interfaces, which I think can reasonably be achieved and will add those tests, but I think the whole thing is a bit messier than I'd like.

paddycarver:
OK, I think this is ready for review again.

internal/reflect/reflection.go
```diff
@@ -0,0 +1,141 @@
...
+// to give consumers the ability to override its default behaviors wherever
+// possible.
+func buildReflectValue(ctx context.Context, val tftypes.Value, target reflect.Value, opts Options, path *tftypes.AttributePath) (reflect.Value, error) {
+	// if this isn't a valid reflect.Value, bail before we accidentally
+	// panic
+	if !target.IsValid() {
+		return target, path.NewErrorf("invalid target")
+	}
+	// if this is an attr.Value, build the type from that
+	if target.Type().Implements(reflect.TypeOf((*attr.Value)(nil)).Elem()) {
```
kmoe, line 49: Why not `if _, ok := target.(attr.Value); ok` here?

paddycarver: Because target _doesn't_ implement attr.Value, the interface _inside_ it does. I could use `if _, ok := target.Interface().(attr.Value); ok`, but [`.Interface()` can panic](https://golang.org/pkg/reflect/#Value.Interface) so if I don't _have_ to use it, I'd rather stick with the `Implements` method so we don't wind up in a situation in a few months/years where we accidentally access unexported struct fields and panic.

paddycarver: (we could guard it with a `.CanInterface()`, but `Implements()` just seemed cleaner, even though the code is uglier.)

kmoe (approved):

PR #39, opened 2021-06-02
Implement reflect.OutOf

kmoe:
Implement `reflect.OutOf`.


paddycarver:
I think this is now ready for review. It has a good amount of test coverage, and seems to be able to do what we want from it, which is generating an `attr.Value` from an arbitrary Go type and a schema. Map support still needs testing, and tuples and sets still need handled, but we can do that as those types land, just like reflect.Into.

internal/reflect/options.go
```diff
@@ -18,3 +18,6 @@ type Options struct {
 	// returning errors. Numbers will always be rounded towards 0.
 	AllowRoundingNumbers bool
 }
+
+type OutOfOptions struct {
```
kmoe, line 22: Is `OutOfOptions` needed at all now?

internal/reflect/number.go
```diff
@@ -189,3 +189,79 @@ func Number(ctx context.Context, typ attr.Type, val tftypes.Value, target reflec
 	}
 	return target, path.NewErrorf("can't convert number to %s", target.Type())
 }
+
+func FromInt(ctx context.Context, typ attr.Type, val int64, opts OutOfOptions, path *tftypes.AttributePath) (attr.Value, error) {
```
kmoe, line 193: Needs a quick golint - these and other exported functions need comments.

internal/reflect/options.go
paddycarver: No, I left it in case, but it's all internal, anyways, so I guess there's no reason not to YAGNI it.

kmoe:
Excellent! I hereby approve the latest changes (I can't mark the PR approved as I'm the author).

PR #41, opened 2021-06-03
Start supplying attr.Type to reflection.

paddycarver:
// TODO: fill in a better commit message later.


kmoe (approved):
LGTM depending on https://github.com/hashicorp/terraform-plugin-framework/pull/44

paddycarver:
> Actually, I'm not sure I like having to duplicate our type implementations inside the reflect package with `testStringType` etc. Why not just use a `reflect_test` package?

Same, but the point of these tests is to test unexported functions, which we can't do if we use a `reflect_test` package. :( But no amount of package shenanigans will get us around this one, because `types` has to rely on these functions and these functions have to have concrete types. We could combine them into a single package, but then we can't reuse them in Get, etc. We could move the type implementations internal, and reference them from the types package, but I have a hunch that's gonna get messy real quick.

I don't particularly like the copying, either, but I couldn't think of something that was less bad.

kmoe:
Yes, good point. I suppose the only other option is to make those functions exported, which has no user impact because the reflect package is internal.

paddycarver:
Hm, let me give it a shot and see how we like the result.

attr/type.go
```diff
@@ -33,7 +33,7 @@ type Type interface {
 type TypeWithAttributeTypes interface {
 	Type
 
-	// WithAttributeTypes returns a new copy of the type with its
+	// WithAttributesTypes returns a new copy of the type with its
```
kmoe, line 36:
```suggestion
	// WithAttributeTypes returns a new copy of the type with its
```

kmoe:
Looks good!

paddycarver:
Doing some last tweaking of comments in the reflect package to make `golint` happy, then I'll rebase and squash it all down with a nice commit message.

paddycarver:
Okay, golint wanted some pretty invasive changes, but I think we're good now if you're still happy, @kmoe

PR #43, opened 2021-06-06
Implement State, State.Get, State.GetAttribute

kmoe:
WIP. The main work here is the addition of `ApplyTerraform5AttributePathStep` to all schema types, implementing the `tftypes.AttributePathStepper` interface so that we can traverse the schema with [`tftypes.WalkAttributePath`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-go@v0.3.0/tftypes#WalkAttributePath).


kmoe:
Superseded by https://github.com/hashicorp/terraform-plugin-framework/pull/46

PR #44, opened 2021-06-07
add attr.ObjectType

kmoe:
Adds the `ObjectType` interface, needed for reflection work.

The `AttributeTypes()` function could also be called `GetAttributeTypes()` - I've tried to make it consistent with `Type.TerraformType()`.

Will add changelog entry when approved.

### DIFF
--- a/attr/type.go
+++ b/attr/type.go
@@ -28,6 +28,47 @@ type Type interface {
 	Equal(Type) bool
 }
 
+// TypeWithAttributeTypes extends the Type interface to include information about
+// attribute types. Attribute types are part of the definition of an object type.
+type TypeWithAttributeTypes interface {
+	Type
+
+	// WithAttributeTypes returns a new copy of the type with its
+	// attribute types set.
+	WithAttributeTypes(map[string]Type) TypeWithAttributeTypes
+
+	// AttributeTypes returns the object's attribute types.
+	AttributeTypes() map[string]Type
+}
+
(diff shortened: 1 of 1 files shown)


attr/type.go
```diff
@@ -28,6 +28,23 @@ type Type interface {
...
+	// WithAttributeTypes returns a new copy of the type with its
+	// attribute types set.
+	WithAttributeTypes(map[string]Type) ObjectType
+
+	// AttributeTypes returns the object's attribute types.
+	AttributeTypes() map[string]Type
+
+	// Equal must return true if the Type is considered semantically equal
+	// to the Type passed as an argument.
+	Equal(Type) bool
```
paddycarver, line 45: I think this is included with the embedding of `Type`....

```suggestion
```

paddycarver (approved):
One minor comment, but LGTM. 🚀

kmoe:
Thanks for highlighting the additional types added in #41 - since these will be used in a number of open PRs I think it's best to have the conversation here. 

New proposal is to use naming consistent with the other types in `types.go`, i.e. `TypeWith*`. I think this is explicit; although it is unfortunate that a type called `TypeWithAttributeTypes` must have a method `WithAttributeTypes`, it shouldn't mean we have to read a stutter anywhere.

The versions proposed in #41 have the advantage of brevity but I think a method name of `WithElementsTypes` is inherently confusing.

PR #46, opened 2021-06-08
Implement State, State.Get, State.GetAttribute, Config, Plan

kmoe:
The changes to `internal/reflect` on `main` made rather a mess of `katystate3`. We're up to `katystate5` now.

The main work here is the addition of `ApplyTerraform5AttributePathStep` to all schema types, implementing the `tftypes.AttributePathStepper` interface so that we can traverse the schema with [`tftypes.WalkAttributePath`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-go@v0.3.0/tftypes#WalkAttributePath).

### DIFF
--- a/config.go
+++ b/config.go
@@ -1,0 +1,51 @@
+package tfsdk
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/internal/reflect"
+	"github.com/hashicorp/terraform-plugin-framework/schema"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+)
+
+// Config represents a Terraform config.
+type Config struct {
+	Raw    tftypes.Value
+	Schema schema.Schema
+}
+
+// Get populates the struct passed as `target` with the entire config.
+func (c Config) Get(ctx context.Context, target interface{}) error {
+	return reflect.Into(ctx, c.Schema.AttributeType(), c.Raw, target, reflect.Options{})
+}
+
+// GetAttribute retrieves the attribute found at `path` and returns it as an
+// attr.Value. Consumers should assert the type of the returned value with the
+// desired attr.Type.
+func (c Config) GetAttribute(ctx context.Context, path *tftypes.AttributePath) (attr.Value, error) {
+	attrType, err := c.Schema.AttributeTypeAtPath(path)
+	if err != nil {
+		return nil, fmt.Errorf("error walking schema: %w", err)
+	}
+
+	attrValue, err := c.terraformValueAtPath(path)
+	if err != nil {
+		return nil, fmt.Errorf("error walking config: %w", err)
+	}
+
+	return attrType.ValueFromTerraform(ctx, *attrValue)
+}
+
+func (c Config) terraformValueAtPath(path *tftypes.AttributePath) (*tftypes.Value, error) {
+	rawValue, remaining, err := tftypes.WalkAttributePath(c.Raw, path)
+	if err != nil {
+		return nil, fmt.Errorf("%v still remains in the path: %w", remaining, err)
+	}
+	attrValue, ok := rawValue.(tftypes.Value)
+	if !ok {
+		return nil, fmt.Errorf("got non-tftypes.Value result %v", rawValue)
+	}
+	return &attrValue, err
+}
--- a/config.go
+++ b/config.go
@@ -35,17 +35,17 @@ func (c Config) GetAttribute(ctx context.Context, path *tftypes.AttributePath) (
 		return nil, fmt.Errorf("error walking config: %w", err)
 	}
 
-	return attrType.ValueFromTerraform(ctx, *attrValue)
+	return attrType.ValueFromTerraform(ctx, attrValue)
 }
 
-func (c Config) terraformValueAtPath(path *tftypes.AttributePath) (*tftypes.Value, error) {
+func (c Config) terraformValueAtPath(path *tftypes.AttributePath) (tftypes.Value, error) {
 	rawValue, remaining, err := tftypes.WalkAttributePath(c.Raw, path)
 	if err != nil {
-		return nil, fmt.Errorf("%v still remains in the path: %w", remaining, err)
+		return tftypes.Value{}, fmt.Errorf("%v still remains in the path: %w", remaining, err)
 	}
 	attrValue, ok := rawValue.(tftypes.Value)
 	if !ok {
-		return nil, fmt.Errorf("got non-tftypes.Value result %v", rawValue)
+		return tftypes.Value{}, fmt.Errorf("got non-tftypes.Value result %v", rawValue)
 	}
-	return &attrValue, err
+	return attrValue, err
 }
--- a/schema/nested_attributes.go
+++ b/schema/nested_attributes.go
@@ -1,5 +1,11 @@
 package schema
 
+import (
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+)
+
 type nestingMode uint8
 
 const (
@@ -34,6 +40,7 @@ const (
 type NestedAttributes interface {
 	getNestingMode() nestingMode
 	getAttributes() map[string]Attribute
+	tftypes.AttributePathStepper
 }
 
 type nestedAttributes map[string]Attribute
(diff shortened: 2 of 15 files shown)


schema/nested_attributes.go
```diff
@@ -59,6 +66,16 @@ func (s singleNestedAttributes) getNestingMode() nestingMode {
 	return nestingModeSingle
 }
 
+// ApplyTerraform5AttributePathStep applies the given AttributePathStep to the
+// nested attributes.
+func (s singleNestedAttributes) ApplyTerraform5AttributePathStep(step tftypes.AttributePathStep) (interface{}, error) {
+	if _, ok := step.(tftypes.ElementKeyString); !ok {
```
paddycarver, line 72: I think we'd want AttributeName here.

schema/nested_attributes.go
```diff
@@ -59,6 +66,16 @@ func (s singleNestedAttributes) getNestingMode() nestingMode {
...
+		return nil, fmt.Errorf("cannot apply step %T to SingleNestedAttributes", step)
+	}
+
+	return s.nestedAttributes, nil
```
paddycarver, line 76: I think we'd want to return _just_ the attribute named in the path step.

schema/nested_attributes.go
```diff
@@ -118,6 +145,16 @@ func (s setNestedAttributes) getNestingMode() nestingMode {
 	return nestingModeSet
 }
 
+// ApplyTerraform5AttributePathStep applies the given AttributePathStep to the
+// nested attributes.
+func (s setNestedAttributes) ApplyTerraform5AttributePathStep(step tftypes.AttributePathStep) (interface{}, error) {
+	if _, ok := step.(tftypes.ElementKeyInt); !ok {
+		return nil, fmt.Errorf("cannot apply step %T to ListNestedAttributes", step)
```
paddycarver, line 152:
```suggestion
		return nil, fmt.Errorf("cannot apply step %T to SetNestedAttributes", step)
```

schema/nested_attributes.go
```diff
@@ -118,6 +145,16 @@ func (s setNestedAttributes) getNestingMode() nestingMode {
 	return nestingModeSet
 }
 
+// ApplyTerraform5AttributePathStep applies the given AttributePathStep to the
+// nested attributes.
+func (s setNestedAttributes) ApplyTerraform5AttributePathStep(step tftypes.AttributePathStep) (interface{}, error) {
+	if _, ok := step.(tftypes.ElementKeyInt); !ok {
```
paddycarver, line 151: I think we want ElementKeyValue here, as Sets are indexed by Value.

schema/nested_attributes.go
```diff
@@ -147,3 +184,13 @@ type MapNestedAttributesOptions struct {
 func (m mapNestedAttributes) getNestingMode() nestingMode {
 	return nestingModeMap
 }
+
+// ApplyTerraform5AttributePathStep applies the given AttributePathStep to the
+// nested attributes.
+func (m mapNestedAttributes) ApplyTerraform5AttributePathStep(step tftypes.AttributePathStep) (interface{}, error) {
+	if _, ok := step.(tftypes.ElementKeyString); !ok {
+		return nil, fmt.Errorf("cannot apply step %T to SingleNestedAttributes", step)
```
paddycarver, line 192:
```suggestion
		return nil, fmt.Errorf("cannot apply step %T to MapNestedAttributes", step)
```

schema/nested_attributes.go
paddycarver: Thinking about it, I think this method should just go on the `nestedAttributes` type, with `SingleNestedAttributes` inheriting it and the others overwriting it. This would allow the `nestedAttributes` the others are returning to be indexed into.

state_test.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+	if val.Bar.Elems[0].(types.String).Value != "red" {
+		t.Errorf("Expected Bar's first element to be %q, got %q", "red", val.Bar.Elems[0].(types.String).Value)
+	}
+	if val.Bar.Elems[1].(types.String).Value != "blue" {
+		t.Errorf("Expected Bar's second element to be %q, got %q", "blue", val.Bar.Elems[1].(types.String).Value)
+	}
+	if val.Bar.Elems[2].(types.String).Value != "green" {
+		t.Errorf("Expected Bar's third element to be %q, got %q", "green", val.Bar.Elems[2].(types.String).Value)
+	}
```
paddycarver, line 80: I'm sure I probably wrote this, but I think it can be replaced with

```suggestion
	expected := myType{
	  Foo: types.String{Value: "hello, world"},
	  Bar: types.List{
	    ElemType: types.StringType,
	    Elems: []attr.Value{
	      types.String{Value: "red"},
	      types.String{Value: "blue"},
	      types.String{Value: "green"},
	    },
	  },
	}
	diff := cmp.Diff(expected, val); diff != "" {
	  t.Errorf("unexpected diff (+wanted, -got): %s", diff)
	}
```

without losing anything, and gaining some resilience to future changes.

types/list.go
```diff
@@ -91,6 +91,16 @@ func (l ListType) Equal(o attr.Type) bool {
...
+	if _, ok := step.(tftypes.ElementKeyInt); !ok {
+		return nil, fmt.Errorf("cannot apply step %T to ListType", step)
+	}
+
+	return l.ElemType, nil
+}
+
 // List represents a list of AttributeValues, all of the same type, indicated
 // by ElemType.
 type List struct {
```
paddycarver, line 106: Do we need an `ApplyTerraform5AttributePathStep` on the `List` type, too?

kmoe: Why?

kmoe:
It turns out that nested attributes did not need the `AttributePathStepper` implementation, at least if we want to represent (for example) `ListNestedAttributes` as a `types.ListType`. Instead, the schema and its nested attribute types get a recursive `AttributeType()` method, which returns a reasonable `attr.Type` representation. This is needed in any case for `State.Get`.

Please take a look at the tests to see if this provides the appropriate UX.

types/list.go
paddycarver: Ugh okay I see. For some reason I thought `GetAttribute` would be traversing an `attr.Value`, but it's trying to _make_ an `attr.Value` at that point. So we need the `attr.Type` to be an AttributePathStepper (so we can get the relevant schema) and we need `tftypes.Value` to be an AttributePathStepper (so we can get the sub-`tftypes.Value` we're talking about) and then we can combine those two to make an `attr.Value`.

That makes sense. Sorry for the confusion here.

types/object.go
```diff
@@ -101,6 +101,16 @@ func (o ObjectType) Equal(candidate attr.Type) bool {
 	return true
 }
 
+// ApplyTerraform5AttributePathStep applies the given AttributePathStep to the
+// object.
+func (o ObjectType) ApplyTerraform5AttributePathStep(step tftypes.AttributePathStep) (interface{}, error) {
+	if _, ok := step.(tftypes.ElementKeyString); !ok {
```
paddycarver, line 107: I think this wants to be an AttributeName.
```suggestion
	if _, ok := step.(tftypes.AttributeName); !ok {
```

Because objects have attributes, not elements.

types/object.go
```diff
@@ -101,6 +101,16 @@ func (o ObjectType) Equal(candidate attr.Type) bool {
...
+		return nil, fmt.Errorf("cannot apply step %T to ObjectType", step)
+	}
+
+	return o.AttrTypes[string(step.(tftypes.ElementKeyString))], nil
```
paddycarver, line 111:
```suggestion
	return o.AttrTypes[string(step.(tftypes.AttributeName))], nil
```

paddycarver (commented):
Some minor feedback on this one. I think the only part that's super important is the object using the wrong attribute path step type.

types/object.go
kmoe: Thanks, adding a test

config.go
```diff
@@ -0,0 +1,51 @@
...
+func (c Config) terraformValueAtPath(path *tftypes.AttributePath) (*tftypes.Value, error) {
+	rawValue, remaining, err := tftypes.WalkAttributePath(c.Raw, path)
+	if err != nil {
+		return nil, fmt.Errorf("%v still remains in the path: %w", remaining, err)
+	}
+	attrValue, ok := rawValue.(tftypes.Value)
+	if !ok {
+		return nil, fmt.Errorf("got non-tftypes.Value result %v", rawValue)
+	}
+	return &attrValue, err
```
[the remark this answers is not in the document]
kmoe: Thanks, fixing

paddycarver (approved):
Let's do it. 🚀

PR #48, opened 2021-06-11
Implement State.Set, State.SetAttribute, Plan.Set, Plan.SetAttribute

### LINKED ISSUE #16: Figure out how to expose config/state/plan values.
### Module version

```
N/A
```

### Use-cases

We're going to need a way for provider developers to obtain config, state, and plan data from the request types. We could expose these as `tftypes.Value`s, but we think we can come up with something less verbose and more intuitive to use.

kmoe:
This is the final piece that closes #16.

We don't need setters on `Config`.

### DIFF
--- a/plan.go
+++ b/plan.go
@@ -38,6 +38,58 @@ func (p Plan) GetAttribute(ctx context.Context, path *tftypes.AttributePath) (at
 	return attrType.ValueFromTerraform(ctx, attrValue)
 }
 
+// Set populates the entire plan using the supplied Go value. The value `val`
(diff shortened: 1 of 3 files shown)


paddycarver (approved):
Seems right to me! The GoDoc comments may want tweaked, but I think I want to propose that we do that more holistically as a concerted effort of its own, rather than piecemeal. (And not just for this PR, either.)

But I don't see any reason to block other code paths on that, so I'd say let's get this merged, and we can refine and tweak documentation in separate PRs.

PR #52, opened 2021-06-16
Sketch out a new attribute path package.

paddycarver:
Wanted to capture some ideas and see what an interface might look like,
so I started sketching out a new attribute path package.


paddycarver:
I think we want to separate out the concept of an AttributeMatcher and an AttributePath, which this conflates.

AttributePath is good for specifying a single, concrete attribute. It's used in diagnostics and to surface context to provider developers during validate and plan modification, for example.

AttributeMatcher is good for specifying the types of attributes something works on, and is what the `Wild*` methods in this implementation are trying to achieve. It's useful in things like validation, to be able to say "don't allow this to be set when any other attribute is set", etc.

I think separating those out into separate types/concepts would be helpful in keeping straight which we're talking about, as the above technically would allow provider developers to return wildcards as part of diagnostics, which we'd need to catch and throw an error on because Terraform doesn't support that. We'd also be at risk of accidentally passing wildcard paths to provider developers as part of validation, which is less than helpful.

paddycarver:
AttributePaths also likely want a way to be navigated. For example, a `Parent()` method that returns the parent of the attribute that the AttributePath is pointing to is something that is useful in most situations. A `ChildOf` method that takes an AttributePath and returns whether the AttributePath it's being called on begins with AttributePath or not would also likely be useful.

bflad:
It's only taken a few months of refactoring to catch up to this 😂 but we finally got to a place where we wanted with native attribute path handling: https://github.com/hashicorp/terraform-plugin-framework/pull/390

Next up is migrating Config/Plan/State.Raw and separately on path expression handling!

PR #59, opened 2021-06-24
Update README.

paddycarver:


README.md
```diff
@@ -2,22 +2,36 @@
...
+terraform-plugin-framework is a module for building Terraform providers built on [terraform-plugin-go](https://github.com/hashicorp/terraform-plugin-go). It aims to provide as much of the power, predictability, and versatility of terraform-plugin-go as it can while abstracting away implementation details and repetitive, verbose tasks.
+
+## Status
+
+terraform-plugin-framework is still experimental. We are committed to moving forward with the module, but cannot guarantee any of its interfaces will not change as long as it is in version 0. We're waiting for more feedback, usage, and maturity before we're comfortable committing to APIs with the same years-long support timelines that [terraform-plugin-sdk](https://github.com/hashicorp/terraform-plugin-sdk) brings.
```
aareet, line 9: we created a fancy badge for experimental things for k8s- alpha - https://camo.githubusercontent.com/8ad47215ae8b556345c074d2636cdf5e8a7f54068c110d1a1795501b43fab52e/68747470733a2f2f696d672e736869656c64732e696f2f62616467652f7374617475732d6578706572696d656e74616c2d454141413332

README.md
```diff
@@ -2,22 +2,36 @@
...
+terraform-plugin-framework is a module for building Terraform providers built on [terraform-plugin-go](https://github.com/hashicorp/terraform-plugin-go). It aims to provide as much of the power, predictability, and versatility of terraform-plugin-go as it can while abstracting away implementation details and repetitive, verbose tasks.
```
aareet, line 5: > building Terraform providers built on [terraform-plugin-go](https://github.com/hashicorp/terraform-plugin-go)
```suggestion
terraform-plugin-framework is a module for building Terraform providers. It is built on [terraform-plugin-go](https://github.com/hashicorp/terraform-plugin-go). It aims to provide as much of the power, predictability, and versatility of terraform-plugin-go as it can while abstracting away implementation details and repetitive, verbose tasks.
```
Suggest reversing the order @

README.md
```diff
@@ -2,22 +2,36 @@
...
 
 We recommend only using tagged releases of this module, and examining the CHANGELOG when upgrading to a new release. Breaking changes will only be made in minor versions; patch releases will always maintain backwards compatibility.
 
 We welcome and appreciate issues and PRs discussing both the design and implementation of this module.
 
 ## Terraform CLI Compatibility
 
-Plugins built with this framework are only compatible with Terraform versions above 0.12.0. Terraform 0.12.26 or higher is needed to run acceptance tests for providers built with this framework.
+Plugins built with this framework are only compatible with Terraform versions above v0.15.0. Terraform v0.15.4 or higher is needed to run acceptance tests for providers built with this framework.
```
aareet, line 21: How about we simply pin at v0.15.4 and above?

aareet (changes requested):

README.md
paddycarver: I just don't want to mislead people, because Terraform users 0.15.0 and higher will get the provider when `terraform init` is run, and I don't want someone to rely on it being only 0.15.4 users and above, if that makes sense.

bflad: Could we also simplify compatibility promises and say that 1.0.0 and above is the tested CLI version since it provides its own compatibility promises, even if it is slightly off? Might be easier to centralize on that nice round major milestone than nitpick on 0.x versions.

paddycarver: It's just that if we say it's 1.0+, we need to wait until June of 2022 for our official providers to adopt this. If we say 0.15.0, they can adopt it in April. (Plus or minus some wiggle room here and there because of major versions, etc etc)

README.md
```diff
@@ -2,22 +2,38 @@
...
+
+## Status
+
+![experimental](https://camo.githubusercontent.com/8ad47215ae8b556345c074d2636cdf5e8a7f54068c110d1a1795501b43fab52e/68747470733a2f2f696d672e736869656c64732e696f2f62616467652f7374617475732d6578706572696d656e74616c2d454141413332)
+
+terraform-plugin-framework is still experimental. We are committed to moving forward with the module, but cannot guarantee any of its interfaces will not change as long as it is in version 0. We're waiting for more feedback, usage, and maturity before we're comfortable committing to APIs with the same years-long support timelines that [terraform-plugin-sdk](https://github.com/hashicorp/terraform-plugin-sdk) brings.
+
+terraform-plugin-framework is also not at full feature parity with terraform-plugin-sdk yet. Notably, it doesn't offer support for validation, modifying plans (including marking resources as needing to be recreated), importing resources, or upgrading resource state. We plan to add these features soon.
```
bflad, line 13: Should we make this a list and link to the relevant issues?

bflad (approved):
One non-blocking question, otherwise looks good to me 🚀

README.md
paddycarver: I'd like to take some more time to gather that properly, I think. Linking to the GitHub issues for the things I mention here, though, sounds like a good idea to me.

Issue #34, opened 2021-06-01
Support customizing plans


paddycarver:
### Module version

```
N/A
```

### Use-cases

Provider developers sometimes need to massage the plans that Terraform generates for them, squashing differences in bytes that are not semantically distinct, or supplying default values that are known at plan time. We should accept an optional method on resources that supports modifying the plan that Terraform generates.

### Attempted Solutions

N/A

### Proposal

Detect a ModifyPlan method when defined on resources and call it as part of our PlanResourceChange handling, to let provider developers modify the plan.

PR #64, opened 2021-07-07
Plan modification design doc

kmoe:
The first step towards #34.

I'm not fully happy with the `RequiresReplaceIf` part of the proposal and am looking for feedback. The main puzzle is how to keep verbosity low in the provider code while keeping the interface flexible enough.

### DIFF
--- a/docs/design/plan-modification.md
+++ b/docs/design/plan-modification.md
@@ -376,7 +376,7 @@ func RequiresReplaceIf(f RequiresReplaceIfFunc, description markdownDescription 
   }
 }
 
-type RequiresReplaceIfFunc func(context.Context, state, config attr.Value) bool
+type RequiresReplaceIfFunc func(context.Context, state, config attr.Value) (bool, error)
 
 type RequiresReplaceIfModifier struct {
   f RequiresReplaceIfFunc
@@ -410,16 +410,16 @@ func (f fileResourceType) GetSchema(_ context.Context) (schema.Schema, []*tfprot
 				Type:     types.NumberType,
 				Required: true,
 				PlanModifiers: schema.AttributePlanModifiers{
-				  RequiresReplaceIf(func(ctx context.Context, state, config attr.Value) bool {
+				  RequiresReplaceIf(func(ctx context.Context, state, config attr.Value) (bool, error) {
 				    stateVal := state.(types.Number)
 				    configVal := config.(types.Number)
 				  
 				    if !stateVal.Unknown && !stateVal.Null && !configVal.Unknown && !configVal.Null {
 				      if configVal.Value.Cmp(stateVal.Value) > 0 {
-				        return true
+				        return true, nil
 				      }
 				    }
-				    return false
+				    return false, nil
 				  }),
 				  CustomModifier,
 				  // ...
@@ -445,6 +445,47 @@ Similarly to option 3, the plan modifier functions are easily unit testable, and
 This solution aims to improve on the compatibility of option 3 by ensuring that no further fields need be added to `schema.Attribute` in future for the purposes of plan modification.
 
 If we were to change the fields of `ModifyResourcePlanRequest` to allow access to the full plan, state, and config (so that the result of the modify plan function could depend on the planned value of another attribute, for example), it would come at the cost of verbosity in the user-defined `RequiresIfFunc`, since the user must now retrieve the attribute values from the full `Config` and `State` rather than having them supplied as `attr.Value`s in the function arguments.
+
+### 4a. `attr.ValueAs()`
+
+This option is to be considered an extension to option 4.
+
+Plan modifier functions (`AttributePlanModifier.Modify()`), and helper functions such as `RequiresReplaceIf`, are supplied `attr.Value` arguments from which the provider code must derive attribute values. In the above example, this is done with type assertions:
+
+```go
+RequiresReplaceIf(func(ctx context.Context, state, config attr.Value) (bool, error) {
+  stateVal := state.(types.Number)
+  configVal := config.(types.Number)
+  // ...
+}
+```
+
+Avoiding such type assertions is one of the design goals of the new framework.
+
+Solution 4 should therefore be considered in combination with a way of avoiding such type assertions in provider code. One such proposal, which will be included in a separate design doc, is an `attr.ValueAs()` function:
+
+```go
+func ValueAs(ctx context.Context, val attr.Value, target interface{}) error {}
+```
+
+Similar to functions such as `state.Get(ctx, val, target)`, this function could use reflection to populate `target` with the value in `val`, returning an error if `target` is not of a compatible type. The provider code could then be:
+
+
+```go
+RequiresReplaceIf(func(ctx context.Context, state, config attr.Value) (bool, error) {
+  var stateInt int
+  var configInt int
+  err: = attr.ValueAs(ctx, state, stateInt)
+  if err != nil {
+  	return false, err
+  }
+    err: = attr.ValueAs(ctx, config, configInt)
+  if err != nil {
+  	return false, err
+  }
+  // ...
+}
+```
 
 ### 5. `attr.TypeWithModifyPlan`
 
--- a/docs/design/plan-modification.md
+++ b/docs/design/plan-modification.md
@@ -293,7 +293,7 @@ func RequiresReplaceIfSet(f RequiresReplaceFunc) RequiresReplaceFunc {
 
 Note that this would require the addition of `IsNull()` and `IsUnknown()` functions to the `attr.Value` interface, since there is at present no way to determine whether a generic `attr.Value` is null or unknown.
 
-### 4. `schema.Attribute.PlanModifier`
+### 4. `schema.Attribute.PlanModifiers`
 
 Extending the abstraction of `RequiresReplaceIf` one level higher, we can add a `PlanModifiers` field on the `schema.Attribute` struct, with the following framework code:
 
@@ -464,4 +464,4 @@ Without knowing how custom types will be used by provider developers, this optio
 
 ## Recommendations
 
-We recommend implementing `schema.Attribute.PlanModifier`, and the `ResourceWithModifyPlan` interface. Composition, `attr.TypeWithModifyPlan`, and other helpers can be implemented as required.
+We recommend implementing `schema.Attribute.PlanModifiers`, and the `ResourceWithModifyPlan` interface. Composition, `attr.TypeWithModifyPlan`, and other helpers can be implemented as required.
--- a/docs/design/plan-modification.md
+++ b/docs/design/plan-modification.md
@@ -1,0 +1,298 @@
+# Modifying Plans
+
+The Terraform [resource instance change lifecycle](https://github.com/hashicorp/terraform/blob/3b6c1ef156387e6652ed8516bce2af2422b7fcf1/docs/resource-instance-change-lifecycle.md) involves the following sequence of RPCs:
+ 1. `ValidateResourceTypeConfig`
+ 1. `PlanResourceChange`
+ 1. `ApplyResourceChange`
+
+The plugin framework handles requests and responses for these RPCs, allowing providers to hook in to `ValidateResourceTypeConfig` via validation helpers (https://github.com/hashicorp/terraform-plugin-framework/issues/17), and to `ApplyResourceChange` via resource CRUD functions. This design document concerns the ways we allow provider developers to hook into the `PlanResourceChange` RPC, giving providers control over the plan rendered to users via the Terraform CLI.
+
+## The `PlanResourceChange` RPC
+
+```proto
+message PlanResourceChange {
+    message Request {
+        string type_name = 1;
+        DynamicValue prior_state = 2;
+        DynamicValue proposed_new_state = 3;
+        DynamicValue config = 4;
+        bytes prior_private = 5; 
+        DynamicValue provider_meta = 6;
+    }
+
+    message Response {
+        DynamicValue planned_state = 1;
+        repeated AttributePath requires_replace = 2;
+        bytes planned_private = 3; 
+        repeated Diagnostic diagnostics = 4;
+    }
+}
+```
+
+The `PlanResourceChange` RPC request contains `config`, `prior_state` and `proposed_new_state` values, from which the provider is required to determine and return the `planned_state` in the response.
+
+Terraform CLI renders a diff between `prior_state` and `planned_state` to the user for confirmation.
+
+The `PlanResourceChange` RPC response also contains a list of attribute paths: `repeated AttributePath requires_replace`. For each attribute path in this list, Terraform checks whether its value has changed, and if it has, then the user is shown a plan saying that the resource instance must be replaced, with the attribute in question marked with `# forces replacement`.
+
+In allowing providers to control the `PlanResourceChange` response, i.e. "modify the plan", the plugin framework therefore enables providers not only to modify the diff that will be displayed to the user (and ultimately applied), but also to branch into a destroy-and-create lifecycle phase, triggering other RPCs.
+
+Plan modification has two distinct use cases for providers:
+  - Modifying plan values, and
+  - Forcing resource replacement.
+
+This design document therefore distinguishes "ModifyPlan" from "RequiresReplace" behaviour, the former being a superset of the latter.
+
+## History: `ForceNew` and `CustomizeDiff`
+
+In `helper/schema`, there are two ways, both somewhat indirect, that a provider can customise the plan.
+
+### `ForceNew`
+
+`ForceNew` is a boolean schema behaviour that marks a resource field as requiring the resource to be destroyed and recreated if the value of the field is changed. Schema example:
+```go
+"base_image": {
+  Type:     schema.TypeString,
+  Required: true,
+  ForceNew: true,
+},
+```
+After receiving the `PlanResourceChange` request, the SDK determines whether any change was made to a field marked as `ForceNew`, and if so, adds it to the returned `RequiresNew` array.
+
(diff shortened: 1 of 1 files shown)


docs/design/plan-modification.md
```diff
@@ -0,0 +1,298 @@
...
+  ForceNew: true,
+},
+```
+After receiving the `PlanResourceChange` request, the SDK determines whether any change was made to a field marked as `ForceNew`, and if so, adds it to the returned `RequiresNew` array.
+
+The SDK also executes logic at resource validation time (`InternalValidate`) to enforce the condition that if a resource does not have an Update function, all non-Computed attributes must have ForceNew set; and that if all fields are ForceNew or Computed without Optional, Update must _not_ be defined.
+
+### `CustomizeDiff`
+
+Rather than exposing Terraform plans to provider developers, `helper/schema` has as a first-class concept the _resource diff_. Providers can optionally define a `CustomizeDiff` method on the `Resource` struct, which resembles a CRUD method, except that instead of `ResourceData` the function is supplied a `*ResourceDiff`, and is called during several points in the resource lifecycle, and must therefore be "resilient to support all scenarios".
```
ewbankkit, line 66:
```suggestion
Rather than exposing Terraform plans to provider developers, `helper/schema` has as a first-class concept the _resource diff_. Providers can optionally define a `CustomizeDiff` method on the `Resource` struct, which resembles a CRUD method, except that instead of `ResourceData` the function is supplied a `ResourceDiff`, and is called during several points in the resource lifecycle, and must therefore be "resilient to support all scenarios".
```

bflad (commented):
Overall this is looking really good -- I'm curious about a few points here, which happy to discuss offline:

* Are documentation hooks desired for this functionality now or in the future?
* Are there concerns with proliferating fields on attributes?
* Is it important to allow attribute-level plan modifications access to the provider?

docs/design/plan-modification.md
```diff
@@ -0,0 +1,298 @@
...
+
+#### Tradeoffs
+
+A single `RequiresReplace` schema attribute is easily discoverable, Go-native, minimally verbose, and a reasonably transparent representation of the protocol's concept of `RequiresReplace`.
+
+A cursory survey of existing provider code finds `ForceNew` very widely used, and seldom the subject of individual acceptance tests, likely because provider developers do not feel the need to test a behaviour whose implementation resides in the SDK itself. For this reason the question of unit testability is also moot. 
+
+One possible disadvantage of this approach is that it is atomic with respect to compatibility - unlike provider-defined plan modification functions like those described below, the framework must deprecate the field or undergo a breaking change in order to modify the behaviour of `schema.Attribute.RequiresReplace`.
```
paddycarver, line 183: We should probably also call out here that `ForceNew`, while widely useful, also isn't sufficient in a lot of cases, like when conditionally marking a field as requiring the resource to be replaced. A drawback of this approach is that it is _declarative_, and so there's no provider control over when or how it's applied; it's either always ForceNew or never.

docs/design/plan-modification.md
```diff
@@ -0,0 +1,298 @@
...
+type fileResourceType struct{}
+
+// GetSchema returns the schema for this resource.
+func (f fileResourceType) GetSchema(_ context.Context) (schema.Schema, []*tfprotov6.Diagnostic) {
+	return schema.Schema{
+		Attributes: map[string]schema.Attribute{
+			"favorite_number": {
+				Type:     types.NumberType,
+				Required: true,
+				RequiresReplaceIf: func(ctx context.Context, old, new attr.Value) bool {
```
paddycarver, line 211: nitpick: I'd very much like to stay away from "old" or "new" being parameter names. First, `new` is a keyword in Go, and that feels like asking for trouble. Second, `old` and `new` _what_? `state` and `config` may be better names.

docs/design/plan-modification.md
```diff
@@ -0,0 +1,298 @@
...
+```go
+type TypeWithRequiresReplace interface {
+  Type
+  
+  RequiresReplaceIf(context.Context, old, new Value) bool
+```
+
+This would allow bundling reusable RequiresReplace behaviour up with a custom type's validation and other behaviours.
+
+Without knowing how custom types will be used by provider developers, this option seems premature, and makes less sense than bundling validation functions with custom types.
```
paddycarver, line 257: I think bundling a `ModifyPlan` with the type makes sense here. `RequiresReplaceIf` less so. For example, if we're implementing a timestamp resource, a `ModifyPlan` method would allow it to squash semantically meaningless diffs, rather than making provider developers do that everywhere the type is used.

docs/design/plan-modification.md
```diff
@@ -0,0 +1,298 @@
...
+
+Similarly to `helper/schema`, the `All` composition helper runs all `RequiresReplaceFunc`s and returns true only if all funcs return true.
+
+```go
+func All(funcs ...RequiresReplaceFunc) bool {}
+```
+
+#### `Sequence()`
+
+Similarly to `helper/schema`, the `Sequence` composition helper runs all `RequiresReplaceFunc`s in sequence, stopping at the first that returns false.
```
paddycarver, line 271: Should this be stopping at the first that returns true? I'd expect `All` would stop at the first `false` because it would know that all funcs aren't returning true....

docs/design/plan-modification.md
```diff
@@ -0,0 +1,298 @@
...
+  return func (ctx context.Context, old, new attr.Value) bool {
+    if old.IsNull() || old.IsUnknown() || new.IsNull() || new.IsUnknown() {
+      return false
+    }
+    return f(ctx, old, new)
+  }
+}
+```
+
+Note that this would require the addition of `IsNull()` and `IsUnknown()` functions to the `attr.Value` interface, since there is at present no way to determine whether a generic `attr.Value` is null or unknown.
```
paddycarver, line 294: I feel like this should be called out at a higher level. The reusability of these functions is incredibly limited if they receive an `attr.Value`, because you can't really get to the data in a meaningful way. So helpers probably wouldn't want to accept `attr.Value`, because they can't _do_ anything with it. They'd probably want a `tftypes.Value`. But provider developers writing one-offs or resource-specific ones or `attr.Type` implementations (maybe??) would probably want the easier-to-use `attr.Value` rather than the `tftypes.Value`.

We may want to explore how to address these competing concerns.

docs/design/plan-modification.md
```diff
@@ -0,0 +1,298 @@
...
+
+## Recommendations
+
+We recommend implementing `schema.Attribute.RequiresReplace`, `schema.Attribute.RequiresReplaceIf`, and the `ResourceWithModifyPlan` interface. Composition and other helpers can be implemented as required.
```
paddycarver, line 298: I think `RequiresReplaceIf` is on the wrong side of the "usefulness" vs "limitations" barrier for me. I think RequiresReplace makes sense given its utility, and I think the `PlanModifier` @bflad proposed above makes sense (I've actually had that design on my whiteboard for a few months, though I called it `PlanTransformer`), but I think:

1) `RequiresReplaceIf` can easily be a `PlanModifier` helper.
2) Including `RequiresReplaceIf` is going to lead to people who can't do what they want because of its constraints to request other variants be added.

I think the `PlanModifier` approach makes sense because it can meet the needs of `RequiresReplaceIf` and _also_ replace DiffSuppressFuncs.

paddycarver (commented):
Left some thoughts inline. I didn't see a convenient place to put it in, but I'm also interested in how this may impact the docs generator / language server and how we may be able to surface these behaviors to those tools, and how these possible paths forward interact with our capabilities there.

docs/design/plan-modification.md
```diff
@@ -0,0 +1,298 @@
...
+        repeated Diagnostic diagnostics = 4;
+    }
+}
+```
+
+The `PlanResourceChange` RPC request contains `config`, `prior_state` and `proposed_new_state` values, from which the provider is required to determine and return the `planned_state` in the response.
+
+Terraform CLI renders a diff between `prior_state` and `planned_state` to the user for confirmation.
+
+The `PlanResourceChange` RPC response also contains a list of attribute paths: `repeated AttributePath requires_replace`. For each attribute path in this list, Terraform checks whether its value has changed, and if it has, then the user is shown a plan saying that the resource instance must be replaced, with the attribute in question marked with `# forces replacement`.
```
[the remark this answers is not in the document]
kmoe: Terraform checks whether the value has changed here: https://github.com/hashicorp/terraform/blob/aaaeeffa81857c41c05d81b2829a063f708abafe/internal/terraform/node_resource_abstract_instance.go#L839

kmoe:
Thanks, all. Yes, `PlanModifier` (whatever we call it) is the missing piece here! Revising and dealing with all comments.

docs/design/plan-modification.md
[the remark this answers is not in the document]
paddycarver: Well, I'll be! Thanks, I was trying to get at that from the SDK and I should've just checked with Terraform. I could've sworn we could force things to be ForceNew using ResourceDiff, but I suppose that just set them to Unknown, which could then force new. Neat!

docs/design/plan-modification.md
kmoe: `RequiresReplaceIfSet` is as reusable as `RequiresReplaceIf` - I think the `RequiresReplaceIfFunc`s will become too verbose if we make the `state, config` arguments `tftypes.Value`s. Do you have an example in mind where a provider developer would want to implement a `RequiresReplaceIfFunc` that is generic with respect to the attribute type? If not, why would they not be able to do anything with the `attr.Value`s?

docs/design/plan-modification.md
kmoe: The signature is `type CustomizeDiffFunc func(context.Context, *ResourceDiff, interface{}) error` - any particular reason to omit the pointer here?

docs/design/plan-modification.md
kmoe: I mention in the above section that it is declarative. Added another short paragraph elaborating that.

docs/design/plan-modification.md
```diff
@@ -0,0 +1,298 @@
...
+
+In allowing providers to control the `PlanResourceChange` response, i.e. "modify the plan", the plugin framework therefore enables providers not only to modify the diff that will be displayed to the user (and ultimately applied), but also to branch into a destroy-and-create lifecycle phase, triggering other RPCs.
```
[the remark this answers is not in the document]
kmoe: I'm using "RPC"s here to refer to the RPCs defined in the Terraform protocol itself. I think considering "destroy-and-create" its own RPC in this context would be confusing.

kmoe:
Hmm, actually I really like the typed validators in https://github.com/hashicorp/terraform-plugin-framework/pull/65 but I didn't dare propose a `StringAttributePlanModifier`. Since it looks like our proposed solutions for plan customisation and validation are converging, I think it's worth considering offering the same kind of strong typing for both.

docs/design/plan-modification.md
```diff
@@ -0,0 +1,467 @@
...
+```
+
+This would allow bundling reusable `ModifyPlan` behaviour up with a custom type's validation and other behaviours. This could be useful, for example, in a custom timestamp type to squash semantically meanignless diffs, so provider developers do not have to specify the attribute plan modifier wherever the attribute appears in a schema.
+
+Without knowing how custom types will be used by provider developers, this option seems premature, and makes less sense than bundling validation functions with custom types.
+
+
+## Recommendations
+
+We recommend implementing `schema.Attribute.PlanModifier`, and the `ResourceWithModifyPlan` interface. Composition, `attr.TypeWithModifyPlan`, and other helpers can be implemented as required.
```
paddycarver:
```suggestion
We recommend implementing `schema.Attribute.PlanModifiers`, and the `ResourceWithModifyPlan` interface. Composition, `attr.TypeWithModifyPlan`, and other helpers can be implemented as required.
```

paddycarver (commented):
I'm in agreement with your conclusions. I am still unhappy about the type assertions going on with `attr.Value`s. I think it limits our composability of these blocks significantly.

What if we offer an `attr.ValueAs` helper that does the same reflection `Get` does (and maybe `attr.ValueIsKnown` / `attr.ValueIsNull` helpers that use the `ToTerraformValue` to compare to `nil` and `tftypes.UnknownValue`) that will allow us to use `attr.Value`s without limiting the types of the values? Then we offer the same way to get as values that CRUD uses, for example.

docs/design/plan-modification.md
```diff
@@ -0,0 +1,543 @@
...
+
+type ModifyAttributePlanRequest struct {
+	// ResourceConfig is the configuration the user supplied for the resource.
+	ResourceConfig Config
+
+	// ResourceState is the current state of the resource.
+	ResourceState State
+
+	// ResourcePlan is the planned new state for the resource.
+	ResourcePlan Plan
```
paddycarver, line 351: You added these here, and then below say they're not available. I think because I made things confusing with the last review.

My hot take is let's omit them initially, but know that we can add them later, if we want, we're just waiting for a reason to. But if you feel they should be in from the get-go, I am fine with that answer to, so this is not a merge-blocking comment.

docs/design/plan-modification.md
```diff
@@ -0,0 +1,543 @@
...
+```go
+type TypeWithModifyPlan interface {
+  Type
+  
+  ModifyPlan(context.Context, ModifyAttributeTypePlanRequest, *ModifyAttributeTypePlanResponse) bool
+```
+
+This would allow bundling reusable `ModifyPlan` behaviour up with a custom type's validation and other behaviours. This could be useful, for example, in a custom timestamp type to squash semantically meaningless diffs, so provider developers do not have to specify the attribute plan modifier wherever the attribute appears in a schema.
+
+Without knowing how custom types will be used by provider developers, this option seems premature, and makes less sense than bundling validation functions with custom types.
```
paddycarver, line 538: I disagree, I think it makes as much sense, but also am fine with adding it later given we've called out the design and use case here and acknowledged that the only reason we're not doing it is we're waiting for a use case to present itself, and not because we think the design is flawed or problematic or something.

paddycarver (approved):
A couple minor comments, but it seems solid and reasonable to me, and the comments are things it's reasonable enough to change later, I think, so this has my 👍 🚀

docs/design/plan-modification.md
kmoe: Thanks, I meant to remove the bit that said they're not available. Let's see what @bflad thinks about having them in the initial version.

docs/design/plan-modification.md
```diff
@@ -0,0 +1,543 @@
...
+
+The fields in the `ModifyAttributePlanRequest` and `ModifyAttributePlanResponse` struct are not the same as those in `ModifyResourcePlanRequest` and `ModifyResourcePlanResponse` from option 1. In particular, it is not possible to change the planned value of _other_ attributes inside an attribute's `PlanModifier`. If it were, it would be possible for two or more attributes to have `PlanModifier`s modifying each other's planned values, with no clear indication of the order in which those operations would be performed. 
+
+Framework documentation should make it clear that `schema.Attribute.PlanModifier` is scoped to a single attribute with no access to other attribute values or API requests. If either of these is needed, provider developers should use `tfsdk.Resource.ModifyPlan()`.
```
bflad, line 476: For my own personal edification, could we theoretically support API requests here as well by extending the interface types (such as a `WithProvider` variant, so developers can strongly type the provider in their own function signatures) or extending the request struct with a `Provider tfsdk.Provider` field (this would unfortunately require type assertions)?

docs/design/plan-modification.md
bflad: I'm 👍 (which I think we all are!) for having custom type design be its own discussion and I don't think this design's recommendations are particularly preventing those efforts.

docs/design/plan-modification.md
```diff
@@ -0,0 +1,543 @@
...
+
+`CustomizeDiff` is also used to implement multi-attribute validation, e.g. checking that at most one of two particular attributes is set. The present framework, on the other hand, separates plan modification from validation, and the solutions proposed below are concerned only with plan modification.
+
+### `helper/customdiff`
+
+The legacy SDK provides a set of reusable and composable helper functions in its `customdiff` package, intended to reduce the need for long and complex logic to reside in a single `CustomizeDiff` function in provider code. This small package is best summarised in its [documentation](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/customdiff).
+
+## Solution options
+
+### 1. `tfsdk.Resource.ModifyPlan()`
```
bflad, line 94: Is another option here for providers to return a slice of plan modifiers that match an interface type? Then they can be composed together and sequencing is implicit. For example:

```go
type ResourcePlanModifier interface {
    Description(context.Context) string
    MarkdownDescription(context.Context) string
    ModifyPlan(context.Context, ModifyResourcePlanRequest, *ModifyResourcePlanResponse)
}

type ResourcePlanModifiers []ResourcePlanModifier

type ResourceWithPlanModifiers interface {
    Resource

    PlanModifiers(context.Context) ResourcePlanModifiers
}
```

This will allow the same documentation hooks provided by the `schema.Attribute.PlanModifiers` and let provider developers be a little more declarative:

```go
type customResource struct{}

type (r customResource) PlanModifiers(_ context.Context) ResourcePlanModifiers {
    return ResourcePlanModifiers{
        Modifier1,
        Modifier2,
    }
}
```

bflad (approved):
Overall very well done!

One thing I'm curious about is how the provider instance will be available to this functionality if necessary, but it is fine to leave that as an implementation detail if the design does not inhibit that possibility.

docs/design/plan-modification.md
bflad: Naming aside (would data sources have separate fields?), since this is just a design document my recommendation would be to fix the verbiage about availability. I think abandoned ideas or future ideas do have some importance to capture since those can be representative of everything the team was thinking about and aware of at the time of design -- whether and when those designs become reality is more of an implementation and prioritization question.

Issue #17, opened 2021-05-07
Support validation


paddycarver:
### Module version

```
N/A
```

### Use-cases

Provider developers need a way to validate the values that practitioners enter in their configs, a way of hooking into the `ValidateResourceType` RPC. We should provide a hook into that and potentially a library of common helpers.

bflad:
This has been covered by:

- #65 
- #75 
- #82 

With some minor additional unit testing coverage being followed up in #92.

PR #65, opened 2021-07-09
docs/design: Initial Validation design document

bflad:

Covering background, prior implementations, and goals. Further updates to this document will outline proposals and ultimately recommendations.


docs/design/validation.md
```diff
@@ -0,0 +1,2048 @@
...
+// While it is generally preferred to use the typed validation interfaces,
+// such as StringValueValidator, this interface allows custom implementations
+// where the others may not be suitable. The Validate function is responsible
+// for protecting against attr.Value type assertion panics.
+type GenericValueValidator interface {
+    ValueValidator
+    Validate(context.Context, *tftypes.AttributePath, attr.Value) error
+}
+
+// StringValueValidator is an interface type for implementing String value validation.
```
kmoe, line 1883: `StringValueValidator` has great UX - no type assertions needed in the provider code!

How would this work with complex types, e.g. `types.List`? Since there are as many list types as there are element types, it seems that the options are either:
1. Create a single `ListValueValidator` interface, and require type assertions on the list elements inside the `Validate` function,
2. Create multiple `ListOfStringsValueValidator`, etc, interfaces, accepting that not all cases can be covered but trying to provide helpers for the most common ones, or
3. Do not create strongly typed validator interfaces for complex types.

kmoe:
Really well structured design doc. One option I'd like to see considered is the request/response pattern for the parameters in the validation functions. All other things being equal (which they likely aren't), it might be good to use that pattern consistently across CRUD, plan modification, and validation.

docs/design/validation.md
```diff
@@ -0,0 +1,2048 @@
...
+      ConflictingAttributesWithValues(*tftypes.AttributePath, ValueValidator, *tftypes.AttributePath, ValueValidator),
+      PrerequisiteAttribute(*tftypes.AttributePath, *tftypes.AttributePath),
+      PrerequisiteAttributeWithValue(*tftypes.AttributePath, ValueValidator, *tftypes.AttributePath),
+    }
+}
+```
+
+This setup would allow for the framework to provide flexible resource level validation with a low amount of friction for provider developers. Helper functions would be extensible and make the behaviors clear.
+
+#### Defining Attribute Validation Functions
```
paddycarver, line 1720: What's an attribute validation function, and how is it different from a multiple attribute validation function or a single attribute validation function? I'm struggling to keep track of the distinctions you're introducing here. So far I've got:

* single attribute value validations
  * I _think_ this is called value validation, sometimes?
* multiple attribute value validations
  * I _think_ this is called path validation, sometimes?
* attribute validations

Is there any way to make the distinction between these concepts clearer?

paddycarver (commented):
I'd love to see options outlined for:

* having a `Validate` method on the resource/provider/data source that allows for validation to be implemented inline at the resource level, instead of needing to return validator interfaces.
* Using the request and response pattern for these validation interfaces.

I have concerns about the strongly typed function signatures. I think that may hurt discoverability, and I think it's going to be an absolute mess when complex types get involved. One option that's raised in #64 is to have an `attr.ValueAs` helper that works like `Get`, meaning we could use the same plan/config/state access as the rest of the framework _and_ be consistent with plan modification (if it goes that route). It may be worth exploring that here, too.

I have concerns with the typed error returns and using errors in general when the rest of the framework is diagnostics. It seems to prioritize sharing information with callers, but we're the only callers, and I think the primary audience for that error information is the practitioner. I'm worried this is either going to be too restrictive for provider developers to offer a good experience, or add cognitive load on provider developers that we can avoid. I'm interested in hearing how you feel about returning diagnostics like we do everywhere else, and iterating on the interface to and design of diagnostics to try and reap some of the benefits you're identifying.

I'm having some trouble keeping track of the distinctions you're introducing here. It may be easier to track if we collapse some of the sections--for example, we may not need to talk about how something is set on a Schema and what its type/type signature is separately. I think it's okay to pick one example when showing it on the schema, and then just show alternate types/type signatures. It may reduce the number of headings and the level of nesting in the headings, which gets hard to follow. But mostly, the first time you refer to something (say, "PathValidation") it may be helpful to just note what makes it distinct from the other, similar concepts you've introduced.

I think this is great, a lot of context, and you've covered the problem space really well. I'd love to see the other solutions I mentioned considered and their benefits/tradeoffs weighed. I'd also like to see some more information about how complex types would be handled in your proposed strongly-typed helpers.

docs/design/validation.md
bflad: In my head I was thinking in the area of 2 and leaving an escape hatch interface (3). 1 could make 2 slightly easier to do, so might be beneficial.

I'm also curious though of what validation space exists here -- in the previous framework there is not much available validation for complex types (e.g. you can only validate the number of elements in a list/set/map). There have been very odd cases of needing a list type for ordering, but needing element uniqueness (like a set).

I think my focus would be covering the very common string/number cases then allowing those more odd cases to bubble up.

docs/design/validation.md
```diff
@@ -0,0 +1,2048 @@
...
+    }
+
+    // ... rest of logic ...
+```
+
+This proposal would strongly encourage all implementations to handle value type conversions since using a stronger type in the function signature would risk panics that the framework cannot prevent and the compiler cannot check. Any error handling here could become inconsistent across implementations. This type conversion logic feels like an unnecessary burden on implementors and could reduce the developer experience as this logic would always need to be repeated with little to no actual utility.
+
+##### `types.T` Type
+
+If using an `attr.ValueValidator` interface approach, multiple new Go interface types could be created that define extensible value validation functions with strong typing. For example:
```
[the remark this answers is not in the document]
bflad: Upfront, it doesn't have to if we provide a "generic" validator that has `attr.Value` as a parameter and potentially lean on the `ValueAs()` proposal. I'm talking a little bit more about this with Katy here: https://github.com/hashicorp/terraform-plugin-framework/pull/65#discussion_r670361078

docs/design/validation.md
```diff
@@ -0,0 +1,2048 @@
...
+```go
+// Validate performs all validation functions.
+//
+// Each type performs conversion or returns a conversion error
+// prior to executing the typed validation function.
+func (vs ValueValidators) Validate(ctx context.Context, path *tftypes.AttributePath, rawValue attr.Value) error {
+    for _, validator := range vs {
+        switch typedValidator := validator.(type) {
+        case StringValueValidator:
+            value, ok := rawValue.(types.String)
```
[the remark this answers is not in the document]
bflad: We haven't talked together about the design of custom attributes, but implicitly all values can (should?) boil down to their complex and primitive value types. Otherwise that custom attribute design could have to recommend switching value types to an interface, which seems outside this design to discuss or postulate on as it would have broader effect.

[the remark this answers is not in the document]
bflad: It seems like a type switch isn't the correct implementation detail in this scenario beyond what value types the framework implements -- custom value types would need to use a generic `attr.Value` (still provided as an escape hatch) and proposed `ValueAs` handling. This would really just serve to simplify framework validation helpers (or developers using `types.String` (for example).

docs/design/validation.md
```diff
@@ -0,0 +1,2048 @@
...
+This section includes examples and details with `schema.Attribute` implemented as a Go structure type as it exists today. Future design considerations around creating specialized or custom attribute types may warrant switching this to an interface type with separate concrete types.
+
+##### `ValueValidation` Field on `schema.Attribute`
+
+Similar to the previous framework, a new field can be added to the `schema.Attribute` type. For example:
+
+```go
+schema.Attribute{
+    // ...
+    ValueValidation: T,
```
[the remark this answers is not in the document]
bflad: Sigh, I don't know why I changed that from my original brainstorming. Should be fixed in 779f7a8ea89a8daaad987d72c5cb1ae17f9daadf, but please let me know if I missed anything.

docs/design/validation.md
```diff
@@ -0,0 +1,2048 @@
...
+This proposal colocates the value validation behaviors in the schema definition, meaning it is easier for provider developers to discover this type of validation and correlate the validation logic to the name and type information.
+
+##### `ValueValidations` Field on `schema.Attribute`
+
+A new field that accepts a list of functions can be added to the `schema.Attribute` type. For example:
+
+```go
+schema.Attribute{
+    // ...
+    ValueValidations: []T{
```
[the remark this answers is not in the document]
bflad: Sigh, I don't know why I changed that from my original brainstorming. Should be fixed in 779f7a8ea89a8daaad987d72c5cb1ae17f9daadf, but please let me know if I missed anything.

docs/design/validation.md
```diff
@@ -0,0 +1,2048 @@
...
+    value, ok := rawValue.(types.String)
+    
+    if !ok {
+        return fmt.Errorf("%s with incorrect type: %T", path, rawValue)
+    }
+
+    // ... rest of logic ...
+```
+
+This proposal would strongly encourage all implementations to handle value type conversions since using a stronger type in the function signature would risk panics that the framework cannot prevent and the compiler cannot check. Any error handling here could become inconsistent across implementations. This type conversion logic feels like an unnecessary burden on implementors and could reduce the developer experience as this logic would always need to be repeated with little to no actual utility.
```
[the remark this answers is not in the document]
bflad: I apologize for my poor wording there. Given the scenario that a developer wrote a validation like the below:

```go
func (v someValidator) Validate(ctx context.Context, path *tftypes.AttributePath, value types.String) error {
    // ... rest of logic ...
```

But injected it into the wrong attribute type:

```go
schema.Attribute{
    Type: types.NumberType,
    Validators: /* validation func from above */,
}
```

I _think_ the Go compiler would be fine with it since `types.String` satisfies the interface, but it would panic at runtime? I guess I should try it and reword this better.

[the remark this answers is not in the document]
bflad: Seems like the compiler catches this hypothetical, so I'll remove this.

```
./prog.go:99:22: cannot use StringLengthBetween(1, 5) (type stringLengthBetweenValidator) as type ValueValidator in slice literal:
	stringLengthBetweenValidator does not implement ValueValidator (wrong type for Validate method)
		have Validate(context.Context, *tftypes.AttributePath, types.String) error
		want Validate(context.Context, *tftypes.AttributePath, attr.Value) error
```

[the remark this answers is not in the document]
bflad: Fixed in be40423bbd992e667de26a69e9e97b726a2abc8a

docs/design/validation.md
```diff
@@ -0,0 +1,2048 @@
...
+        }
+    }
+
+    return nil
+}
+```
+
+In this scenario, it the implementor's responsibility to generate the appropriate diagnostic back, but they have full control of the output. It could be difficult for the framework to enforce implementation rules around these responses or potentially allow configuration overrides for them without creating more abstractions on top of this type or additional helper functions. Differing diagnostic implementations could introduce confusion for practitioners.
+
+In general, this proposal feels very similar to either the generic `error` type or typed error proposals above (depending on the implmentation details) with minimal utility over them beyond complete output customization.
```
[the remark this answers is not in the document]
bflad: Updated this in 37a249a0826aa23d585350cafc5580a94948d888, but forthcoming changes as part of request/response pattern will alter things further.

kmoe (commented):
A more thorough review prior to the request/response pattern additions (edit: which I see you've just added).

docs/design/validation.md
```diff
@@ -0,0 +1,2025 @@
...
+}
+```
+
+Terraform supports the following validation for provider implementations:
+
+- Provider configurations
+- Resource configurations and plans
+- Data Source configurations and plans
+
+Within these, there are two types of validation:
```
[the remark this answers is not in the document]
bflad: This is an excellent point! Noted in 41fbf96d756d5bf8bbdf434071e0c0dd6e14b20d

docs/design/validation.md
```diff
@@ -0,0 +1,2025 @@
...
+    Sensitive bool
+
+    // DeprecationMessage defines a message to display to practitioners
+    // using this attribute, warning them that it is deprecated and
+    // instructing them on what upgrade steps to take.
+    DeprecationMessage string
+}
+```
+
+Although later designs surrounding the ability to allow providers to define custom schema types may change this particular Go typing detail.
```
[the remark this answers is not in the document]
bflad: I received an educational session on this and it now makes much more sense to me! Removed in 47086a60ded91251f0f83d32fe834a84b2c64f8e

docs/design/validation.md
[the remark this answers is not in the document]
bflad: I think where we've landed on this is that, not as part of these proposals but potentially in the future, each type implementation (such as `types.String`, but others too!) could optionally offer helpers to simplify validation implementations to be strongly typed. The design in its current form does not prevent them and the details have been removed from the recommendations, instead left as a note within future considerations.

docs/design/validation.md
```diff
@@ -0,0 +1,1989 @@
...
+- Provider configurations
+- Resource configurations (and configurations during plans)
+- Data Source configurations (and configurations during plans)
+
+Within these, there are two types of validation:
+
+- Single attribute value validation (e.g. string length)
+- Multiple attribute validation (e.g. attributes or attribute values that conflict with each other)
+
+There is no difference between these types of validation to Terraform, as Terraform just works with errors and warnings being returned from providers, but the previous framework surfaced these different conceptually.
```
paddycarver, line 125: > these different conceptually

Maybe "these concepts differently"?

docs/design/validation.md
```diff
@@ -0,0 +1,1989 @@
...
+- Provider configurations
+- Resource configurations (and configurations during plans)
+- Data Source configurations (and configurations during plans)
+
+Including where possible:
+
+- Single attribute value validation (e.g. string length)
+- Multiple attribute validation (e.g. attributes or attribute values that conflict with each other)
+
+In terms of implementation, the following core concepts:
```
paddycarver, line 982: Feel like a verb is missing here. The following core concepts what?

docs/design/validation.md
```diff
@@ -0,0 +1,1989 @@
...
+
+- Low level primitives (e.g other portions of the framework, external packages, and provider developers can implement higher level functionality)
```
paddycarver, line 984: This sounds a bit like plugin-go to me. Is it fair to frame this instead as "composable building blocks"?

docs/design/validation.md
```diff
@@ -0,0 +1,1989 @@
...
+- Reusability between single attribute and multiple attribute validation functionality (e.g. attribute value functions)
+- Hooks for documentation (e.g. for future tooling such as provider documentation generators to self-document attributes)
+
+Finally, these other considerations:
```
paddycarver, line 988: Feel like a verb is missing here, too. These other considerations what?

docs/design/validation.md
```diff
@@ -0,0 +1,1989 @@
...
+
+#### Extending Plan Modifications
+
+The [Plan Modifications design documentation](./plan-modifications.md) outlines proposals which broadly replace the previous framework's `CustomizeDiff` functionality. See that documentation for considerations and recommendations there. In this proposal for validation, new functions for validation would be provided within that framework, rather than introducing separate handling.
+
+Implementing against that design could prove complex for the framework as they are intended to serve differing purposes. It could also be confusing for provider developers in the same way that `CustomizeDiff` was confusing where differing logical rules applied to differing attribute value and operation scenarios. Another wrinkle is that plan modifications are only intended to run during `terraform plan` (`PlanResourceChanges` RPC) and `terraform apply` (`ApplyResourceChanges` RPC), so the framework would be introducing its own additional logic to extract and perform any validation functions during the `terraform validate` (`ValidateDataSourceConfig`/`ValidateProviderConfig`/`ValidateResourceConfig` RPCs).
+
+#### New Handling
```
paddycarver, line 1007: "New Handling" confused me for a minute. Is something like "New Abstraction" clearer? 🤔

docs/design/validation.md
```diff
@@ -0,0 +1,1989 @@
...
+
+#### Validation Function Type
+
+New Go type(s) could be created that define the signature of a validation function, similar to the previous framework `SchemaValidateFunc`. For example:
+
+```go
+type ValidationFunc func(context.Context, ValidationRequest, *ValidationResponse)
+```
+
+To support passing through the provider instance to the function, the parameters would also need to include a `tfsdk.Provider` interface type:
```
paddycarver, line 1045: Technically it could be included in the request info, I think?

docs/design/validation.md
```diff
@@ -0,0 +1,1989 @@
...
+
+```go
+type ValidationFunc func(context.Context, provider tfsdk.Provider, ValidationRequest, *ValidationResponse)
+```
+
+The main drawback of this approach is that it does not allow for documentation hooks. This also drifts from other design decisions of the framework without providing much benefit for the differing implementation.
```
paddycarver, line 1051: A benefit that may be worth calling it is it's slightly less verbose to do one-off functionalities, as you don't need to implement a type _and_ a function, just the function. I don't think it changes the outcome here, just thought I'd call it out.

docs/design/validation.md
```diff
@@ -0,0 +1,1989 @@
...
+
+## Recommendations
+
+This section will summarize the proposals into specific recommendations for each topic. Code examples are provided in following sections to illustrate the concepts. The final section provides some future considerations for the framework and terraform-plugin-go.
+
+### Overview
+
+Defining all validation functionality via interface types will offer the framework the most flexibility for future enhancements while ensuring consistent implementations. The request and response pattern should be used to enable backwards (in the case of field deprecations) and forwards compatibility.
+
+All validation should be implemented separately from plan modifications as they address differing concerns and operations within the Terraform. Attribute validations should be implemented as a slice of the interface type on `schema.Attribute` while Data Source, Provider, and Resource level validation should be implemented as new extension interface types. Further helper functions and designs can reduce implementation details.
```
paddycarver, line 1309: "within the Terraform"

docs/design/validation.md
```diff
@@ -0,0 +1,1989 @@
...
+
+### Data Source Example Implementation
+
+Example framework code:
+
+```go
+// ValidateDataSourceConfigRequest contains request information from the ValidateDataSourceConfig RPC.
+type ValidateDataSourceConfigRequest struct {
```
paddycarver, line 1317: Bikeshed: my super spicy take here is we can probably get away without including `Config` in all of these, making them `ValidateDataSourceRequest`, etc. just to cut down on the verbosity. I don't know that any semantics are lost. There's a danger of a new RPC being introduced in the future that makes things more ambiguous, but I think we could make _that_ specially named if that unlikely scenario occurs.

docs/design/validation.md
```diff
@@ -0,0 +1,1989 @@
...
+    // ...
+    Validators []AttributeValidator
+}
+```
+
+Example validation function code:
+
+```go
+type stringLengthBetweenValidator struct {
+    StringValueValidator
```
paddycarver, line 1507: What is this type? I don't think, based on the code you've supplied, this is necessary.

docs/design/validation.md
```diff
@@ -0,0 +1,1989 @@
...
+        StringLengthBetween(1, 256),
+    },
+}
+```
+
+### Future Considerations
+
+It is recommended that the framework provide an abstracted `*tftypes.AttributePath` rather than depend on that type directly, but this can converted after an initial implementation. This is purely for decoupling the two projects, similar to other abstracted types already created in the framework.
+
+To better support provider-based validation functionality in the future, it is also recommended that the `Provider` interface type also add a new `Configured(context.Context) bool` function or another methodology for easily checking the configuration state of a provider instance. Adding a setter function could also allow the framework to manage the provider configuration state automatically. This would simplify validations that require provider instances since it will likely be required that implementations need to check on this status as part of the validation logic.
```
paddycarver, line 1579: I'm 🤔 on this and would be interested in hearing your thoughts. Maybe instead of "recommended", this is something we'd like to further explore and give design consideration to? I just don't want to pre-commit to it being a Thing We Should Do without talking about it a bit more, but I don't know that that conversation needs to happen now or in this design. But I think calling it out as a thing that would impact this that we may want to look into doing is worthwhile.

paddycarver (approved):
This latest revision looks great, and the proposal looks reasonable to me. There are a couple of doc suggestions here that don't really touch the design, and some bikeshedding on some naming, but overall, I'm really happy with the doc and the direction. I'm super interested in hearing @kmoe's thoughts, though, and how she thinks it will intersect with plan modification, as I think we want those designs to be similar.

docs/design/validation.md
bflad: Indeed! Will be fixed in the next commit. Thanks for calling this out.

docs/design/validation.md
bflad: Will be updated in the next commit. 👍

docs/design/validation.md
bflad: Yes, indeed. Thank you for this suggestion.

docs/design/validation.md
bflad: Indeed! Will be fixed in the next commit. Thanks for calling this out.

docs/design/validation.md
bflad: Love it, will be included in next commit.

docs/design/validation.md
bflad: I believe it can! I will add that as a possibility too.

docs/design/validation.md
bflad: Yes, indeed. Updated in next commit. 💯

docs/design/validation.md
bflad: "silly the brain" 🤪  Thanks!

docs/design/validation.md
bflad: Famous last words 😄  I think I will leave this for implementation consideration as to not confuse readers about naming differences from the RPCs, unless we want a naming section in _this_ documentation about why we strayed. Seems more important to capture in the commit message or a code comment.

docs/design/validation.md
bflad: Sorry, older code based on the strongly typed sketches, will update next commit.

docs/design/validation.md
bflad: Sure, as is true with all of these. Will update this section's verbiage to recommend potentially discussing these things.

docs/design/validation.md
```diff
@@ -0,0 +1,2048 @@
...
+```go
+schema.Attribute{
+    // ...
+    ConflictsWith: []func(AttributeValueValidator, tftypes.AttributePath, AttributeValueValidator),
+}
+```
+
+Regardless of the potential value handling, this proposal would feel familiar for existing provider developers and be relatively trivial for them to implement. One noticable downside to this approach however is that there can be any number of related, but disjointed attribute behaviors. The previous framework supported four of these already and there is logical room for addtional behaviors, making updates to the `schema.Attribute` type a limiting factor in this validation space. This proposal also differs from value validation proposals, which are focused around a single field.
+
+##### `PathValidation` Field on `schema.Attribute`
```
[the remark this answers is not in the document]
bflad: Section reworking has refactored this away.

docs/design/validation.md
```diff
@@ -0,0 +1,2048 @@
...
+
+##### Resource Level Attribute Value Validation Handling
+
+This proposal would introduce no changes to `schema.Attribute`. Instead, this would require value validation declarations at the `DataSource` and `Resource` level similar to other proposed attribute validations in the [Declaring Multiple Attribute Validation for Resources](#declaring-multiple-attribute-validation-for-resources) section.
+
+This proposal makes value validation behaviors occur at a distance, meaning it is harder for provider developers to correlate the validation logic to the name/path and type information. It would also be very verbose for even moderately sized schemas with thorough value validation. The only real potential benefit to this type of value validation is that the framework implementation is very straightforward, to just go through this single list of validations instead of walking all attributes.
```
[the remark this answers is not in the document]
paddycarver: Sorry, missed this in the shuffle. I think what I'm trying to say is that, for example:

```go
// ignore the type signature, it's a place holder
func (r computeInstanceResource) Validate(ctx context.Context, req ValidateResourceRequest, resp *ValidateResourceResponse) {
  // assume we want to ensure, e.g., that no disks combine to more than one terabyte of storage, or something
  totalStorage := 0
  for _, disk := range plan.Disks { // not how any of this works, but pseudo-code!
    totalStorage += disk.Size
  }
  if totalStorage >= 1024*1024*1024*1024 {
    // return an error!
  }
}
```

This is a weird thing to define a reusable helper for, and it's weirder if we throw more details on this: maybe it can only be 1tb of spinning rust disks, 512gb of SSDs, and no more than 1tb of disks total. Maybe the size of the instance comes into play. That's what I mean by "incredibly complex"--I guess a better way to phrase it would be "unlikely to be reused across resources".

To implement this declaratively, you'd need to do something like this:

```go
// ignore the type signature, it's a place holder
func (r computeInstanceResource) Validators(ctx context.Context) []Validators {
  return []Validators{computeInstanceValidator{}}
}

type computeInstanceValidator struct{}

func (r computeInstanceValidator) Validate(ctx context.Context, req ValidateResourceRequest, resp *ValidateResourceResponse) {
  // assume we want to ensure, e.g., that no disks combine to more than one terabyte of storage, or something
  totalStorage := 0
  for _, disk := range plan.Disks { // not how any of this works, but pseudo-code!
    totalStorage += disk.Size
  }
  if totalStorage >= 1024*1024*1024*1024 {
    // return an error!
  }
}
```

I'm arguing this is less straightforward because you need to introduce the indirection of declaring the validation logic separately from the resource, in a reusable chunk, even though you're only using it for the resource. The code is a bit more verbose, and feels a little less intuitive, imho.

docs/design/validation.md
```diff
@@ -0,0 +1,2025 @@
...
+err := validator.Validate(ctx, value)
+
+if err != nil {
+    return fmt.Errorf("%s: %w", attributePath.String(), err)
+}
+```
+
+This could be a double edged sword for extensibility. Implementators do not need to worry about handling the attribute path in error messages that are returned to practitioners or manually adding logging around it. This does however prevent the ability to provide that additional context to the validation logic, if for example the logic warrants making decisions based on the given path or additional logging that includes the full path. In practice with validation functions in the previous framework, path based decisions are rare at best, and this framework could be opinionated against that particular pattern.
+
+##### Adding Attribute Path to Context
```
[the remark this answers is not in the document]
paddycarver: I am _incredibly_ in favor of "you probably shouldn't use `context.WithValue`", with the exception of request-scoped data. What is considered to be request-scoped data I think will continue to be the source of some debate.

kmoe (approved):

PR #72, opened 2021-07-18
Add support for retrieving tfprotov6.ProviderServer.

paddycarver:
We need access to the tfprotov6.ProviderServer underlying `tfsdk.Serve`
for things like testing and muxing. Add a function that will expose it.


bflad (approved):
Looks good to me 🚀 

Not sure if it's worth mentioning that this new function is typically used for testing and muxing purposes for future travelers or those going through the Go package documentation.

kmoe (approved):

PR #73, opened 2021-07-18
Fix out of date guidance for GetSchema.

paddycarver:
The Provider interface's GetSchema method told provider developers to
return nil, but they can't actually return nil anymore. Update to
correct the guidance.


bflad (approved):
Looks good to me 🚀

kmoe (approved):

PR #74, opened 2021-07-18
Default to prior state in responses.

### LINKED ISSUE #51: Support partial state applies
### Module version

```
N/A
```

### Use-cases

When applying changes, it's possible for a plan to get partially applied. This can happen, for example, when creating or updating a resource takes multiple API calls, and one succeeds and the other fails.

Normally, this isn't a problem for providers. On the next refresh, the modifications will be picked up, and everything will be fine. It is, however, a problem under two scenarios:

1. The resource is being created, and doesn't get set in state at all because of an error. If there's nothing to refresh, we can't pick up the changes.
2. The resource is being updated or created, and the API returns data in the modification response that isn't present in the read response. If we discard this data, we can't get it back at refresh time.

To fight these limitations, Terraform will update its state with whatever is sent back as the new state in the response even in the face of an error, though it will halt the graph walk if there's an error.

We need to figure out how we want to send Terraform back the state that we actually want to preserve.

### Attempted Solutions

helper/schema solved this by using `schema.ResourceData.Partial` and `schema.ResourceData.SetPartial`. This would turn "partial apply" mode on and off, and set state that would only be retained if partial apply mode was on when the error hit.

This confused people, and doesn't really work properly to this day. Nobody knew when to set partial apply mode to on and when to turn it off. The partial state was yet another thing to keep track of, and how it interacted with the regular state wasn't super clear.

### Proposal

I'm proposing that we default the state being returned to the prior state, the state before apply. This means, in the face of an error, the provider will set the state back to what it was at the beginning, with whatever changes they've made in the meantime retained. The drawback is that providers then need to think about what state they've modified when returning an error.

Alternatively, we could add a `ResetStateOnError` property on the response type:

```go
func (r myResource) Create(ctx context.Context, req tfsdk.CreateResourceRequest, resp *tfsdk.CreateResourceResponse) {
  resp.ResetStateOnError = true
}
```

If `ResetStateOnError` is `true` when the function exits and `resp` has an error diagnostic, the response's state will be overwritten with the prior state from the request. If we think it makes sense, we could even default `ResetStateOnError` to `true`, making the behavior opt-out.

### References

- hashicorp/terraform-plugin-sdk#201

paddycarver:
Don't pass empty state values into response types. Instead, default to
the prior or current state, which is almost always what providers will
want.

(Partially?) resolves #51.

### DIFF
--- a/.changelog/74.txt
+++ b/.changelog/74.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+Responses now default to returning the current state, meaning state will only change when provider developers actively change it. Previously, an empty state value would be returned, which caused problems.
+```
--- a/tfsdk/serve.go
+++ b/tfsdk/serve.go
@@ -723,7 +723,6 @@ func (s *server) ApplyResourceChange(ctx context.Context, req *tfprotov6.ApplyRe
 		})
 		return resp, nil
 	}
-	return resp, nil
 }
 
 func (s *server) ImportResourceState(ctx context.Context, _ *tfprotov6.ImportResourceStateRequest) (*tfprotov6.ImportResourceStateResponse, error) {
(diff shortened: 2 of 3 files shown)


bflad (approved):
Looks good to me 🚀

kmoe (approved):

PR #75, opened 2021-07-20
tfsdk: Initial support for Data Source, Provider, and Resource validation

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/pull/65


tfsdk/data_source_validation.go
```diff
@@ -0,0 +1,33 @@
...
+
+// DataSourceWithConfigValidators is an interface type that extends DataSource to include declarative validations.
+type DataSourceWithConfigValidators interface {
+	DataSource
+
+	// ConfigValidators returns a list of functions which will all be performed during validation.
+	ConfigValidators(context.Context) []DataSourceConfigValidator
+}
+
+// DataSourceWithValidateConfig is an interface type that extends DataSource to include imperative validation.
```
paddycarver, line 27: Ideally, there'd be some description of when to use this versus the DataSourceConfigValidators.

tfsdk/request_validation.go
```diff
@@ -0,0 +1,46 @@
...
+type ValidateDataSourceConfigRequest struct {
+	// Config is the configuration the user supplied for the data source.
+	//
+	// This configuration may contain unknown values if a user uses
+	// interpolation or other functionality that would prevent Terraform
+	// from knowing the value at request time.
+	Config Config
+
+	// TypeName is the data source type name.
+	TypeName string
```
paddycarver, line 16: Do we want to surface this? I think we opted not to in other places....

tfsdk/serve.go
```diff
@@ -216,15 +216,73 @@ func (s *server) GetProviderSchema(ctx context.Context, _ *tfprotov6.GetProvider
...
+			}
+		}
+	}
+
+	if provider, ok := s.p.(ProviderWithValidateConfig); ok {
+		vpcRes := &ValidateProviderConfigResponse{}
+
+		provider.ValidateConfig(ctx, vpcReq, vpcRes)
+
+		if vpcRes.Diagnostics != nil {
```
paddycarver, line 278: Is it worth using `if len(vpcRes.Diagnostics) > 0`? I don't know that this check is necessary at all, honestly, but.

tfsdk/serve.go
```diff
@@ -216,15 +216,73 @@ func (s *server) GetProviderSchema(ctx context.Context, _ *tfprotov6.GetProvider
...
+			configValidator.Validate(ctx, vpcReq, vpcRes)
+
+			if vpcRes.Diagnostics != nil {
+				resp.Diagnostics = append(resp.Diagnostics, vpcRes.Diagnostics...)
+			}
+		}
+	}
+
+	if provider, ok := s.p.(ProviderWithValidateConfig); ok {
+		vpcRes := &ValidateProviderConfigResponse{}
```
paddycarver, line 274: Do we want to insert any existing diagnostics into the response before passing it on?

bflad:
I'm going to address the above comments and pull in the `Attribute` validation from #76 so it can all be reviewed together. 👍

tfsdk/request_validation.go
bflad: I'm guessing it could be useful in the case where there are multiple resources against the same remote system that are similar, but slightly different. If we're not passing through this information that's available over the wire though in other places, happy to remove this for now (we can always add it back).

tfsdk/data_source_validation.go
```diff
@@ -0,0 +1,33 @@
+package tfsdk
+
+import (
+	"context"
+)
+
+// DataSourceConfigValidator describes reusable DataSource configuration validation functionality.
+type DataSourceConfigValidator interface {
+	// Description describes the validation in plain text formatting.
+	Description(context.Context) string
```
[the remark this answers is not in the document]
bflad: I think its preferable to introduce this upfront and potentially make it optional later, rather than vice-versa.

tfsdk/serve.go
bflad: It is not 😄  I will remove the conditional.

bflad: Removed these across the board in 209de761804e2367f0fd0e9cdcc513146cc1bae5 👍

tfsdk/serve.go
bflad: This is an interesting proposition! I see the value in it -- essentially being able to change or remove diagnostics with these validation methods. I'm a little torn on the implementation however, as it feels weird to pre-populate part of a response and have developers of these methods understand the semantics of those potentially pre-populated responses. It _feels_ like to support this we may want to potentially pass in a "prior diagnostics" request field and either allow developers to return updated/removed diagnostics in the existing response `Diagnostics` field (although it seems like they could forget to passthrough existing ones) or provide separate fields in the response for this special handling. I'm a little worried that this would be different than other diagnostics handling, which is currently append-only everywhere else.

tfsdk/data_source_validation.go
```diff
@@ -0,0 +1,33 @@
+package tfsdk
+
+import (
+	"context"
+)
+
+// DataSourceConfigValidator describes reusable DataSource configuration validation functionality.
+type DataSourceConfigValidator interface {
```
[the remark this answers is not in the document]
bflad: I'm certainly amenable to updating the name, I'm just not sure if there's any concerns with keeping the names and types relatively close to the RPCs similar to as has been done elsewhere. Should there be any future other implementations with differing semantics, it could be weird to have `DataSourceValidator` vs `DataSourceStateValidator` for example.

tfsdk/data_source_validation.go
```diff
@@ -0,0 +1,33 @@
...
+	// Description describes the validation in plain text formatting.
```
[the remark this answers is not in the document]
bflad: Added some additional comments in 4be5375202b14b800adabfcf3bb722c2faa8c75c -- the wording could probably be improved though so please give a shout.

tfsdk/data_source_validation.go
bflad: I hope I've covered this okay with bffb3f6d5c96ef58d0094ed78a1e9c708617a22e, please suggest better comments!

tfsdk/request_validation.go
bflad: Removed in 572ee67c06437fed09622e9a74e7b046422acdeb

bflad:
This now includes the initial `Schema` and `Attribute` validation from #76 wired together with the Validate RPC calls. It needs a test fix from the testing changing on main and unit testing for the `Schema` and `Attribute` validation. I've run out of time for today, but it might be worth a cursory look to ensure it's going in the right direction.

tfsdk/serve.go
paddycarver: What I'm trying to get at is I _think_ everywhere else we populate the response object we pass to the provider with the diagnostics we already have (if any), and they append to it, and we replace the gRPC response diagnostics with what comes back. ... I think.

I don't know that it matters, I'm mostly just arguing for consistency.

tfsdk/schema.go
```diff
@@ -129,6 +129,23 @@ func (s Schema) AttributeAtPath(path *tftypes.AttributePath) (Attribute, error)
 	return a, nil
 }
 
+// Validate performs all Attribute validation.
+func (s Schema) Validate(ctx context.Context, req ValidateSchemaRequest, resp *ValidateSchemaResponse) {
```
paddycarver, line 133: Does this _need_ to be exported?

paddycarver (commented):
Seems like it's on the right path to me.

tfsdk/schema.go
bflad: Likely not -- updated in https://github.com/hashicorp/terraform-plugin-framework/pull/75/commits/3371b5470a35cefa1cb183303c88fc07a8557ef4

tfsdk/serve.go
```diff
@@ -222,15 +222,78 @@ func (s *server) GetProviderSchema(ctx context.Context, _ *tfprotov6.GetProvider
...
+
+		provider.ValidateConfig(ctx, vpcReq, vpcRes)
+		resp.Diagnostics = append(resp.Diagnostics, vpcRes.Diagnostics...)
+	}
+
+	// TODO: Add unit testing
```
paddycarver, line 283: Did this get added?

tfsdk/serve.go
```diff
@@ -272,12 +335,86 @@ func (s *server) StopProvider(ctx context.Context, _ *tfprotov6.StopProviderRequ
...
+	}
+
+	if resource, ok := resource.(ResourceWithValidateConfig); ok {
+		vrcRes := &ValidateResourceConfigResponse{}
+
+		resource.ValidateConfig(ctx, vrcReq, vrcRes)
+		resp.Diagnostics = append(resp.Diagnostics, vrcRes.Diagnostics...)
+	}
+
+	// TODO: Add unit testing
```
paddycarver, line 404: Did this get added?

tfsdk/serve.go
```diff
@@ -823,12 +960,86 @@ func (s *server) ImportResourceState(ctx context.Context, _ *tfprotov6.ImportRes
...
+	}
+
+	if dataSource, ok := dataSource.(DataSourceWithValidateConfig); ok {
+		vrcRes := &ValidateDataSourceConfigResponse{}
+
+		dataSource.ValidateConfig(ctx, vrcReq, vrcRes)
+		resp.Diagnostics = append(resp.Diagnostics, vrcRes.Diagnostics...)
+	}
+
+	// TODO: Add unit testing
```
paddycarver, line 1029: Did this get added?

tfsdk/serve_resource_config_validators_test.go
```diff
@@ -0,0 +1,96 @@
...
+func (r testServeResourceConfigValidators) Update(ctx context.Context, req UpdateResourceRequest, resp *UpdateResourceResponse) {
+}
+func (r testServeResourceConfigValidators) Delete(ctx context.Context, req DeleteResourceRequest, resp *DeleteResourceResponse) {
+}
+
+func (r testServeResourceConfigValidators) ConfigValidators(ctx context.Context) []ResourceConfigValidator {
+	r.provider.validateResourceConfigCalledResourceType = "test_config_validators"
+
+	return []ResourceConfigValidator{
+		newTestResourceConfigValidator(r.provider.validateResourceConfigImpl),
```
paddycarver, line 74: Do we want to test having multiple validators? 🤔

tfsdk/serve.go
bflad: Updated in faa5afb1c9f191bf4533743c8848966713337a69 and ee6ea5de2d21dafd19ecdc0e5bf2e6c50a3633b8 for consistency with other functions.

tfsdk/serve.go
bflad: Removed in followup commit.

tfsdk/serve.go
bflad: Removed in followup commit.

tfsdk/serve.go
bflad: Removed in followup commit.

tfsdk/attribute.go
```diff
@@ -207,3 +212,134 @@ func (a Attribute) tfprotov6SchemaAttribute(ctx context.Context, name string, pa
 
 	return schemaAttribute, nil
 }
+
+// validate performs all Attribute validation.
+func (a Attribute) validate(ctx context.Context, req ValidateAttributeRequest, resp *ValidateAttributeResponse) {
+	if (a.Attributes == nil || len(a.Attributes.GetAttributes()) == 0) && a.Type == nil {
```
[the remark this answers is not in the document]
bflad: `(Attribute).tfprotov6Schema()` currently performs this same check: https://github.com/hashicorp/terraform-plugin-framework/blob/ee6ea5de2d21dafd19ecdc0e5bf2e6c50a3633b8/tfsdk/attribute.go#L161-L163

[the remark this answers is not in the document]
paddycarver: Ah, ignore me. We should probably refactor this into a common helper or something so if we change this logic in the future we can change it once and be done, but that's outside the scope of this PR.

paddycarver (approved):
LGTM, great work @bflad.

tfsdk/serve_resource_config_validators_test.go
bflad: I'll follow this up with one last commit to cover that 👍

tfsdk/schema.go
```diff
@@ -183,3 +183,20 @@ func (s Schema) tfprotov6Schema(ctx context.Context) (*tfprotov6.Schema, error)
...
+			AttributePath: tftypes.NewAttributePath().WithAttributeName(name),
+			Config:        req.Config,
+		}
+		attributeResp := &ValidateAttributeResponse{
+			Diagnostics: resp.Diagnostics,
+		}
+
+		attribute.validate(ctx, attributeReq, attributeResp)
+
+		resp.Diagnostics = attributeResp.Diagnostics
```
kmoe, line 200: Should this append instead, in case multiple attribute validators return warning diags?

bflad: Initially this was appending, but I was asked to switch it to "passthrough" diagnostics: https://github.com/hashicorp/terraform-plugin-framework/pull/75#discussion_r678710013

Personally I think pre-populating a response, while offering overwrite capabilities easily, could be confusing for implementors to understand the semantics. We could offer prior diagnostics in the request with diagnostic modifications in the response to clarify things on both sides of these calls (or not allow diagnostic modification). Please let me know if you would like me to raise an issue regarding this.

PR #76, opened 2021-07-20
schema: Proof of concept Schema and Attribute validation

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/pull/65

Note that this code as written introduces an import cycle between tfsdk and schema packages as they need to cross-reference types. Proposal pull requests will be submitted to migrate certain package functionality between packages to alleviate the issue.

### DIFF
--- a/schema/attribute.go
+++ b/schema/attribute.go
@@ -1,9 +1,11 @@
 package schema
 
 import (
+	"context"
 	"errors"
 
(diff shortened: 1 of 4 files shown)


bflad:
Now that #77 has been merged, I'm going to recreate the changes of this pull request into #75, which should compile and work as expected! 👍

PR #77, opened 2021-07-20
Migrate schema package and internal/proto6 Schema/Attribute into tfsdk package

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/pull/76


paddycarver (approved):
I think this would be my preferred approach, honestly.

kmoe (approved):
Great, works for me

paddycarver (approved):

Issue #24, opened 2021-05-18
Define Diagnostics type and helpers.


paddycarver:
### Module version

```
N/A
```

### Use-cases

We want people to be able to return diagnostics and use them ergonomically, so we should provide some helper functions and ways to use them that makes it easier and less verbose to define diagnostics.

paddycarver:
Do we want diagnostics to be an interface? Maybe a superset of the `error` interface? Maybe with a built-in implementation?

That would help address some of the validation stuff @bflad wants to do, e.g. being able to progressively enhance to support `Not`, etc. helpers on validation.

It would also allow providers to treat diagnostics more like errors, detecting on them, programming with them, allowing them to be more embedded in provider code rather than being for practitioner consumption _only_.

We could (in theory) have an interface something like this:

```go
type Diagnostic interface {
  error

  DiagnosticSeverity() Severity
  DiagnosticSummary() string
  DiagnosticDetail() string
}

type DiagnosticWithPath interface {
  Diagnostic

  DiagnosticPath() tfsdk.Path
}
```

I think that would give the framework everything it needs to construct the `tfprotov6.Diagnostic`, and leaves us room to expand in the future if future protocol versions add more fields/whatever to diagnostics. It also means that diagnostics could be smuggled through things that are only meant to return errors, and detected (with fallback) on the other side.

bflad:
Picking this up -- especially with recent `Get`/`Set` changeover to diagnostics, this is much more critical path now.

paddycarver:
Do you want to start with a design doc, or do you want to head right into implementation?

bflad:
I'm thinking design documentation if time allows, but happy to also just submit a very basic starter implementation, e.g.

```go
package diag

type Diagnostic struct { /* */ }

type Diagnostics []Diagnostic

func (d Diagnostics) Append(in Diagnostics)

func (d Diagnostics) HasError() bool
```

To unblock the current situation, until we have more time to noodle on whether interface types are a good idea (it seems valuable 😄 ) and write out design ideas.

paddycarver:
My hot take is to do a timeboxed design document for, say, a week and let's figure out how to ship enough of an MVP to make it work if it doesn't look like the design document is getting consensus?

PR #82, opened 2021-07-27
Introduce attr.TypeWithValidate handling to internal/reflect and tfsdk

bflad:

The next step will be deciding how to handle the returned diagnostics since there can be multiple warnings and/or errors, however these callers return a single error type.


paddycarver:
> these callers return a single error type.

If it's cleaner/better that they return Diagnostics, I think that's at least worth _exploring_ as an option.

bflad:
Something that might also be worth considering with this type of validation is what might happen if a type wants to introduce a non-backwards compatible change. Warning diagnostics are a good approach to prepare practitioners for such a change, but if in a major provider release the change occurs, errors during state reading operations could leave practitioners in a tough spot without first applying/refreshing an updated value on the older version.

paddycarver:
Hmmmm, yeah, that's a solid point. Would we want to introduce migrations in that case? Do we _want_ to validate on read? I think we have to, right, to ensure we have the appropriate value for the type? 🤔 I think providers could work around it with state migrations, but that may not be ideal.

bflad:
I still need to circle through all the places that are introducing the `attr.TypeWithValidation` handling and add covering unit testing that diagnostics are surfaced, but the code should be mostly there.

bflad:
This is ready, @kmoe!

internal/reflect/pointer_test.go
```diff
@@ -101,3 +102,31 @@ func TestFromPointer_null(t *testing.T) {
...
 	}
 }
+
+func TestFromPointer_AttrTypeWithValidate_Error(t *testing.T) {
+	v := "hello, world"
+	_, diags := refl.FromPointer(context.Background(), testtypes.StringTypeWithValidateError{}, reflect.ValueOf(&v), tftypes.NewAttributePath())
+	if len(diags) == 0 {
+		t.Fatalf("expected diagnostics, got none")
+	}
+	if !reflect.DeepEqual(diags[0], testtypes.TestErrorDiagnostic) {
```
paddycarver, line 112: I think so far we've been using `cmp.Equal` for these. Does that not work here? I think it's a bit more flexible than `reflect.DeepEqual`.

internal/testing/types/bool.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+)
+
+var (
+	_ attr.Type = BoolType{}
+)
+
+// BoolType is a reimplementation of types.BoolType that can be used as a base
+// for other extension types in testing.
+type BoolType struct{}
```
paddycarver, line 18: Is there a reason the test types can't just be based off types.BoolType?

tfsdk/config_test.go
```diff
@@ -0,0 +1,222 @@
...
+	"reflect"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	testtypes "github.com/hashicorp/terraform-plugin-framework/internal/testing/types"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+)
+
+func TestConfigGet(t *testing.T) {
```
paddycarver, line 14: Is there a reason not to make this a table-driven test?

tfsdk/config_test.go
```diff
@@ -0,0 +1,222 @@
...
+	expected := testConfigGetData{
+		Name: types.String{Value: "namevalue"},
+	}
+
+	if diff := cmp.Diff(val, expected); diff != "" {
+		t.Errorf("unexpected diff (+wanted, -got): %s", diff)
+	}
+}
+
+func TestConfigGet_AttrTypeWithValidate_Error(t *testing.T) {
```
paddycarver, line 54: I think all these tests need the `t.Parallel` call in them so they'll run in parallel.

types/diags.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-go/tfprotov6"
+)
+
+// TODO: Replace with diagnostics abstraction
+// Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/24
+func diagsHasErrors(in []*tfprotov6.Diagnostic) bool {
```
paddycarver, line 12: These have been implemented in a number of different packages now. Is it worth just pulling them into an `internal/diags` package at this point?

internal/reflect/interfaces.go
```diff
@@ -22,41 +23,84 @@ type Unknownable interface {
...
 	})
 	err := results[0].Interface()
 	if err != nil {
-		return target, path.NewError(err.(error))
+		var underlyingErr error
+		switch e := err.(type) {
+		case error:
+			underlyingErr = e
+		default:
+			underlyingErr = fmt.Errorf("unknown error type: %T", e)
```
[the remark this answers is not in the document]
bflad: Makes sense, will add. 👍

internal/reflect/pointer_test.go
bflad: Sorry not used to that yet, will refactor.

internal/testing/types/bool.go
bflad: If I recall, it was because I was having trouble extending a `primitive` value. I'll try again today.

tfsdk/config_test.go
bflad: I was following the style of the other State tests, personally I'd prefer TDTing as well.

tfsdk/config_test.go
bflad: Will update these to include `t.Parallel()` regardless of TDTing or not.

types/diags.go
bflad: 🙁 that's what I was alluding to when I wanted to start working on diagnostics before this, but I didn't want to sway the design by doing so. I will refactor this common diagnostics functionality into an internal package as suggested.

internal/reflect/pointer_test.go
bflad: Updated in 28696c08696c746a258bd590b55481bfd2dfccad

types/diags.go
bflad: Updated in c417a51b79a1c88cd3fed1f3da16b39137e748e8

tfsdk/config_test.go
bflad: `t.Parallel()`-ized all the `Config`, `Plan`, and `State` testing (including existing tests) in c8cef290dea05aa5330a6645be43bf2b5350e23a

tfsdk/config_test.go
bflad: I'm like 3/4 of the way through converting these... will pick up effort our pairing.

bflad: Latest commit switched all `Config`, `Plan`, and `State` testing to table driven testing.

internal/testing/types/bool.go
bflad: Here's where I was getting stuck.

Trying with type aliasing fails:

```go
type BoolTypeWithValidateError types.BoolType
```

```
types.BoolType (constant 2 of type types.primitive) is not a type (NotAType)
```

Creating a struct type (with pointer or concrete) fails:

```go
type BoolTypeWithValidateError struct {
	types.BoolType
}
```

```
types.BoolType (constant 2 of type types.primitive) is not a type (NotAType)
```

Trying as a variable also sensibly fails:

```go
var BoolTypeWithValidateError = types.BoolType

func (t BoolTypeWithValidateError) Validate(ctx context.Context, in tftypes.Value) []*tfprotov6.Diagnostic {
	return []*tfprotov6.Diagnostic{TestErrorDiagnostic}
}
```

```
BoolTypeWithValidateError (variable of type types.primitive) is not a type (NotAType)
```

paddycarver (approved):
I think this looks good. Great work, @bflad. I think once merged we're committing ourselves to tackling #24 before release, though, as people will need to know, at the very least, if `Get` and `Set` (etc.) are returning errors they should abort on or warnings they should roll through. I think that's okay, though.

internal/testing/types/bool.go
paddycarver: Annoying. :(

bflad: I'll leave this information in a new issue for further consideration.

PR #93, opened 2021-08-10
tfsdk: Return warning diagnostic when using `Attribute` or `Schema` type `DeprecationMessage` field

### LINKED ISSUE #71: Actually use DeprecationMessage on resources, data sources, providers, and attributes.
### Module version

```
v0.1.0
```

### Use-cases

When deprecating an attribute, provider, resource, or data source, a warning is supposed to be shown to users that attempt to use that attribute, provider, resource, or data source. We even have a place in `schema.Schema` and `schema.Attribute` to specify the message for this warning. However, those properties don't do anything at the moment and no warnings are generated for using deprecated attributes, providers, resources, and data sources.

### Proposal

We should check those values in various RPC calls, generating warnings when the deprecated attribute, provider, resource, or data source is used.

bflad:
Closes #71


tfsdk/attribute.go
```diff
@@ -344,4 +344,13 @@ func (a Attribute) validate(ctx context.Context, req ValidateAttributeRequest, r
 			return
 		}
 	}
+
+	if a.DeprecationMessage != "" {
```
paddycarver, line 348: Wouldn't we only want to return the error if the value wasn't `null`?

tfsdk/attribute_test.go
```diff
@@ -803,6 +803,39 @@ func TestAttributeValidate(t *testing.T) {
 			},
 			resp: ValidateAttributeResponse{},
 		},
+		"deprecation-message": {
```
paddycarver, line 806: Would love to see a test for when the value is:

* present
* absent
* unknown

paddycarver (changes requested):

tfsdk/attribute.go
bflad: Yes, we would! Good catch. Testing updated to catch this:

```
--- FAIL: TestAttributeValidate (0.00s)
    --- FAIL: TestAttributeValidate/deprecation-message-null (0.00s)
        attribute_test.go:1313: Unexpected response (+wanted, -got):   tfsdk.ValidateAttributeResponse{
            -   Diagnostics: []*tfprotov6.Diagnostic{
            -           &{
            -                   Severity:  s"WARNING",
            -                   Summary:   "Attribute Deprecated",
            -                   Detail:    "Use something else instead.",
            -                   Attribute: s`AttributeName("test")`,
            -           },
            -   },
            +   Diagnostics: nil,
              }
```

Fix will be included in next commit.

paddycarver (approved):

PR #101, opened 2021-08-16
Fix version constraints in README.

paddycarver:
We actually need Terraform v1.0.3 or higher, not 0.15.


bflad (approved):
Looks good to me 🚀

PR #102, opened 2021-08-16
Attribute plan modification

### LINKED ISSUE #34: Support customizing plans

kmoe:
Closes #34 

Design doc: #64 

Needs #107

This PR implements `schema.Attribute.PlanModifiers` and the `RequiresReplace` and `RequiresReplaceIf` helpers, as described in the design doc.

Some more design points that arose during implementation:
1. When multiple `AttributePlanModifiers` are specified, should the `req.AttributePlan` value supplied to each be the `resp.AttributePlan` value from the previous modifier? This seems the intuitive behaviour (see `chained_modifiers` test for example).
2. If an error diagnostic is returned at some point during attribute plan modification, should `RequiresReplace` and the planned state still be populated in the server response? This PR assumes so, and that it is Core's responsibility to decide what to do in this case. This also makes it easier to write deterministic tests, as there is otherwise a race condition between early return on certain errors, and RequiresReplace being populated.

### DIFF
--- a/tfsdk/attribute.go
+++ b/tfsdk/attribute.go
@@ -73,7 +73,50 @@ type Attribute struct {
 
 	// Validators defines validation functionality for the attribute.
 	Validators []AttributeValidator
+
+	// PlanModifiers defines a sequence of modifiers for this attribute at
+	// plan time.
+	// Please note that plan modification only applies to resources, not
+	// data sources. Setting PlanModifiers on a data source attribute will
+	// have no effect.
+	PlanModifiers AttributePlanModifiers
 }
+
+// AttributePlanModifier represents a modifier for an attribute at plan time.
+// An AttributePlanModifier can only modify the planned value for the attribute
+// on which it is defined. For plan-time modifications that modify the values of
+// several attributes at once, please instead use the ResourceWithModifyPlan
+// interface by defining a ModifyPlan function on the resource.
+type AttributePlanModifier interface {
+	// Description is used in various tooling, like the language server, to
+	// give practitioners more information about what this modifier is,
+	// what it's for, and how it should be used. It should be written as
+	// plain text, with no special formatting.
+	Description(context.Context) string
+
+	// MarkdownDescription is used in various tooling, like the
+	// documentation generator, to give practitioners more information
+	// about what this modifier is, what it's for, and how it should be
+	// used. It should be formatted using Markdown.
+	MarkdownDescription(context.Context) string
+
+	// Modify is called when the provider has an opportunity to modify
+	// the plan: once during the plan phase when Terraform is determining
+	// the diff that should be shown to the user for approval, and once
+	// during the apply phase with any unknown values from configuration
+	// filled in with their final values.
+	// The Modify function has access to the config, state, and plan for
+	// both the attribute in question and the entire resource, but it can
+	// only modify the value of the one attribute.
+	//
+	// Please see the documentation for ResourceWithModifyPlan#ModifyPlan
+	// for further details.
+	Modify(context.Context, ModifyAttributePlanRequest, *ModifyAttributePlanResponse)
+}
+
+// AttributePlanModifiers represents a sequence of AttributePlanModifiers, in
+// order.
+type AttributePlanModifiers []AttributePlanModifier
 
 // ApplyTerraform5AttributePathStep transparently calls
 // ApplyTerraform5AttributePathStep on a.Type or a.Attributes, whichever is
--- a/tfsdk/attribute.go
+++ b/tfsdk/attribute.go
@@ -82,97 +82,6 @@ type Attribute struct {
 	PlanModifiers AttributePlanModifiers
 }
 
-// AttributePlanModifier represents a modifier for an attribute at plan time.
-// An AttributePlanModifier can only modify the planned value for the attribute
-// on which it is defined. For plan-time modifications that modify the values of
-// several attributes at once, please instead use the ResourceWithModifyPlan
-// interface by defining a ModifyPlan function on the resource.
-type AttributePlanModifier interface {
-	// Description is used in various tooling, like the language server, to
-	// give practitioners more information about what this modifier is,
-	// what it's for, and how it should be used. It should be written as
-	// plain text, with no special formatting.
-	Description(context.Context) string
-
-	// MarkdownDescription is used in various tooling, like the
-	// documentation generator, to give practitioners more information
-	// about what this modifier is, what it's for, and how it should be
-	// used. It should be formatted using Markdown.
-	MarkdownDescription(context.Context) string
-
-	// Modify is called when the provider has an opportunity to modify
-	// the plan: once during the plan phase when Terraform is determining
-	// the diff that should be shown to the user for approval, and once
-	// during the apply phase with any unknown values from configuration
-	// filled in with their final values.
-	// The Modify function has access to the config, state, and plan for
-	// both the attribute in question and the entire resource, but it can
-	// only modify the value of the one attribute.
-	//
-	// Please see the documentation for ResourceWithModifyPlan#ModifyPlan
-	// for further details.
-	Modify(context.Context, ModifyAttributePlanRequest, *ModifyAttributePlanResponse)
-}
-
-// AttributePlanModifiers represents a sequence of AttributePlanModifiers, in
-// order.
-type AttributePlanModifiers []AttributePlanModifier
-
-// RequiresReplace returns AttributePlanModifiers specifying the attribute as
-// requiring replacement. This behaviour is identical to the ForceNew behaviour
-// in terraform-plugin-sdk.
-func RequiresReplace() AttributePlanModifiers {
-	return []AttributePlanModifier{RequiresReplaceModifier{}}
-}
-
-// RequiresReplaceModifier is an AttributePlanModifier that sets RequiresReplace
-// on the attribute.
-type RequiresReplaceModifier struct{}
-
-func (r RequiresReplaceModifier) Modify(ctx context.Context, req ModifyAttributePlanRequest, resp *ModifyAttributePlanResponse) {
-	resp.RequiresReplace = true
-}
-
-func (r RequiresReplaceModifier) Description(ctx context.Context) string {
-	return "If the value of this attribute changes, Terraform will destroy and recreate the resource."
-}
-
-func (r RequiresReplaceModifier) MarkdownDescription(ctx context.Context) string {
-	return "If the value of this attribute changes, Terraform will destroy and recreate the resource."
-}
-
-func RequiresReplaceIf(f RequiresReplaceIfFunc, description, markdownDescription string) AttributePlanModifier {
-	return RequiresReplaceIfModifier{
-		f:                   f,
-		description:         description,
-		markdownDescription: markdownDescription,
-	}
-}
-
-type RequiresReplaceIfFunc func(ctx context.Context, state, config attr.Value) (bool, error)
-
-type RequiresReplaceIfModifier struct {
-	f                   RequiresReplaceIfFunc
-	description         string
-	markdownDescription string
-}
-
-func (r RequiresReplaceIfModifier) Modify(ctx context.Context, req ModifyAttributePlanRequest, resp *ModifyAttributePlanResponse) {
-	res, err := r.f(ctx, req.State, req.Config)
-	if err != nil {
-		resp.AddError("Error running RequiresReplaceIf func for attribute", err.Error())
-	}
-	resp.RequiresReplace = res
-}
-
-func (r RequiresReplaceIfModifier) Description(ctx context.Context) string {
-	return r.description
-}
-
-func (r RequiresReplaceIfModifier) MarkdownDescription(ctx context.Context) string {
-	return r.markdownDescription
-}
-
 // ApplyTerraform5AttributePathStep transparently calls
 // ApplyTerraform5AttributePathStep on a.Type or a.Attributes, whichever is
 // non-nil. It allows Attributes to be walked using tftypes.Walk and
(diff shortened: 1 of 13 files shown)


paddycarver:
> When multiple AttributePlanModifiers are specified, should the req.AttributePlan value supplied to each be the resp.AttributePlan value from the previous modifier? This seems the intuitive behaviour (see chained_modifiers test for example).

This matches my expectations. I'm not sure how we'd let them all operate on the request; what if two or more modified the same field in different ways? Which result would we use?

> If an error diagnostic is returned at some point during attribute plan modification, should RequiresReplace and the planned state still be populated in the server response? This PR assumes so, and that it is Core's responsibility to decide what to do in this case. This also makes it easier to write deterministic tests, as there is otherwise a race condition between early return on certain errors, and RequiresReplace being populated.

This also matches my expectations, and my mental model of how plans works tells me it doesn't matter. Unlike applies, plans shouldn't get persisted in the face of an error, I believe, so if an error diagnostic is being returned, I think the plan is just going to get discarded, anyways. So this seems like reasonable behavior that won't have any external negative impacts, as far as I know.

tfsdk/schema.go
```diff
@@ -208,3 +210,189 @@ func (s Schema) validate(ctx context.Context, req ValidateSchemaRequest, resp *V
...
+						Summary:   "Attribute Plan Modification Error",
+						Detail:    "Attribute plan modifier cannot walk schema. Report this to the provider developer:\n\n" + err.Error(),
+						Attribute: attrPath,
+					})
+
+					return
+				}
+
+				for idx := range l.Elems {
+					for nestedName, nestedAttr := range a.Attributes.GetAttributes() {
```
paddycarver, line 270: I think we want this to be recursive; right now, it looks like you only have one level of nesting supported. So a nested attribute inside a nested attribute inside a schema wouldn't have its plan modifiers run, as best I can tell.

If we made this method into a function that accepted a `map[string]Attribute` and an attribute path, we could just recursively call it for nested attributes:

```go
func modifyAttributesPlans(ctx context.Context, attrs map[string]Attribute, path *tftypes.AttributePath, req ModifySchemaPlanRequest, resp *ModifySchemaPlanResponse) {
```

Then this section could be something like

```go
for idx := range l.Elems {
  modifyAttributesPlans(ctx, a.Attributes.GetAttributes(), path.WithElementKeyInt(idx), req, resp)
  if diagnostics.DiagsHasErrors(resp.Diagnostics) {
    return
  }
}
```

Or at least that's what my vague imagining is suggesting it could look like :D

tfsdk/serve_provider_test.go
```diff
@@ -431,6 +431,7 @@ func (t *testServeProvider) GetResources(_ context.Context) (map[string]Resource
 	return map[string]ResourceType{
 		"test_one":               testServeResourceTypeOne{},
 		"test_two":               testServeResourceTypeTwo{},
+		"test_three":             testServeResourceTypeThree{},
```
paddycarver, line 434: Is it worth giving this a name like the validators below, to make it clear it's testing plan modifiers? 🤔

kmoe:
> what if two or more modified the same field in different ways? Which result would we use?

`Attribute.PlanModifiers` is a slice of `AttributePlanModifiers`, executed in order. Each one can modify only the planned value for that attribute. Given this I think it's clear that if several plan modifiers on one attribute modify its value, the last modified value should be used.

Some complexity comes with nested attributes - currently it's possible to define `PlanModifiers` at several levels of the nested attribute hierarchy, e.g.:

```go
Schema{
		"disks": {
			Optional: true,
			Computed: true,
			PlanModifiers: ..., // modifies "disks"
			Attributes: ListNestedAttributes(map[string]Attribute{
				"name": {
					Required: true,
					Type:     types.StringType,
					PlanModifiers: ..., // modifies "disks[0].name" etc
				},
				"size_gb": {
					Required: true,
					Type:     types.NumberType,
				},
				"boot": {
					Required: true,
					Type:     types.BoolType,
				},
			}, ListNestedAttributesOptions{}),
		},
	},
}
```

Currently, the top-level PlanModifiers are run first, followed by any modifiers on nested attributes.

This at least needs to be noted in the GoDoc and a test case added, so I'll do that.

If we like we could also prohibit PlanModifiers on Attributes without a Type, but I don't see a need to do this as long as the order is unambiguous and clearly documented.

tfsdk/attribute.go
```diff
@@ -369,3 +376,42 @@ func (a Attribute) validate(ctx context.Context, req ValidateAttributeRequest, r
...
+		Plan:            req.Plan,
+		AttributeConfig: req.AttributeConfig,
+		AttributeState:  req.AttributeState,
+		AttributePlan:   req.AttributePlan,
+		ProviderMeta:    req.ProviderMeta,
+	}
+	for _, planModifier := range a.PlanModifiers {
+		planModifier.Modify(ctx, modifyReq, resp)
+		modifyReq.AttributePlan = resp.AttributePlan
+	}
```
paddycarver, line 416: Is it worth making this recursive, like `Attribute.validate` is, so it goes through all the nested attributes of this attribute and calls their `modifyPlan` method, too?

tfsdk/schema.go
paddycarver: This could also (maybe?) be resolved by doing the recursion on the `Attribute.modifyPlan` method, above.

tfsdk/serve_test.go
```diff
@@ -1874,14 +1899,318 @@ func TestServerPlanResourceChange(t *testing.T) {
...
+			resourceType:            testServeResourceTypeThreeType,
+			expectedRequiresReplace: []*tftypes.AttributePath{tftypes.NewAttributePath().WithAttributeName("scratch_disk").WithAttributeName("interface")},
+		},
 	}
 
 	for name, tc := range tests {
 		name, tc := name, tc
 
 		t.Run(name, func(t *testing.T) {
-			t.Parallel()
```
paddycarver, line 1883: Hmm, why did this need to stop running in parallel? 🤔 Would love a comment here explaining why it can't run in parallel. That makes me mildly worried about race conditions.

tfsdk/serve_test.go
```diff
@@ -1940,7 +2269,9 @@ func TestServerPlanResourceChange(t *testing.T) {
 				t.Errorf("Expected planned private to be %q, got %q", tc.expectedPlannedPrivate, got.PlannedPrivate)
 				return
 			}
-			if diff := cmp.Diff(got.RequiresReplace, tc.expectedRequiresReplace, cmpopts.EquateEmpty()); diff != "" {
+			if diff := cmp.Diff(got.RequiresReplace, tc.expectedRequiresReplace, cmpopts.EquateEmpty(), cmpopts.SortSlices(func(x, y *tftypes.AttributePath) bool {
```
paddycarver, line 2272: Hmm, rather than new modifiers on cmp.Diff, does implementing an `Equal` method allow us to make this work without the cmpopts?

tfsdk/serve_test.go
```diff
@@ -3088,8 +3419,6 @@ func TestServerApplyResourceChange(t *testing.T) {
 			}
 			if tc.config.Type() != nil {
 				if diff := cmp.Diff(s.applyResourceChangeConfigValue, tc.config); diff != "" {
-					t.Errorf("Unexpected diff in config (+wanted, -got): %s", diff)
```
paddycarver, line 3091: This slightly concerns me; are we just not using this `if` check anymore?

paddycarver (changes requested):
A few comments, mostly about some testing stuff and about recursion.

paddycarver:
Also, some merge conflicts to resolve, I believe.

paddycarver:
> `Attribute.PlanModifiers` is a slice of `AttributePlanModifiers`, executed in order. Each one can modify only the planned value for that attribute. Given this I think it's clear that if several plan modifiers on one attribute modify its value, the last modified value should be used.

Yeah, sorry, I think we're getting at the same thing here. I was trying to say I don't know how it would work if we _didn't_ take that approach.

> Some complexity comes with nested attributes - currently it's possible to define `PlanModifiers` at several levels of the nested attribute hierarchy, e.g.:
> 
> [snip]
> 
> Currently, the top-level PlanModifiers are run first, followed by any modifiers on nested attributes.
> 
> This at least needs to be noted in the GoDoc and a test case added, so I'll do that.
> 
> If we like we could also prohibit PlanModifiers on Attributes without a Type, but I don't see a need to do this as long as the order is unambiguous and clearly documented.

Yeah, I'm fine with just "top-level goes first, nested ones follow", that seems like a really reasonable approach and I think the complexity is worth the extra power.

appilon:
Would this feature allow a provider to specify defaults for optional fields?

tfsdk/attribute.go
```diff
@@ -73,6 +73,13 @@ type Attribute struct {
 
 	// Validators defines validation functionality for the attribute.
 	Validators []AttributeValidator
+
+	// PlanModifiers defines a sequence of modifiers for this attribute at
+	// plan time.
+	// Please note that plan modification only applies to resources, not
+	// data sources. Setting PlanModifiers on a data source attribute will
+	// have no effect.
```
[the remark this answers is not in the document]
kmoe: Opened https://github.com/hashicorp/terraform-plugin-framework/issues/113 to track `InternalValidate` options. I think we might be able to get away without it but feel free to add anything else that might be appropriate to that issue.

tfsdk/serve_test.go
kmoe: Thanks, I don't think this removal is needed any more since I got to the bottom of the race conditions. Will restore it.

kmoe:
@appilon Yes, thanks for raising this. Adding a test case to show this working as expected.

A helper func like this makes sense I think:
```go
Schema{
		"size": {
			Optional: true,
			PlanModifiers: []AttributePlanModifier{AttributeDefault(val)},
		},
	},
}
```
where we have either `func AttributeDefault(v interface{}) AttributePlanModifier` or `func AttributeDefault(v attr.Value) AttributePlanModifier`. Similar helper makes sense for a `DefaultFunc` equivalent.

Will follow up with a further PR for any such helpers. Also considering putting `RequiresReplace` and any other helpers in their own package so they are namespaced.

tfsdk/serve_test.go
kmoe: Wow, I have no idea where this completely unrelated change came from. Thanks for spotting it - reverted

tfsdk/serve_test.go
kmoe: Yes, that would work. We could do one of the following:

1. Define a `type  RequiresReplacePaths []*tftypes.AttributePath` in the framework with an `Equal` method that ignores ordering
2. Do something similar but in plugin-go/tftypes itself.

However, I think it might be better just to have the framework server sort them so the response is deterministic, rather than ignoring the change of order during testing. Added 583f552d0530a5c7d4ddff4b551aa3da32eb8cd3 and a8991d73b60056734455d5afaded3611c1204fa0 for this purpose.

kmoe:
Implementation of full recursion of nested attributes works, but the tests are currently running into https://github.com/hashicorp/terraform-plugin-framework/issues/112 as `tftypes.WalkAttributePath` (called in `Plan.terraformValueAtPath()`) cannot handle a nil `tftypes.Object`. I suspect this is a plugin-go fix and will write it up there.

tckb:
+1 waiting for this PR merged - I have the same issue as mentioned in #112

kmoe:
In fact I think this doesn't require a plugin-go change (see comment on #112).

Changes since review:
 - fixed recursion, so attribute plan modifiers run for arbitrarily nested attributes
 - added sorting and deduplication to `resp.RequiresReplace` slice so it is entirely deterministic
 - added test case for setting default attribute values
 - added early return after first plan modifier that errors, instead of continuing with possibly invalid plan values

Ready for review.

tfsdk/attribute_plan_modification.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+// AttributePlanModifiers represents a sequence of AttributePlanModifiers, in
+// order.
+type AttributePlanModifiers []AttributePlanModifier
+
+// RequiresReplace returns an AttributePlanModifier specifying the attribute as
+// requiring replacement. This behaviour is identical to the ForceNew behaviour
+// in terraform-plugin-sdk.
+func RequiresReplace() AttributePlanModifier {
```
paddycarver, line 50: nitpick: if we're going to have a constructor, I feel like the underlying type could be unexported to limit our compatibility surface. If we're going to export the underlying type, I don't know that we need a constructor. 🤔 

(This should not block merge, we can iterate later.)

tfsdk/plan.go
```diff
@@ -167,6 +167,7 @@ func (p *Plan) SetAttribute(ctx context.Context, path *tftypes.AttributePath, va
 func (p Plan) terraformValueAtPath(path *tftypes.AttributePath) (tftypes.Value, error) {
 	rawValue, remaining, err := tftypes.WalkAttributePath(p.Raw, path)
 	if err != nil {
+
```
paddycarver, line 170: nit: extraneous whitespace.

(Should not block merge)

paddycarver (approved):
Awesome work! This looks great to me.

tfsdk/schema.go
```diff
@@ -208,3 +210,120 @@ func (s Schema) validate(ctx context.Context, req ValidateSchemaRequest, resp *V
...
+		if nestedAttrResp.RequiresReplace {
+			resp.RequiresReplace = append(resp.RequiresReplace, attrPath)
+		}
+
+		setAttrDiags := resp.Plan.SetAttribute(ctx, attrPath, nestedAttrResp.AttributePlan)
+		resp.Diagnostics = append(resp.Diagnostics, setAttrDiags...)
+		if diagnostics.DiagsHasErrors(setAttrDiags) {
+			return
+		}
+		resp.Diagnostics = nestedAttrResp.Diagnostics
```
bflad, line 249: Is this last line expected? I _think_ its just extraneous since they should point to the same address, but it could be  confusing as it looks like it might be removing `SetAttribute` diagnostics.

PR #107, opened 2021-08-17
return nil when GetAttribute called on nil state

### LINKED ISSUE #106: Calling GetAttribute() on nil Config/State/Plan throws error
For a `Config` in which the `Schema` is valid and `Raw` is a nil value of the appropriate type, calling `Config.GetAttribute()` with a valid AttributePath throws an error. 

Error is thrown from this line: https://github.com/hashicorp/terraform-plugin-framework/blob/abe64d602ac541e9aa398b10b4b619ccba023686/tfsdk/config.go#L42

Same behaviour for `State` and `Plan`.

An error should not be thrown here. In a nil state/plan/config, the value of any given schema attribute should be nil.

kmoe:
Fixes #106 

Instead of returning an error when GetAttribute is called on a nil state, plan, or config, return nil, provided the attribute path is valid and appears in the schema.

In a nil state/plan/config, the value of any given schema attribute should be nil.

### DIFF
--- a/tfsdk/config.go
+++ b/tfsdk/config.go
@@ -38,6 +38,12 @@ func (c Config) GetAttribute(ctx context.Context, path *tftypes.AttributePath) (
 		})
 	}
 
+	// if the whole config is nil, the value of a valid attribute is also
+	// nil
+	if c.Raw.IsNull() {
+		return nil, nil
+	}
+
 	tfValue, err := c.terraformValueAtPath(path)
 	if err != nil {
 		return nil, append(diags, &tfprotov6.Diagnostic{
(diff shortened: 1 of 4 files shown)


paddycarver (approved):
Yeah, okay, I'd buy that reasoning. My initial :/ is that _technically_ the _attribute_ isn't there, not the _attribute value_, but I don't know that that's a meaningful distinction and feels pretty pedantic when compared to what users are likely to expect, which I think is what this PR is doing.

kmoe:
@bflad I merged this so we can get on with #102, but feel free to comment if you think this isn't the intuitive behaviour and we can certainly change it.

PR #110, opened 2021-08-18
Create diag package and switch all applicable []tfprotov6.Diagnostic usage

### LINKED ISSUE #24: Define Diagnostics type and helpers.

bflad:
Pending https://github.com/hashicorp/terraform-plugin-framework/pull/108
Closes #24 

This is an implementation of diagnostics as an interface type. It is intended to serve as a discussion starting point and may be thrown away or updated depending on the outcomes of #108.


diag/diagnostic.go
```diff
@@ -0,0 +1,49 @@
...
+//
+// See the ErrorDiagnostic and WarningDiagnostic concrete types for generic
+// implementations.
+type Diagnostic interface {
+	// Severity returns the desired level of feedback for the diagnostic.
+	Severity() Severity
+
+	// Summary is a short description for the diagnostic.
+	//
+	// Typically this is implemented as a title, such as "Example Error",
```
paddycarver, line 21: Would it be possible to use a real-world example here? Like "Invalid resource name" or something? I think it may be more illustrative.

diag/diagnostic.go
```diff
@@ -0,0 +1,49 @@
...
+	// practitioners or provider developers.
+	Detail() string
+
+	// Equal returns true if the other diagnostic is wholly equivalent.
+	Equal(Diagnostic) bool
+}
+
+// DiagnosticWithPath is a diagnostic associated with an attribute path.
+//
+// This attribute information used to display contextual source configuration
```
paddycarver, line 39:
```suggestion
// This attribute information is used to display contextual source configuration
```

diag/diagnostic.go
```diff
@@ -0,0 +1,49 @@
...
+// to practitioners.
+type DiagnosticWithPath interface {
+	Diagnostic
+
+	// Path represents the location of the attribute within a schema.
```
paddycarver, line 44: Technically, I think it's within a value, not a schema. Schemas don't have e.g. indexes.

diag/diagnostics.go
```diff
@@ -0,0 +1,97 @@
...
+	diags.Append(NewErrorDiagnostic(summary, detail))
+}
+
+// AddWarning adds a generic warning diagnostic to the collection.
+func (diags *Diagnostics) AddWarning(summary string, detail string) {
+	diags.Append(NewWarningDiagnostic(summary, detail))
+}
+
+// Append adds non-empty and non-duplicate diagnostics to the collection.
+func (diags *Diagnostics) Append(in ...Diagnostic) {
```
paddycarver, line 35: Does this work with `Diagnostics` being returned? E.g.

```go
diags := doTheThing() // diags is now of type Diagnostics
resp.Diagnostics.Append(diags...) // does this work? diags is of type Diagnostics, not []Diagnostic, but Go may convert for us?
```

diag/diagnostics.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	return false
+}
+
+// ToTfprotov6Diagnostics converts the diagnostics into the tfprotov6 collection type.
+//
+// Usage of this method outside the framework is not supported nor considered
+// for backwards compatibility promises.
+func (diags Diagnostics) ToTfprotov6Diagnostics() []*tfprotov6.Diagnostic {
```
paddycarver, line 79: two things:

1) This doesn't appear to require any privileged access to the `diag` package or benefit from being a method on the `Diagnostics` type. Is there a reason we shouldn't just make it a function in `tfsdk`?
2) I think `Tfprotov6` in this may be unnecessarily verbose. `ToProto6Diagnostics` is probably sufficient; we know which protocol we're talking about, and we're unlikely to be talking about any other protocol's diagnostics except Terraform's.

diag/attribute_error_diagnostic.go
```diff
@@ -0,0 +1,41 @@
...
+
+// Equal returns true if the other diagnostic is wholly equivalent.
+func (d AttributeErrorDiagnostic) Equal(other Diagnostic) bool {
+	aed, ok := other.(AttributeErrorDiagnostic)
+
+	if !ok {
+		return false
+	}
+
+	return aed.Summary() == d.Summary() && aed.Detail() == d.Detail() && aed.Path().Equal(d.Path())
```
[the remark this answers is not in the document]
bflad: You're right! Fixed up.

diag/diagnostics.go
bflad: `Diagnostics` is just a type alias for `[]Diagnostic` so `diags...` is equivalent to `...Diagnostic`. See `tfsdk/serve.go` updates for this in action (compiles and unit tests fine).

diag/diagnostics.go
bflad: > This doesn't appear to require any privileged access to the diag package or benefit from being a method on the Diagnostics type. Is there a reason we shouldn't just make it a function in tfsdk?

It does keep the diagnostics implementation altogether.

paddycarver: Hm. Is that worth having an exported method we don't want people to call? 🤔 (I don't know the answer, I'm wondering out loud.)

diag/diagnostics_test.go
```diff
@@ -0,0 +1,527 @@
...
+			tc.diags.Append(tc.in...)
+
+			if diff := cmp.Diff(tc.diags, tc.expected); diff != "" {
+				t.Errorf("Unexpected response (+wanted, -got): %s", diff)
+			}
+		})
+	}
+}
+
+func TestDiagnosticsContains(t *testing.T) {
```
kmoe, line 342: Needs some test cases with attribute paths

.changelog/110.txt
```diff
@@ -0,0 +1,7 @@
+```release-note:breaking-change
+All `[]*tfprotov6.Diagnostic` handling has been converted to `diag.Diagnostics` except where interfacing directly with the underlying protocol type.
```
kmoe, line 2: Can we be more precise about what the breaking change is here? Specifically, is it that functions that used to return `[]*tfprotov6.Diagnostic` now return `diag.Diagnostics`?

bflad: Happy to update as necessary! Some starter ideas:

1. All `[]*tfprotov6.Diagnostic` handling has been converted to `diag.Diagnostics` except where interfacing directly with the underlying protocol type. This includes all function signatures, which will require provider updates. Replacing `[]*tfprotov6.Diagnostic` with `diag.Diagnostics` and import path `github.com/hashicorp/terraform-plugin-go/tfprotov6` with `github.com/hashicorp/terraform-plugin-framework/diag` should suffice in most cases.
1. All `[]*tfprotov6.Diagnostic` handling has been converted to `diag.Diagnostics` except where interfacing directly with the underlying protocol type. Providers will require updates to function signatures and import paths.
1. All `[]*tfprotov6.Diagnostic` handling has been converted to `diag.Diagnostics` except where interfacing directly with the underlying protocol type. Providers will require updates to function signatures (`[]*tfprotov6.Diagnostic` to `diag.Diagnostics`) and import paths (`github.com/hashicorp/terraform-plugin-go/tfprotov6` to `github.com/hashicorp/terraform-plugin-framework/diag`).

diag/diagnostics_test.go
bflad: Added in https://github.com/hashicorp/terraform-plugin-framework/pull/110/commits/473b6e1ac4ac0c34a2ef9faa033bf627b4f2f041 -- please let me know if you want anything else!

.changelog/110.txt
paddycarver:
```suggestion
Most uses of `[]*tfprotov6.Diagnostic` have been replaced with a new `diag.Diagnostics` type. Please update your type signatures, and use one of the `diags.New*` helper functions instead of constructing `*tfprotov6.Diagnostic`s by hand.
```

diag/attribute_error_diagnostic.go
```diff
@@ -0,0 +1,41 @@
+package diag
```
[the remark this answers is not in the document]
bflad: @kmoe not sure if you have any opinions here -- I lean slightly towards keeping this self-contained functionality separate to reduce the cognitive burden of looking at the documentation for one large Go package, but happy to combine if consensus pulls there.

diag/diagnostics.go
bflad: Another option here would be to move that exported function into a separate `internal` package, maybe along the lines of `diag/internal/convert`. I'd be more apt to do this so at least the code is tangential the other diagnostics code. I'd be worried about import cycles if we tried to setup something more generic like `internal/toproto6`.

@kmoe do you have any opinions?

diag/attribute_error_diagnostic.go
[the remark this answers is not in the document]
bflad: (Oh sorry I didn't explicitly answer this -- I think import cycles would be a definite possibility since these types get referenced in `tfsdk` and `internal/reflect`.)

kmoe (approved):

diag/attribute_error_diagnostic.go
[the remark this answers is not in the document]
paddycarver: Thinking on it, I've come around to the conclusion that you're probably right. Let's leave it here for now, and we'll move it later if we want to/can.

diag/diagnostics.go
paddycarver: I'd still prefer this to be unexported as we don't expect provider developers to use it, and I'm worried about what they'll do with this unintended compatibility interface, but I'm not gonna hold this up over it.

tfsdk/serve.go
```diff
@@ -208,12 +220,24 @@ func (s *server) GetProviderSchema(ctx context.Context, _ *tfprotov6.GetProvider
 	resp.ProviderMeta = providerMeta6Schema
 	resp.ResourceSchemas = resource6Schemas
 	resp.DataSourceSchemas = dataSource6Schemas
-	return resp, nil
+}
+
+// validateProviderConfigResponse is a thin abstraction to allow native Diagnostics usage
```
paddycarver, line 225: nitpick: ideally we'd call out _why_ this allows native Diagnostics usage, but we can fast-follow with that.

paddycarver (approved):
This looks good to me. Great work, @bflad. 🚀

PR #118, opened 2021-08-25
Fix `SingleNestedAttribute` validation failure

### LINKED ISSUE #112: SingleNestedAttributes validation failure if the attribute is not present in configuration
Relates #17.
With the latest (fad6afe33058c5a2581d1e424a50c7fc9a842ff3) code, suppose I have a schema like this:

```go
tfsdk.Schema{
	Attributes:  map[string]tfsdk.Attribute{
		"name": {
			Optional:    true,
		},

		"blah_configuration": {
			Attributes: tfsdk.SingleNestedAttributes(
				map[string]tfsdk.Attribute{
					"insights_enabled": {
						Type:        types.BoolType,
						Optional:    true,
					},
					"notifications_enabled": {
						Type:        types.BoolType,
						Optional:    true,
					},
				},
			),
			Optional: true,
		},
	},
}
```

and a configuration like this (`blah_configuration` not specified):

```hcl
resource "blah" "test" {
  name = "testing"
}
```

`terraform plan` errors:

```console
% terraform plan
╷
│ Error: Configuration Read Error
│ 
│   with blah.test,
│   on main.tf line 46, in resource "blah" "test":
│   46: resource "blah" "test" {
│ 
│ An unexpected error was encountered trying to read an attribute from the configuration. This is always an error in the provider. Please report the following to the
│ provider developer:
│ 
│ AttributeName("insights_enabled") still remains in the path: step cannot be applied to this value
╵
╷
│ Error: Configuration Read Error
│ 
│   with blah.test,
│   on main.tf line 46, in resource "blah" "test":
│   46: resource "blah" "test" {
│ 
│ An unexpected error was encountered trying to read an attribute from the configuration. This is always an error in the provider. Please report the following to the
│ provider developer:
│ 
│ AttributeName("notifications_enabled") still remains in the path: step cannot be applied to this value
╵
```

ewbankkit:
Closes #112.

Don't attempt validation if the attribute is missing from configuration.

### DIFF
--- a/.changelog/118.txt
+++ b/.changelog/118.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+tfsdk: Don't attempt validation on a Null-valued `SingleNestedAttribute`
+```
--- a/.changelog/118.txt
+++ b/.changelog/118.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-tfsdk: Don't attempt validation on a Null-valued `SingleNestedAttribute`
+tfsdk: Don't attempt validation on the nested attributes of a null or unknown `SingleNestedAttribute`
 ```
--- a/tfsdk/attribute.go
+++ b/tfsdk/attribute.go
@@ -355,7 +355,7 @@ func (a Attribute) validate(ctx context.Context, req ValidateAttributeRequest, r
 				return
 			}
 
-			if len(o.Attrs) > 0 {
+			if !o.Null {
 				for nestedName, nestedAttr := range a.Attributes.GetAttributes() {
 					nestedAttrReq := ValidateAttributeRequest{
 						AttributePath: req.AttributePath.WithAttributeName(nestedName),
--- a/tfsdk/attribute.go
+++ b/tfsdk/attribute.go
@@ -355,7 +355,7 @@ func (a Attribute) validate(ctx context.Context, req ValidateAttributeRequest, r
 				return
 			}
 
-			if !o.Null {
+			if !o.Null && !o.Unknown {
 				for nestedName, nestedAttr := range a.Attributes.GetAttributes() {
 					nestedAttrReq := ValidateAttributeRequest{
 						AttributePath: req.AttributePath.WithAttributeName(nestedName),
--- a/tfsdk/attribute.go
+++ b/tfsdk/attribute.go
@@ -341,18 +341,34 @@ func (a Attribute) validate(ctx context.Context, req ValidateAttributeRequest, r
 				}
 			}
 		case NestingModeSingle:
-			for nestedName, nestedAttr := range a.Attributes.GetAttributes() {
-				nestedAttrReq := ValidateAttributeRequest{
-					AttributePath: req.AttributePath.WithAttributeName(nestedName),
-					Config:        req.Config,
-				}
-				nestedAttrResp := &ValidateAttributeResponse{
-					Diagnostics: resp.Diagnostics,
-				}
+			o, ok := req.AttributeConfig.(types.Object)
 
-				nestedAttr.validate(ctx, nestedAttrReq, nestedAttrResp)
+			if !ok {
+				err := fmt.Errorf("unknown attribute value type (%T) for nesting mode (%T) at path: %s", req.AttributeConfig, nm, req.AttributePath)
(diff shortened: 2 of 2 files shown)


kmoe (approved):

tfsdk/attribute.go
```diff
@@ -341,18 +341,34 @@ func (a Attribute) validate(ctx context.Context, req ValidateAttributeRequest, r
...
+					Summary:   "Attribute Validation Error",
+					Detail:    "Attribute validation cannot walk schema. Report this to the provider developer:\n\n" + err.Error(),
+					Attribute: req.AttributePath,
+				})
+
+				return
+			}
 
-				resp.Diagnostics = nestedAttrResp.Diagnostics
+			if len(o.Attrs) > 0 {
```
paddycarver, line 358: Is there a reason this became necessary?

The for loop should cover this case, by just not entering the for loop, which it looks like is what this does?

Unless this is to cover the case where the schema has attributes defined, but Terraform sends a value omitting those attributes, which as far as I know, should never happen?

paddycarver (changes requested):
Could we also add a .changelog entry? https://github.com/hashicorp/engineering-docs/blob/master/terraform/sdk/changelog.md

(I know that link is to a private repo, I'll open an issue to get it added to our contributing guide)

tfsdk/attribute.go
ewbankkit: Correct, I'll remove the _if_.

ewbankkit: I think, as @kmoe mentions in https://github.com/hashicorp/terraform-plugin-framework/issues/112#issuecomment-905493498, it's the call to `tftypes.WalkAttributePath` in `Config.GetAttribute` called from `Attribute.validate` on the nested `Null` Attribute value that fails.

paddycarver: Could we use `o.Null` then, instead of checking the number of attributes?

ewbankkit: Let me take a look...

ewbankkit: Yes `!o.Null` works just as well.
Commit pushed.

avinashbot: Just tested this and hit a failure with unknown attributes; it looks like you'll need to add a check for `!o.Unknown` too.

ewbankkit: will do; thanks

.changelog/118.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+tfsdk: Don't attempt validation on a Null-valued `SingleNestedAttribute`
```
paddycarver, line 2:
```suggestion
tfsdk: Don't attempt validation on the nested attributes of a null or unknown `SingleNestedAttribute`
```

I think the validators on the attribute itself will still run, right? This is just for the nested attributes of that attribute?

paddycarver (changes requested):

.changelog/118.txt
ewbankkit: Correct.

tfsdk/attribute.go
```diff
@@ -341,18 +341,34 @@ func (a Attribute) validate(ctx context.Context, req ValidateAttributeRequest, r
...
-				}
-				nestedAttrResp := &ValidateAttributeResponse{
-					Diagnostics: resp.Diagnostics,
-				}
+			o, ok := req.AttributeConfig.(types.Object)
 
-				nestedAttr.validate(ctx, nestedAttrReq, nestedAttrResp)
+			if !ok {
+				err := fmt.Errorf("unknown attribute value type (%T) for nesting mode (%T) at path: %s", req.AttributeConfig, nm, req.AttributePath)
+				resp.Diagnostics = append(resp.Diagnostics, &tfprotov6.Diagnostic{
```
paddycarver, line 348: Really sorry to do this to you, @ewbankkit, but could you update to use the helpers added by #110? Just want to keep our consistency @bflad worked so hard to introduce. 😦 

Sorry, ping me in chat when you do that and we'll get it merged straightaway before something else comes up. Thanks so much for the patience on this.

paddycarver (changes requested):

PR #119, opened 2021-09-01
Add a Type method to attr.Value, add tfsdk.ValueAs.

### LINKED ISSUE #88: Add support for attr.ValueAs
### Module version

```
v0.2.0
```

### Use-cases

As described in #69.

### Proposal

As described in #69.

paddycarver:
Implement a new `tfsdk.ValueAs` helper that uses the same reflection
found in `Get` and `GetAttribute` to coerce an `attr.Value` to a
concrete Go type. It would be `attr.ValueAs`, but we can't do that, as
it would create an import cycle between `internal/reflect` and `attr`.
Adding this functionality allows people to avoid type-asserting on their
`attr.Value`s, instead using a less-brittle approach.

To achieve this, we needed to supply the `attr.Type` of the `attr.Value`
to `reflect.Into`, because it needs that if it's going to be able to
reflect into `attr.Value`s (because not all of them can be modified in
place, meaning we need to defensively re-create them, meaning we need to
know the `attr.Type` used to generate them). Rather than making users
supply the `attr.Type` when calling `tfsdk.ValueAs`, a `Type` method was
added to `attr.Value`, returning the `attr.Type` that generates that
`attr.Value`.

This involved a lot of updates to the tests, because they assumed that
the test types could generate non-test values, but the non-test values
weren't pointing to the right types, then. So when they were reflected,
they threw errors, because the new generated `attr.Value` wasn't the
same Go type as the `attr.Value` they were being reflected into. A mess
all around, but it was fixed to have test versions of the `attr.Value`s
that return the appropriate types for `attr.Value.Type`.

This `attr.Value` interface change _does_ make it harder to compose new
`attr.Type` and `attr.Value` implementations on top of older ones,
because it kinda requires the `attr.Value` to know about the `attr.Type`
that produced it. It's not _much_ extra work--the `Equal` and `Type`
methods need to be overridden--but it is still extra work. I'm not sure
how to avoid this, however, without pushing the complexity onto the
provider developer when `tfsdk.ValueAs` is used, by making them specify
the `attr.Type`. And that (intuitively, not evidenced in data) seems
like a more frequent paper cut than the paper cut of composing
`attr.Type` and `attr.Value` implementations, outside the relatively
bizarre case of the framework tests.

Fixes #88.

### DIFF
--- a/.changelog/119.txt
+++ b/.changelog/119.txt
@@ -1,0 +1,7 @@
+```release-note:breaking-change
+`attr.Value` implementations must now implement a `Type(context.Context)` method that returns the `attr.Type` that created the `attr.Value`.
+```
+
+```release-note:enhancement
+Added a `tfsdk.ValueAs` helper that allows accessing an `attr.Value` without type assertion, by using the same reflection rules used in the `Config.Get`, `Plan.Get`, and `State.Get` helpers.
+```
--- a/attr/value.go
+++ b/attr/value.go
@@ -8,6 +8,9 @@ import (
 // Values allow provider developers to specify data in a convenient format, and
 // have it transparently be converted to formats Terraform understands.
 type Value interface {
+	// Type returns the Type that created the Value.
+	Type(context.Context) Type
+
 	// ToTerraformValue returns the data contained in the Value as
 	// a Go type that tftypes.NewValue will accept.
 	ToTerraformValue(context.Context) (interface{}, error)
--- a/internal/reflect/diags.go
+++ b/internal/reflect/diags.go
@@ -1,6 +1,10 @@
 package reflect
 
 import (
+	"fmt"
+	"reflect"
+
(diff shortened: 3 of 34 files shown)


paddycarver:
This still needs tests for `tfsdk.ValueAs`.

bflad (approved):
This looks good to me 🚀  #120 and #134 will be really good followups. 👍

kmoe:
It would be great to see the `ValueAs()` function used in the plan modifier specified in `serve_resource_attribute_plan_modifiers_test.go`, so we can remove the type assertion there (lines 40-41)

paddycarver:
Just pushed a commit. I had to make `RequiresReplaceIf` return a `diag.Diagnostics` because that's what `ValueAs` returns, and if we're going to return a diagnostic, people should have access to the path to make the diagnostics _good_, so I added that, too.

kmoe (approved):

PR #120, opened 2021-09-01
Add an attr.Convert helper.

paddycarver:
This function converts an `attr.Value` to use a new `attr.Type`, so long
as they share a compatible `tftypes.Type` representation. This allows
code that doesn't know the type of an `attr.Value` to still use it,
provided they know the kind of data it holds.

For example, a validation helper that works on strings--MaxStringLength,
say--could look something like this:

```go
// assume req.Config is an attr.Value
strAttr, err := attr.Convert(ctx, req.Config, types.StringType)
if err != nil {
  panic(err)
}
var str types.String
diags := tfsdk.ValueAs(ctx, strAttr, &str)
if diags.HasError() {
  panic("oops")
}
// do something with `str`, which is now strongly typed
```

Then no matter what `attr.Type` provider devs use in their schema, this
validator would still work, then. Without the conversion step, it
wouldn't work, as `tfsdk.ValueAs` implementation needs to know the
`attr.Type` of the `attr.Value`, and reusable validators _can't_ know
that. So the point of this helper is to get an `attr.Value` into a known
`attr.Type` format.

This still needs tests.


paddycarver:
Depends on #119.

paddycarver:
Still depends on #119, but has tests and is otherwise code complete.

kmoe (approved):

Issue #121, opened 2021-09-01
tfsdk: TestServerPlanResourceChange/attr_plan_modifiers_diags Race Condition


bflad:
### Module version

Post #102 merge

### Relevant framework source code

```go
func (rt testServeResourceTypeAttributePlanModifiers) GetSchema(_ context.Context) (Schema, diag.Diagnostics) {
	return Schema{
		Version: 1,
		Attributes: map[string]Attribute{
			"name": {
				Required: true,
				Type:     types.StringType,
				// For the purposes of testing, these plan modifiers behave
				// differently for certain values of the attribute.
				// By default, they do nothing.
				PlanModifiers: []AttributePlanModifier{
					testWarningDiagModifier{},
					testErrorDiagModifier{},
					testAttrPlanValueModifierOne{},
					testAttrPlanValueModifierTwo{},
				},
			},
			// ... other Attribute ...
		},
	}, nil
}
func (t testWarningDiagModifier) Modify(ctx context.Context, req ModifyAttributePlanRequest, resp *ModifyAttributePlanResponse) {
	attrVal, ok := req.AttributePlan.(types.String)
	if !ok {
		return
	}

	if attrVal.Value == "TESTDIAG" {
		resp.Diagnostics.AddWarning(
			"Warning diag",
			"This is a warning",
		)
	}
}

func (t testErrorDiagModifier) Modify(ctx context.Context, req ModifyAttributePlanRequest, resp *ModifyAttributePlanResponse) {
	attrVal, ok := req.AttributePlan.(types.String)
	if !ok {
		return
	}

	if attrVal.Value == "TESTDIAG" {
		resp.Diagnostics.AddError(
			"Error diag",
			"This is an error",
		)
	}
}
```

```
"attr_plan_modifiers_diags": {
			priorState: tftypes.NewValue(testServeResourceTypeAttributePlanModifiersType, map[string]tftypes.Value{
				"name": tftypes.NewValue(tftypes.String, "TESTDIAG"),
				"size": tftypes.NewValue(tftypes.Number, 3),
				"scratch_disk": tftypes.NewValue(tftypes.Object{AttributeTypes: map[string]tftypes.Type{
					"id":        tftypes.String,
					"interface": tftypes.String,
					"filesystem": tftypes.Object{AttributeTypes: map[string]tftypes.Type{
						"size":   tftypes.Number,
						"format": tftypes.String,
					}},
				}}, map[string]tftypes.Value{
[...]

PR #123, opened 2021-09-01
tfsdk: Ensure other attribute plan modifiers are called after errors

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/121

This is to prevent situations where warning diagnostics in other attributes are skipped based on when an error diagnostic might be non-deterministically found.


paddycarver (approved):

kmoe:
Call an error in provider code that should halt execution of `PlanResourceChange` handling a _fatal error_: for example, the API client gets an expired token response or hits a rate limit, so it is not possible to complete the operation. 

The tradeoff here is now that the framework must take one of the following options:

1. The first fatal error encountered in any attribute plan modifier halts execution of plan modifiers. Since the order the attributes are visited in is currently random, this leads to inconsistent return values.

2. The first fatal error encountered in any plan modifier halts execution, and attributes are always visited in the same order, so the return value is deterministic.

3. Plan modifiers cannot throw fatal errors - error diags do not halt execution.

I believe this PR is proposing option 3, which matches the validation functionality, in which attribute validators cannot throw fatal errors. This seems fine - we just need to make sure we document what it means to return an error diag so users know what to expect.

None of these options is obvious in my opinion, so let's mention in the GoDoc for `Attribute.PlanModifiers` whether the user should expect other plan modifiers to run if one errors.

bflad:
That is an excellent write-up, @kmoe! Let's sync up later today to discuss this further.

paddycarver:
It seems to me it's worth calling out the difference between attribute-level and resource-level fatal.

The current implementation has a _resource-level_ fatal. It stops not only plan modification for that attribute, but for all other attributes.

The proposed implementation has an _attribute-level_ fatal. It stops plan modification for that attribute, but other attributes continue to have _their_ plan modification run.

The only time I can think of that the proposed implementation would have unexpected consequences for provider developers is when a nested attribute's plan customization did not fully complete before its parent's plan customization is run. Can anyone else think of scenarios in which the proposed behavior would be surprising?

kmoe:
Good distinction. Putting it this way makes it even clearer to me that the right approach is to make attribute plan modifier errors attribute-level fatal.

I can't think of a situation in which this behaviour would be surprising or unwelcome as long as it's clearly documented. Happy to approve this PR with something added to the GoDoc for this.

tfsdk/attribute.go
```diff
@@ -75,10 +75,16 @@ type Attribute struct {
...
+	// plan time. Attribute-level plan modifications occur before any
+	// resource-level plan modifications.
+	//
+	// Any errors will prevent further execution of this sequence
+	// of modifiers and modifiers associated with nested Attribute, but have no
+	// effect with other Attribute in the Schema.
+	//
+	// Plan modification only applies to resources, not data sources or
+	// providers. Setting PlanModifiers on a data source or provider attribute
+	// will have no effect.
```
bflad, line 87: I must admit that this makes me 🤔  about `Attribute` being an interface and having `ProviderAttribute`, `ResourceAttribute`, and `DataSourceAttribute` to distinguish available functionality between these at compiler-time.

tfsdk/attribute.go
```diff
@@ -75,10 +75,16 @@ type Attribute struct {
...
 	// PlanModifiers defines a sequence of modifiers for this attribute at
-	// plan time.
-	// Please note that plan modification only applies to resources, not
-	// data sources. Setting PlanModifiers on a data source attribute will
-	// have no effect.
+	// plan time. Attribute-level plan modifications occur before any
+	// resource-level plan modifications.
+	//
+	// Any errors will prevent further execution of this sequence
+	// of modifiers and modifiers associated with nested Attribute, but have no
```
kmoe, line 82:
```suggestion
	// of modifiers and modifiers associated with any nested Attribute, but will not
```

tfsdk/attribute.go
```diff
@@ -75,10 +75,16 @@ type Attribute struct {
...
+	// effect with other Attribute in the Schema.
```
kmoe, line 83:
```suggestion
	// prevent execution of PlanModifiers on any other Attribute in the Schema.
```

tfsdk/attribute.go
kmoe: Sounds like a good idea. I agree it's confusing having a field on the `Attribute` struct that sometimes does nothing.

kmoe (approved):
LGTM with some suggestions for wording

bflad:
It looks like `(Attribute).modifyPlan()` also needed to be taught to ignore previous errors to ensure attribute-level plan modifiers were not skipped. To ensure the expected behaviors of that method, added exhaustive unit testing to cover all the branching logic.

tfsdk/attribute.go
bflad: Created https://github.com/hashicorp/terraform-plugin-framework/issues/132

paddycarver (approved):

PR #126, opened 2021-09-02
types: Support Set and SetType

### LINKED ISSUE #53: Define Set type
Backed by an actual Set implementation.

bflad:
Closes #53


bflad:
I'm going to let things settle with #24, #127, and #128 (all v0.3.0 candidates) before continuing here, personally.

bflad:
My upfront guess is that this implementation may need to change into allowing duplicates by overwriting existing elements during `Set()` and `SetAttribute()` calls instead of returning a duplicate set element validation error. I guess we will want to decide on what the more correct behavior should be -- potentially allowing silently successful duplicates where they may not be intended or potentially causing unnecessary errors with provider implementations that will want the type to just consolidate data automatically.

My first impression is that the current implementation may actually be desirable in this case (and allow the implementation to be much simpler) since practitioners would receive configuration validation errors for duplicate set elements and API fields this is being used against shouldn't have duplicates if this attribute type is being chosen.

I'm not sure if its worth capturing this context anywhere (maybe on the `Set`/`SetType` documentation) no matter which way we decide to go.

types/set.go
```diff
@@ -0,0 +1,276 @@
...
+		err := fmt.Errorf("expected Set value, received %T with value: %v", in, in)
+		diags.AddAttributeError(
+			path,
+			"Set Type Validation Error",
+			"An unexpected error was encountered trying to validate an attribute value. This is always an error in the provider. Please report the following to the provider developer:\n\n"+err.Error(),
+		)
+		return diags
+	}
+
+	if !in.IsKnown() {
```
paddycarver, line 119: Could probably return early here, too, if `in` is null.

types/set.go
```diff
@@ -0,0 +1,276 @@
...
+		}
+
+		for indexInner, elemInner := range elems {
+			if indexInner == indexOuter {
+				continue
+			}
+
+			if !elemInner.Equal(elemOuter) {
+				continue
+			}
```
paddycarver, line 150: This is going to compare each value to each other _twice_. For example, when `indexOuter` is `2` and `indexInner` is `6`, they'll be compared, and when `indexOuter` is `6` and `indexInner` is `2`, they'll be compared.

Could we instead _start_ `indexOuter` at `indexInner+1`, on the understanding the element at `indexOuter` has already been compared to each element before it, when _they_ were `indexOuter`? That may help the quadratic nature of this function. (I don't know that it's a _big_ deal--we should be working with small enough numbers here, in the thousands, that I think even a quadratic function is probably fine, but thought I'd raise it.)

types/set.go
```diff
@@ -0,0 +1,276 @@
...
+	if s.Unknown != other.Unknown {
+		return false
+	}
+	if s.Null != other.Null {
+		return false
+	}
+	if s.ElemType == nil && other.ElemType != nil {
+		return false
+	}
+	if s.ElemType != nil && !s.ElemType.Equal(other.ElemType) {
```
paddycarver, line 254: It looks like it's possible here for `other.ElemType` to be nil. Do we want to check that before calling `Equal`?

types/set.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+	for _, elem := range s.Elems {
+		if !other.contains(elem) {
+			return false
+		}
+	}
+	return true
+}
+
+func (s Set) contains(v attr.Value) bool {
```
paddycarver, line 268: This seems like a useful function to export. Do we want to export it?

types/set.go
```diff
@@ -0,0 +1,276 @@
...
+			)
+		}
+	}
+
+	return diags
+}
+
+// Set represents a set of AttributeValues, all of the same type, indicated
+// by ElemType.
+type Set struct {
```
paddycarver, line 165: Do we want to add helper methods for:

```go
// Add inserts an element into the Set, throwing an error if it's already present
func (s Set) Add(ctx context.Context, el attr.Value) error
```

(we can use a well-known error for "value was already set", so provider devs can ignore it if they want)

```go
// Remove removes the given value from the Set, throwing an error if it wasn't present
func (s Set) Remove(ctx context.Context, el attr.Value) error
```

?

types/set_test.go
```diff
@@ -0,0 +1,843 @@
...
+	}{
+		"null": {
+			in: tftypes.NewValue(
+				tftypes.Set{
+					ElementType: tftypes.String,
+				},
+				nil,
+			),
+		},
+		"null-element": {
```
paddycarver, line 315: This seems like an unknown element, not a null one?

types/set_test.go
```diff
@@ -0,0 +1,843 @@
...
+			in: tftypes.NewValue(
+				tftypes.Set{
+					ElementType: tftypes.String,
+				},
+				[]tftypes.Value{
+					tftypes.NewValue(tftypes.String, tftypes.UnknownValue),
+				},
+			),
+		},
+		"null-elements": {
```
paddycarver, line 325: Same here, these seem like they're unknown, not null.

paddycarver (changes requested):
Looks good. Some small additions I recommended, and looks like some typos in some tests.

types/set.go
bflad: We can certainly check that explicitly -- we should update `List` to match if we are changing this here.

types/set.go
bflad: Is manually working with a framework-defined `Set` a goal for our implementation or do we want to say that is an implementation detail for outside the framework? Currently developers can either build this manually via `Elems` or use reflection, similar to `List`. If we add methods for working with `Set` via framework-defined type methods, my next question would be should we offer these methods for the `List` and `Map` types for consistency?

types/set.go
bflad: We can track decision making in https://github.com/hashicorp/terraform-plugin-framework/pull/126#discussion_r707360807, which will influence the answer here.

types/set_test.go
bflad: Good catch, copypasta 🍝  -- fixed in next commit.

types/set_test.go
bflad: Samesies 🍝  fixed in next commit.

types/set.go
bflad: Adding:

```go
	if other.ElemType == nil {
		return false
	}
```

Prevents `Set{}` == `Set{}` comparisons in testing:

```go
expected: testStateGetDataTestTypes{
	// ...
	Tags:        types.List{},
	TagsSet:     types.Set{},
	// ...
},
```

```
--- FAIL: TestStateGet_testTypes (0.00s)
    --- FAIL: TestStateGet_testTypes/AttrTypeWithValidateError (0.00s)
        state_test.go:846: unexpected value (+wanted, -got):   tfsdk.testStateGetDataTestTypes{
                Name:        {},
                MachineType: "",
                Tags:        {},
            -   TagsSet:     "github.com/hashicorp/terraform-plugin-framework/types".Set{},
            +   TagsSet:     "github.com/hashicorp/terraform-plugin-framework/types".Set{},
                Disks:       nil,
                DisksSet:    nil,
                ... // 2 identical fields
              }
FAIL
```

types/set.go
bflad: I added benchmark tests for this method:

```go
func benchmarkSetTypeValidate(b *testing.B, elementCount int) {
	elements := make([]tftypes.Value, 0, elementCount)

	for idx := range elements {
		elements[idx] = tftypes.NewValue(tftypes.String, strconv.Itoa(idx))
	}

	ctx := context.Background()
	in := tftypes.NewValue(
		tftypes.Set{
			ElementType: tftypes.String,
		},
		elements,
	)
	path := tftypes.NewAttributePath().WithAttributeName("test")
	set := SetType{}

	for n := 0; n < b.N; n++ {
		set.Validate(ctx, in, path)
	}
}

func BenchmarkSetTypeValidate10(b *testing.B) {
	benchmarkSetTypeValidate(b, 10)
}

func BenchmarkSetTypeValidate100(b *testing.B) {
	benchmarkSetTypeValidate(b, 100)
}

func BenchmarkSetTypeValidate1000(b *testing.B) {
	benchmarkSetTypeValidate(b, 1000)
}

func BenchmarkSetTypeValidate10000(b *testing.B) {
	benchmarkSetTypeValidate(b, 10000)
}

func BenchmarkSetTypeValidate100000(b *testing.B) {
	benchmarkSetTypeValidate(b, 100000)
}

func BenchmarkSetTypeValidate1000000(b *testing.B) {
	benchmarkSetTypeValidate(b, 1000000)
}
```

The inner loop optimization seems like a good one, so adjusted the implementation:

```go
	for indexOuter, elemOuter := range elems {
		// Only evaluate fully known values for duplicates.
		if !elemOuter.IsFullyKnown() {
			continue
		}

		for indexInner := indexOuter + 1; indexInner < len(elems); indexInner++ {
			elemInner := elems[indexInner]

			if !elemInner.Equal(elemOuter) {
				continue
			}

			diags.AddAttributeError(
				path.WithElementKeyValue(elemInner),
				"Duplicate Set Element",
				fmt.Sprintf("This attribute contains duplicate values of: %s", elemInner),
			)
		}
	}
```

Before and after code change:

```console
$ go test -bench=. -benchtime=10s ./types
goos: darwin
goarch: amd64
pkg: github.com/hashicorp/terraform-plugin-framework/types
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
BenchmarkSetTypeValidate10-16           123664647               97.27 ns/op
BenchmarkSetTypeValidate100-16          123715761               96.50 ns/op
BenchmarkSetTypeValidate1000-16         124075330               97.07 ns/op
BenchmarkSetTypeValidate10000-16        123091288               97.36 ns/op
BenchmarkSetTypeValidate100000-16       120266646               99.68 ns/op
BenchmarkSetTypeValidate1000000-16      124239894               94.87 ns/op
PASS
ok      github.com/hashicorp/terraform-plugin-framework/types   131.062s
```

```console
$ go test -bench=. -benchtime=10s ./types
goos: darwin
goarch: amd64
pkg: github.com/hashicorp/terraform-plugin-framework/types
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
BenchmarkSetTypeValidate10-16           124338459               95.79 ns/op
BenchmarkSetTypeValidate100-16          124734000               95.74 ns/op
BenchmarkSetTypeValidate1000-16         124954414               97.24 ns/op
BenchmarkSetTypeValidate10000-16        122782779               97.07 ns/op
BenchmarkSetTypeValidate100000-16       120025662               99.47 ns/op
BenchmarkSetTypeValidate1000000-16      126757616               94.44 ns/op
PASS
ok      github.com/hashicorp/terraform-plugin-framework/types   131.266s
```

paddycarver: Those sweet, sweet nanoseconds, though! :D Yeah, that jibes with my suspicions.

types/set.go
paddycarver: Does

```go
if s.ElemType != nil && other.ElemType == nil {
  return false
}
if s.ElemType == nil && other.ElemType == nil {
  return true
}
```

fix it?

And yeah, agreed, if this turns out to be a thing we need to fix here, we should mirror it if necessary on List. :/

types/set.go
paddycarver: You're right, let's make this a follow-up enhancement and apply it to all the types.

types/set.go
paddycarver: Actually, we don't need to do anything here at all. If `other.ElemType` is `nil`, it will pass in as `nil` to the `Equal` method, which the type assertion should catch.

paddycarver (approved):
Yeah, let's do it. Good work! 👍 🚀

PR #129, opened 2021-09-03
attr: Add attribute path parameter to (TypeWithValidate).Validate() method

### LINKED ISSUE #127: Add Attribute Path for Type Validator Diagnostics
### Module version

Pre v0.3.0 with validators support

### Use-cases

Diagnostics from `attr.TypeWithValidate` `Validate()` methods should have the opportunity to return the attribute path associated with the diagnostic, to ensure the feedback is clear for practitioners and provider developers.

### Attempted Solutions

Currently diagnostics must must manually re-associate the value as a string or establish a meaningful context in the diagnostic summary/detail.

### Proposal

Before releasing v0.3.0, make the breaking change to add `*tftypes.AttributePath` to the `attr.TypeWithValidate` interface type `Validate()` method signature:

```go
type TypeWithValidate interface {
	Type

	Validate(context.Context, tftypes.Value, *tftypes.AttributePath) []*tfprotov6.Diagnostic
}
```

All `Validate()` calling code currently has access to the `*tftypes.AttributePath`.

Other changes from https://github.com/hashicorp/terraform-plugin-framework/issues/24 and https://github.com/hashicorp/terraform-plugin-framework/issues/81 may further change this method signature.

### References

Found while looking at #53 and #99.

bflad:
Closes #127

No CHANGELOG is necessary as this functionality has not been released yet.

### DIFF
--- a/attr/type.go
+++ b/attr/type.go
@@ -80,7 +80,7 @@ type TypeWithValidate interface {
 	// being used to populate the Type. It is generally used to check the
 	// data format and ensure that it complies with the requirements of the
 	// Type.
-	Validate(context.Context, tftypes.Value) diag.Diagnostics
+	Validate(context.Context, tftypes.Value, *tftypes.AttributePath) diag.Diagnostics
 }
 
 // TypeWithPlaintextDescription extends the Type interface to include a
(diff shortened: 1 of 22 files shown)


bflad:
Rebased on top of diagnostics refactoring.

paddycarver (approved):

Issue #33, opened 2021-06-01
Support Managed Resource Import


paddycarver:
### Module version

```
N/A
```

### Use-cases

Practitioners sometimes need to have Terraform adopt infrastructure that already exists, rather than creating the infrastructure with Terraform. Providers need to be able to support this, accepting a practitioner-provided ID and resource type and creating that resource in state. The framework should provide hooks for this.

### Attempted Solutions

N/A

### Proposal

This probably looks like an optional Import method that either Resources or ResourceTypes can implement (undecided on which, honestly).

paddycarver:
Context on this one: we only really get [the ID the user entered](https://github.com/hashicorp/terraform-plugin-go/blob/v0.3.0/tfprotov6/resource.go#L408-L417) and are expected to return [a list of states](https://github.com/hashicorp/terraform-plugin-go/blob/v0.3.0/tfprotov6/resource.go#L421-L451).

My first instinct would be:

* It gets put on the `Resource` interface, because it's likely that provider developers will want access to the configured API client.
* It gets a request with the ID
  * we may get more information at some point in the future if Terraform upgrades its concept of importing resources.
* It's expected to modify a response argument, populating a list of states
  * To achieve this, we may need to add a helper function like `func NewState(context.Context, ResourceType, attr.Value) (State, Diagnostics)`, because we've never asked people to _construct_ states before, only supply `attr.Value`s to fill them.

But that's just a naive first glance at the issue, there may be a more elegant answer here.

PR #130, opened 2021-09-03
docs/design: Initial import design document

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/33

NOTE: This includes super 🌶️  takes from painful personal experiences with multiple resource import support. I really do think it might be worth optimizing for the single import case as a first implementation, but as always, happy to discuss that opinion more at length!


docs/design/import.md
```diff
@@ -0,0 +1,314 @@
...
+
+This design documentation will walk through and recommend options for import handling in the framework.
+
+## Background
+
+### Terraform CLI
+
+Terraform CLI supports two forms of interacting with unmanaged resources:
+
+- [`terraform add`](https://www.terraform.io/docs/cli/commands/add.html) (*Experimental*, Terraform CLI version 1.1.0 and later): Given a resource address and using the resource schema provided by the provider, generate a configuration template.
```
bflad, line 13: This isn't live yet, so could switch it to something like https://github.com/hashicorp/terraform/blob/43d753e727c4248e8d5bf8a703b1937f75c55c76/website/docs/cli/commands/add.html.md, if desired.

docs/design/import.md
```diff
@@ -0,0 +1,314 @@
+# Import
+
+Practitioners may have previously existing resources that they wish to bring under Terraform'm management. Conceptually, this is called importing as Terraform must create the state of the resource. Providers are responsible for receiving a request for resource import and respond with the relevant resource state(s).
```
kmoe, line 3:
```suggestion
Practitioners may have previously existing resources that they wish to bring under Terraform's management. Conceptually, this is called importing as Terraform must create the state of the resource. Providers are responsible for receiving a request for resource import and respond with the relevant resource state(s).
```

kmoe (commented):
Great design doc. I agree with your conclusions. Following up on multiple resource handling out of band.

kmoe (approved):

docs/design/import.md
```diff
@@ -0,0 +1,314 @@
...
+
+Practitioners may have previously existing resources that they wish to bring under Terraform's management. Conceptually, this is called importing as Terraform must create the state of the resource. Providers are responsible for receiving a request for resource import and respond with the relevant resource state(s).
+
+This design documentation will walk through and recommend options for import handling in the framework.
+
+## Background
+
+### Terraform CLI
+
+Terraform CLI supports two forms of interacting with unmanaged resources:
```
paddycarver, line 11: nitpick

```suggestion
Terraform CLI supports two forms of interacting with resources not in its statefile:
```

"unmanaged" here may be confusing. Does it mean data sources? etc.

docs/design/import.md
```diff
@@ -0,0 +1,314 @@
...
+```protobuf
+service Provider {
+    // ... other RPCs ...
+    rpc ImportResourceState(ImportResourceState.Request) returns (ImportResourceState.Response);
+}
+
+message ImportResourceState {
+    message Request {
+        string type_name = 1;
+        string id = 2;
```
paddycarver, line 31: It may be helpful somewhere to call out how practitioners see this, e.g.,

```sh
$ terraform import $ADDRESS $ID
```

and to talk about what the ID is/needs to be (i.e., enough information to identify the resource or resources in the API).

docs/design/import.md
```diff
@@ -0,0 +1,314 @@
...
+
+Otherwise, custom provider implementations were required.
+
+## Caveats
+
+Terraform supports the ability to import multiple resource states during a single import (for example, resource import of an `aws_s3_bucket` also imports any associated `aws_s3_bucket_policy`), however almost all resources in the Terraform ecosystem with import support implement a 1:1 import. The extremely uncommon multiple resource support in practice is likely due to:
+
+- Risk of resource destruction for incorrect implementations ([example issue reference](https://github.com/hashicorp/terraform-provider-aws/issues/6036)).
+- Pracitioners may not have required permissions or wish to manage secondary imported resources ([example issue reference](https://github.com/hashicorp/terraform-provider-aws/issues/9508)).
+- Providers must define the resource addresses for secondary resources, which may not match practitioner desires and require additional action such as `state mv` ([example issue reference](https://github.com/hashicorp/terraform-provider-aws/issues/9001)).
```
paddycarver, line 181: This feels like an important thing to call out. How does this work currently? It sounds like the provider can supply the resource address, but I'm not clear on how. Can we tease out this bit a little more to be more concrete about what the state of the world is right now?

docs/design/import.md
```diff
@@ -0,0 +1,314 @@
...
+Importer: &schema.ResourceImporter{
+    State: schema.ImportStatePassthrough,
+},
+```
+
+Otherwise, custom provider implementations were required.
+
+## Caveats
+
+Terraform supports the ability to import multiple resource states during a single import (for example, resource import of an `aws_s3_bucket` also imports any associated `aws_s3_bucket_policy`), however almost all resources in the Terraform ecosystem with import support implement a 1:1 import. The extremely uncommon multiple resource support in practice is likely due to:
```
paddycarver, line 177: This last sentence

> The extremely uncommon multiple resource support in practice is likely due to:

sounded to me like a list of reasons why people would want to do that was going to follow, but it looks like a list of why that's a bad idea to do? Or why it's hard to use? I'm not clear.

docs/design/import.md
```diff
@@ -0,0 +1,314 @@
...
+
+- Risk of resource destruction for incorrect implementations ([example issue reference](https://github.com/hashicorp/terraform-provider-aws/issues/6036)).
+- Pracitioners may not have required permissions or wish to manage secondary imported resources ([example issue reference](https://github.com/hashicorp/terraform-provider-aws/issues/9508)).
+- Providers must define the resource addresses for secondary resources, which may not match practitioner desires and require additional action such as `state mv` ([example issue reference](https://github.com/hashicorp/terraform-provider-aws/issues/9001)).
+- (Previously) Risk of resource destruction for missing configurations ([example issue reference](https://github.com/hashicorp/terraform-provider-aws/issues/9001)).
+- (Previously) Terraform CLI unable to refresh secondary resource states using the correct provider instance ([example issue reference](https://github.com/hashicorp/terraform-provider-aws/issues/394)).
+
```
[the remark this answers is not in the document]
paddycarver: Not really, Google didn't do this as much.

docs/design/import.md
```diff
@@ -0,0 +1,314 @@
...
+#### ResourceType Interface
+
+The framework can provide an extension interface type on the existing `ResourceType`:
+
+```go
+// ResourceTypeWithImportState represents a resource type with import support.
+type ResourceTypeWithImportState interface {
+    ResourceType
+
+    ImportState(context.Context, ImportResourceStateRequest, *ImportResourceStateResponse)
```
paddycarver, line 218: I think it's worth weighing the pros/cons of making import required.

paddycarver (changes requested):
I'm on board with the recommendation. Some thoughts:

* Should the ImportState method be required, with a built-in "not supported" helper? There are some pros/cons to explore there.
* I'd love if we could elaborate a little more on the state of the world with multiple resource imports.

Great job, @bflad.

docs/design/import.md
bflad: Will be included in next commit. 👍

docs/design/import.md
bflad: It didn't actually take too much convincing myself, but I actually really like the idea of requiring implementation on `Resource` assuming we offer an easy opt-out (e.g. return error diagnostic helper). I have added a new commit with this new proposal idea and recommending it. 🥲

docs/design/import.md
bflad: The latest commit now includes lots more information about multiple resource import including what Terraform CLI does implicitly with resource addresses and some examples of how support was implemented in the previous framework. 👍

docs/design/import.md
paddycarver: I think leaving it as it is is probably fine, I don't think it will matter soon (for some value of "soon", I have no idea when 1.1 is planned for).

paddycarver (approved):
Really solid doc @bflad. I'm on board, let's do it!

Issue #148, opened 2021-09-09
SetAttribute on Missing Attribute Path Silently Fails Without Update


bflad:
### Module version

```
v0.3.0
```

### Description

Given a schema like:

```go
Schema{
	Attributes: map[string]Attribute{
		"id": {
			Type:     types.StringType,
			Computed: true,
		},
	},
},
```

And example "empty" State assembly:

```go
resourceType, _ := s.getResourceType(ctx, req.TypeName)
resourceSchema, _ := resourceType.GetSchema(ctx)
state := State{
	Raw:    tftypes.NewValue(resourceSchema.TerraformType(ctx), nil),
	Schema: resourceSchema,
}
```

With the call:

```go
diags := state.SetAttribute(ctx, tftypes.NewAttributePath().WithAttributeName("id"), "example")
// len(diags) == 0
// before state == after state
```

### Expected Behavior

`(Plan).SetAttribute()` and `(State).SetAttribute()` should provide feedback about not updating the state properly or the behavior is documented in the Go documentation for those methods.

### Actual Behavior

Calling `(State).SetAttribute()` will silently fail, but the state will not be updated. This is because the transform function has no paths to match in this scenario and therefore `p.Equal(path)` will never match.

This was found while initially investigating how import support could work, but the same implications could apply to other places where new attribute paths are being introduced, such as `(DataSource).Read()` methods not calling `Set()` first or potentially adding elements to an existing attribute (although I have not verified either of these myself).

kmoe:
The original design for `SetAttribute` intended that the attribute should always be set _for valid attribute paths,_ whether or not the path already exists in the state value. 

There are some cases in which this is non-trivial. For example, (via @paddycarver), consider a case in which a `"list_of_things"` has four elements, then the following is run:
```go
SetAttribute(ctx, tftypes.NewAttributePath().WithAttributeName("list_of_things").WithElementKeyInt(12), "value") when list_of_things
```

In this case, the practitioner probably does not expect that elements 5-11 should be inserted into the list in order that element 12  be added. This call should therefore error.

There are likely other edge cases to discuss below.

This suggests to me that `SetAttribute` should do two things:
1. Validate the attribute path
2. Attempt to set the attribute

bflad:
My proposal (hopefully matching everyone else's thoughts) is the following behaviors for `SetAttribute()`:

- Errors if the attribute path or value is not valid with the current schema definition
- Path-specific rules that error otherwise:
[...]

PR #149, opened 2021-09-09
tfsdk: Initial ImportResourceState support

### LINKED ISSUE #33: Support Managed Resource Import

bflad:
Closes #33


tfsdk/serve_import_test.go
```diff
@@ -0,0 +1,181 @@
...
+
+			impl: func(ctx context.Context, req ImportResourceStateRequest, resp *ImportResourceStateResponse) {
+				diags := resp.State.SetAttribute(ctx, tftypes.NewAttributePath().WithAttributeName("id"), req.ID)
+				resp.Diagnostics.Append(diags...)
+			},
+
+			resp: &tfprotov6.ImportResourceStateResponse{
+				// TODO: SetAttribute should error or import should do the right thing here as
+				//       this seems like a potentially common implementation.
+				// Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/148
```
bflad, line 77: If this is not going to be resolved prior to v0.4.0, I suggest we at least update the ImportState method Go documentation to note that using resp.State.Set() is required for now.

tfsdk/request_import.go
```diff
@@ -0,0 +1,14 @@
...
+
+// ImportResourceStateRequest represents a request for the provider to import a
+// resource. An instance of this request struct is supplied as an argument to
+// the Resource's ImportState method.
+type ImportResourceStateRequest struct {
+	// ID represents the import identifier supplied by the practitioner when
+	// calling the import command. In many cases, this may align with the
+	// unique identifier for the resource, which can optionally be stored
+	// as an Attribute. However, this identifier can also be treated as
+	// treated as its own type of value and parsed during import. This value
```
paddycarver, line 11:
```suggestion
	// its own type of value and parsed during import. This value
```

tfsdk/request_import.go
```diff
@@ -0,0 +1,14 @@
...
+	// is not stored in the state unless implemented by the resource.
```
paddycarver, line 12:
```suggestion
	// is not stored in the state unless the provider explicitly stores it.
```

tfsdk/serve_import_test.go
paddycarver: I'd say we should prioritize it for 0.4.0.

tfsdk/serve_resource_config_validators_test.go
```diff
@@ -67,6 +67,8 @@ func (r testServeResourceConfigValidators) Update(ctx context.Context, req Updat
 }
 func (r testServeResourceConfigValidators) Delete(ctx context.Context, req DeleteResourceRequest, resp *DeleteResourceResponse) {
 }
+func (r testServeResourceConfigValidators) ImportState(ctx context.Context, req ImportResourceStateRequest, resp *ImportResourceStateResponse) {
+}
```
paddycarver, line 71: Why do all of the other test resource types get implementations that update `importResourceStateCalledResourceType`, but the validate ones don't?

paddycarver (commented):
Looks great to me, some small quibbles.

tfsdk/serve_resource_config_validators_test.go
bflad: Ah, this is because when I implemented those two test resources I was trying to omit parts that would be uncalled so future travelers just focused on the importantly implemented methods. It should have been commented as such.

Do you have a preference between these options?

1. Add a comment in these implementations, e.g. `// Intentionally blank. Not expected to be called during testing.`
2. Copy the implementations of other test resources (recognizing they _should_ all stay in sync going forward; not sure how to enforce that)
3. Removing _other_ unexpected test resource implementations, doing something like option 1 so it is more clear

Thanks!

paddycarver: I think my preference is 3, but replacing them with our not supported function and a comment saying that functionality is tested elsewhere. What do you think?

I mostly just am :/ about having code we don't _need_ in the test resources if we're not testing it, because that breeds confusion about whether it's used and where and what can be safely changed and what can't be in the future, I think. 🤔

bflad: Agreed! That will work for the new `ImportState` method implementations and I'll add the comment to the other empty Create/Read/Update/Delete test implementations. Will update shortly after merging main to ensure this is ready to rock. 🎸

bflad:
Everything should be ready for review again.

Something that gives me pause at the moment is how tooling will determine if a resource actually supports import and potentially how it can signal an example import identifier for documentation generation. I'm not sure if that should necessarily block an initial implementation, but it might be something worth iterating on in the future. Would love to hear thoughts and/or if I should create a followup issue.

bflad:
Created #160 for potential followup.

paddycarver (approved):
Yeah, let's go ahead and do it. 👍 🚀 Nice work.

kmoe (approved):
LGTM

tfsdk/serve_import.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	resource, diags := resourceType.NewResource(ctx, s.p)
+	resp.Diagnostics.Append(diags...)
+
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	emptyState := tftypes.NewValue(resourceSchema.TerraformType(ctx), nil)
```
paddycarver, line 85: A thing worth thinking about here:

We need a resolution to #148 to make this work the way we want.

Another option is to have it do something like:

```go
vals := map[string]tftypes.Value{}
typ := resourceSchema.TerraformType(ctx)
for name, t := range typ.(tftypes.Object).AttributeTypes {
  vals[name] = tftypes.NewValue(t, nil)
}
emptyState := tftypes.NewValue(typ, vals)
```

This matches the more-expected (imho) default state, which is an object in state with nothing filled in, instead of nothing in state.

However, this is weird during failure cases. I'm not clear on whether Terraform will persist the object to state or write nothing to state in the face of an error diagnostic. If the former, we should default to an empty state and just rely on #148. If the latter, we should confirm that will continue to be the case, and maybe defaulting to the empty object (instead of the null object) is more expected? I dunno. I don't know that it makes a huge difference, just wanted to call it out here.

bflad: The current logic in Terraform CLI is to return early on ImportResourceState error diagnostics, before saving state information:

https://github.com/hashicorp/terraform/blob/2afa0a5e75d76de7e47157114c579b3b1bff994f/internal/terraform/transform_import_state.go#L141-L154

paddycarver: 🤔 how do we feel about just leaving it and waiting for feedback?

bflad: We're going to merge this in as-is for now and iterate if necessary.

If we do wind up switching the implementation here in the future -- maybe creating the above code snippet as a method such as `(Schema).newState()` would be good to try and prevent "empty" object problems here and in any other potential places that might need it. That might be something we could export if/when multiple resource import support makes sense.

Issue #161, opened 2021-09-15
Add plan modification support to `attr.Type`


paddycarver:
### Module version

```
main@b0e3081ddbdab74c40ed0def9b8c23816a93d8fe
```

### Use-cases

Several types have semantically equivalent representations that are not byte-for-byte equivalent like Terraform requires. A JSON type comes to mind, in which

```json
{
  "key": "value"
}
```

and

```json
{"key": "value"}
```

should be considered equal, but Terraform treats as distinct values. By allowing plan modification strategies to be associated with `attr.Type`s, we can make sure this behavior is applied consistently everywhere the type is used.

### Attempted Solutions

We can work around this by adding an `AttributePlanModifier` to each `schema.Attribute`, but that requires consistency and extra work for provider developers that can be avoided.

### Proposal

Similar to [`attr.TypeWithValidate`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/attr#TypeWithValidate) we create an `attr.TypeWithModifyPlan` that allows defining plan modification for the type. This plan modification runs before attribute plan modification and before whole resource plan modification.

bflad:
(Old) example implementation for reference: https://github.com/hashicorp/terraform-plugin-framework/pull/162

bflad:
This is covered by #737 👍

PR #162, opened 2021-09-15
First stab at attr.TypeWithModifyPlan.

paddycarver:
Addresses #161.

### DIFF
--- a/attr/type.go
+++ b/attr/type.go
@@ -86,6 +86,17 @@ type TypeWithValidate interface {
 	Validate(context.Context, tftypes.Value, *tftypes.AttributePath) diag.Diagnostics
 }
 
+// TypeWithModifyPlan extends the Type interface to include a ModifyPlan method,
+// used to bundle consistent plan modification logic with the Type.
+type TypeWithModifyPlan interface {
+	Type
+
+	// ModifyPlan returns the Value that should be used in the plan. It is
+	// generally used to suppress diffs that do not correspond to semantic
+	// differences. In these cases, the `state` Value should be returned.
+	ModifyPlan(ctx context.Context, state, plan Value, path *tftypes.AttributePath) (Value, diag.Diagnostics)
+}
+
 // TypeWithPlaintextDescription extends the Type interface to include a
 // Description method, used to bundle extra information to include in attribute
 // descriptions with the Type. It expects the description to be written as
--- a/tfsdk/type_plan_modification.go
+++ b/tfsdk/type_plan_modification.go
@@ -1,0 +1,317 @@
+package tfsdk
+
+import (
+	"context"
+	"fmt"
+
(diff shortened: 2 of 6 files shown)


paddycarver:
I think this is ready for review, but is rebased on top of #231 and so depends on that being reviewed/merged first.

bflad:
I'll review once #231 is sorted and this change set is clearer. 👍

paddycarver:
Rebased after merging #231.

tfsdk/type_plan_modification.go
```diff
@@ -0,0 +1,386 @@
...
+		)
+		return plan, diags
+	}
+	return tftypes.NewValue(typ.TerraformType(ctx), planVals), diags
+}
+
+func attributeTypeModifyPlan(ctx context.Context, typ attr.Type, state, plan tftypes.Value, path *tftypes.AttributePath) (attr.Value, diag.Diagnostics) {
+	var diags diag.Diagnostics
+	if plan.IsKnown() && !plan.IsNull() {
+		if typ.TerraformType(ctx).Is(tftypes.Object{}) {
```
bflad, line 314: Nit: Seems this logic could be simplified via:

```go
tfType := typ.TerraformType(ctx)
switch {
case tfType.Is(tftypes.Object{}):
  plan, diags = attributeTypeModifyPlanObject(ctx, typ, state, plan, path)
case tfType.Is(tftypes.Map{}):
  plan, diags = attributeTypeModifyPlanMap(ctx, typ, state, plan, path)
// ...
}
if diags.HasError() {
  return plan, diags
}
```

Unless there's some explicit reason to return a `nil` value instead of the prior plan value during errors.

attr/type.go
```diff
@@ -86,6 +86,17 @@ type TypeWithValidate interface {
...
 
+// TypeWithModifyPlan extends the Type interface to include a ModifyPlan method,
+// used to bundle consistent plan modification logic with the Type.
+type TypeWithModifyPlan interface {
+	Type
+
+	// ModifyPlan returns the Value that should be used in the plan. It is
+	// generally used to suppress diffs that do not correspond to semantic
+	// differences. In these cases, the `state` Value should be returned.
+	ModifyPlan(ctx context.Context, state, plan Value, path *tftypes.AttributePath) (Value, diag.Diagnostics)
```
bflad, line 97: Open question: Should these also have access to the configuration value?

bflad (commented):
Some initial thoughts/considerations -- I apologize I'm out of time today, but can look again tomorrow. This also certainly feels like it'd be much easier if everything was already refactored for `attr.Value` throughout.

bflad:
@paddycarver I _think_ we're finally at a place where we can work on this after a lot of refactoring. Do you have interest in the implementation still?

bflad:
There have been a lot of changes to the framework affecting this particular code, so I'm going to close this.

PR #165, opened 2021-09-16
tfsdk: Allow Plan and State SetAttribute to create attribute paths

### LINKED ISSUE #148: SetAttribute on Missing Attribute Path Silently Fails Without Update

bflad:
Closes #148


bflad:
Hopefully this is much more understandable now. Please let me know if you want anything else adjusted.

tfsdk/plan.go
```diff
@@ -163,6 +172,105 @@ func (p *Plan) SetAttribute(ctx context.Context, path *tftypes.AttributePath, va
 	return diags
 }
 
+// pathExists walks the current state and returns true if the path can be reached.
+// The value at the path may be null or unknown.
+func (p Plan) pathExists(ctx context.Context, path *tftypes.AttributePath) (bool, diag.Diagnostics) {
```
paddycarver, line 177: Is my understanding correct that this is the same method on Plan and State?

tfsdk/schema.go
```diff
@@ -84,6 +84,10 @@ func (s Schema) AttributeTypeAtPath(path *tftypes.AttributePath) (attr.Type, err
 		return n.AttributeType(), nil
 	}
 
+	if s, ok := rawType.(Schema); ok {
```
paddycarver, line 87: If you're doing this because you got an error when walking the path, I think you're just going to want to return from the transform early if the attributepath is empty, instead. 🤔

tfsdk/plan.go
```diff
@@ -163,6 +172,105 @@ func (p *Plan) SetAttribute(ctx context.Context, path *tftypes.AttributePath, va
...
+	}
+
+	return len(remaining.Steps()) == 0, diags
+}
+
+// setAttributeTransformFunc recursively creates a value based on the current
+// Plan values along the path. If the value at the path does not yet exist,
+// this will perform recursion to add the child value to a parent value,
+// creating the parent value if necessary.
+func (p Plan) setAttributeTransformFunc(ctx context.Context, path *tftypes.AttributePath, tfVal tftypes.Value, diags diag.Diagnostics) (transformFunc, diag.Diagnostics) {
```
paddycarver, line 203: Same here, is this the same implementation between plan and state?

tfsdk/plan.go
bflad: Yes. Is there more to this question or a suggestion for how to improve it? The implementation felt small enough with the release time pressure that didn't feel warranted to refactor this further immediately, but can certainly spend cycles to update this further.

I could see that refactoring potentially going multiple ways since we don't own the `Plan.Raw`/`State.Raw` type (`tftypes.Value`), for example:

1. We could create an abstraction type in the framework and create a method on that. Then code becomes `p.Raw.pathExists(ctx, path)`. Seems related to https://github.com/hashicorp/terraform-plugin-framework/issues/172.
1. We could create a shared function that takes in that type as a parameter. Then code becomes `pathExists(p.Raw, ctx, path)`.
1. We could create a shared type that `Config`, `Plan`, and `State` all implement for their "data handling", which owns the method.

tfsdk/plan.go
bflad: See https://github.com/hashicorp/terraform-plugin-framework/pull/165#discussion_r714147857

tfsdk/schema.go
bflad: What would then be the expected behavior? I'm not sure how return early (no transform) helps us, but if I'm missing something please do explain!

If I'm understanding correctly, not returning the root schema attributes at "no" path would not work with this `ImportState` code:

https://github.com/hashicorp/terraform-plugin-framework/blob/f8cc3a2692a533fcbe8aee74e942f2cf11f0d35b/tfsdk/serve_import.go#L88-L99

And testing:

https://github.com/hashicorp/terraform-plugin-framework/blob/f8cc3a2692a533fcbe8aee74e942f2cf11f0d35b/tfsdk/serve_import_test.go#L63-L108

To ensure that actually works (e.g. one of the reasons we are making these changes), the code would then need to be updated here in `setAttributeTransformFunc()` from:

```go
parentPath := path.WithoutLastStep()
parentAttrType, err := s.Schema.AttributeTypeAtPath(parentPath)
```

To something like:

```go
parentPath := path.WithoutLastStep()

var err error
var parentAttrType attr.Type
if len(parentPath.Steps()) == 0 {
	parentAttrType = s.Schema.AttributeType
} else {
	parentAttrType, err = s.Schema.AttributeTypeAtPath(parentPath)
}
```

Although it seems that any future code that potentially hits the "no path" situation would also need to implement the same logic each time.

paddycarver: So I think what I'm trying to say is my mental model of the responsibilities of this code are:

1. `Schema.AttributeType` should be used to get the `attr.Type` of the entire schema.
2. `Schema.AttributeTypeAtPath` should be used to get the `attr.Type` of a specific attribute.

I don't know that I consider it a big deal in isolation, but I _am_ slightly concerned about the conflation of the two when used with the transform function.

Now that I've typed this all out I have realised I had confused myself. I was concerned that we're unnecessarily modifying the _entire resource's_ value when we only need to modify the value for a single attribute. Now that I am looking at it again, and started typing out that objection above, I realise we _do_ need to modify the entire resource's value, because transforms operate on the entire resource, not on a specific attribute of a resource. Because the specific attribute of a resource may have been created during the transform, and so the root needs to be modifiable to be updated to include it. I think. Now I'm unsure and think we _could_ get away with just skipping this in the transform, because it would replace the specific attribute.

This has been a journey, y'all. But I think I'm still on team "an error here probably makes sense, I don't think the transform function should _need_ the root schema's `attr.Type` because it shouldn't be modifying it at all." Unless the root schema is `null`.

....

... which is the exact situation we're trying to support. OK. I'm back on "I see why we're doing this", carry on.

I'd like off this rollercoaster please.

tfsdk/tftypes_value.go
```diff
@@ -0,0 +1,221 @@
...
+)
+
+// createParentValue ensures that the given parent value can have children
+// values upserted. If the parent value is known and not null, it is returned
+// without modification. A null Object or Tuple is converted to known with null
+// children. An unknown Object or Tuple is converted to known with unknown
+// children. List, Map, and Set are created with empty elements.
+//
+// The parentType parameter ensures that the value will match the current
+// schema, not what is currently stored in the parentValue.
```
paddycarver, line 18: I think we would actually like this to actively _not_ be the case; the schema represents type constraints (could be DynamicPseudoType, could be an object with optional attributes) but the value represents a concrete type (any non-DynamicPseudoType type, could be an object without optional attributes, respectively). So when modifying a value, I think we want to stick with the value's interpretation of what the type is supposed to be, not the schema's, so the value can stay internally consistent and match its type. Technically, it may even panic in situations where the value and type of the value start deviating.

tfsdk/tftypes_value.go
```diff
@@ -0,0 +1,221 @@
...
+}
+
+// upsertChildValue will upsert a child value into a parent value. If the
+// path step already has a value, it will be overwritten. Otherwise, the child
+// value will be added.
+//
+// Lists can only have the next element added according to the current length.
+//
+// The parentType parameter ensures that the value will match the current
+// schema, not what is currently stored in the parentValue.
```
paddycarver, line 75: Same thing here.

tfsdk/tftypes_value.go
```diff
@@ -0,0 +1,221 @@
...
+			vals = append(vals, tftypes.NewValue(elementType, childValue))
+		}
+
+		parentValue = tftypes.NewValue(parentType, vals)
+	default:
+		diags.AddAttributeError(
+			parentPath,
+			"Value Conversion Error",
+			"An unexpected error was encountered trying to create a value. This is always an error in the provider. Please report the following to the provider developer:\n\n"+
+				fmt.Sprintf("Unknown parent type %T to create value.", parentValue.Type()),
```
paddycarver, line 60:
```suggestion
				fmt.Sprintf("Unknown parent type %s to create value.", parentValue.Type()),
```

This will give us a friendlier output of what the type actually is semantically, not what the Go implementation of it is.

tfsdk/plan_test.go
```diff
@@ -389,32 +821,2518 @@ func TestPlanSetAttribute(t *testing.T) {
...
 					AttributeTypes: map[string]tftypes.Type{
-						"name": tftypes.String,
+						"other": tftypes.String,
+					},
+				}, map[string]tftypes.Value{
+					"other": tftypes.NewValue(tftypes.String, "should be untouched"),
+				}),
+				Schema: Schema{
+					Attributes: map[string]Attribute{
+						"test": {
```
paddycarver, line 835: This should appear in `Raw` as `nil`, I believe?

tfsdk/plan_test.go
```diff
@@ -389,32 +821,2518 @@ func TestPlanSetAttribute(t *testing.T) {
...
+				Raw: tftypes.NewValue(tftypes.Object{
+					AttributeTypes: map[string]tftypes.Type{
+						"other": tftypes.String,
+					},
+				}, map[string]tftypes.Value{
+					"other": tftypes.NewValue(tftypes.String, "should be untouched"),
+				}),
+				Schema: Schema{
+					Attributes: map[string]Attribute{
+						"tags": {
```
paddycarver, line 869: Same, these attributes would be in the Plan's `Raw`, set to `null`, not omitted from it.

paddycarver (changes requested):

paddycarver (approved):
LGTM, let's ship it, great work 👍 🚀

PR #166, opened 2021-09-17
types: Support Float64, Float64Type, Int64, and Int64Type

### LINKED ISSUE #137: Make a types.Int64 and types.Float64
### Module version

```
e620182e12a92ff9f1c3a8a9e2b2713860e80f89
```

### Use-cases

Right now, we only have `types.Number` which accepts integers _or_ floating point numbers and doesn't give the provider any way to limit that. But using numbers often means only accepting integers, and we should make that easy.

### Attempted Solutions

Using `types.Number` directly. This often means checking for and throwing errors if non-integer numbers are provided in the configuration.

### Proposal

Now that types can have validation, we can implement a `types.Int64` and `types.Float64` type, backed by `int64` and `float64` respectively. This will be easier to work with than `*big.Float` and makes the cognitive overhead lower for developers.

bflad:
Closes #137


types/float64.go
```diff
@@ -0,0 +1,86 @@
...
+
+	var bigF *big.Float
+	err := in.As(&bigF)
+
+	if err != nil {
+		return nil, err
+	}
+
+	// Validation is handled by Validate method.
+	f, _ := bigF.Float64()
```
paddycarver, line 28: I'd say let's still bubble up the error anyways, for when things don't work the way they're supposed to. Error checking in depth. :)

paddycarver (changes requested):

types/float64.go
bflad: You got it!

bflad: Ah, I'm not positive of this, but an interesting caveat to this is that this error I think will always be returned first instead of the `Validate()` diagnostics. I guess the only real difference is the error messaging/formatting, but could be something to consider.

paddycarver: Will it? I thought we designed `Validate` diagnostics to run _before_ trying to create any `attr.Value`s, which is (as far as I can tell) where this gets used? Otherwise, the `Validate` would take an `attr.Value` argument, I think?

paddycarver: https://github.com/hashicorp/terraform-plugin-framework/blob/84afb9afc9ce3e8e4777ca9024bd641ff646c363/tfsdk/config.go#L46-L73 This seems to suggest to me, for GetAttribute on Config specifically at least, that the `Validate` diags will be returned instead of the `ValueFromTerraform` error when both are returned. I think the reason I'm arguing we should handle it here anyways is as a safeguard in case we ever forget to call `Validate` or call them in the wrong order or something.

PR #167, opened 2021-09-17
tfsdk: Update Config, Plan, and State type GetAttribute methods to use reflection

### LINKED ISSUE #134: Make GetAttribute use reflection
### Module version

```
v0.2.0
```

### Use-cases

See #46.

### Attempted Solutions

#46

### Proposal

Rather than having `GetAttribute` return an `attr.Value`, let's mirror `Get` and have it take a `target interface{}` that it reflects into. Otherwise, people are just going to need to use `tfsdk.ValueOf` or type-assert out of it _anyways_, so there's no real point in surfacing the `attr.Value` version. If people want the `attr.Value`, they can make `target` an `attr.Value` implementation, so I don't know that we lose anything here.

This lets us mirror #133 for `reflect.Into` and mirrors how `SetAttribute` works, so it just feels more consistent.

### References

- #133 
- #46

bflad:
Closes #134

### DIFF
--- a/tfsdk/config.go
+++ b/tfsdk/config.go
@@ -21,10 +21,43 @@ func (c Config) Get(ctx context.Context, target interface{}) diag.Diagnostics {
 	return reflect.Into(ctx, c.Schema.AttributeType(), c.Raw, target, reflect.Options{})
 }
 
-// GetAttribute retrieves the attribute found at `path` and returns it as an
+// GetAttribute retrieves the attribute found at `path` and populates the
+// `target` with the value.
+func (c Config) GetAttribute(ctx context.Context, path *tftypes.AttributePath, target interface{}) diag.Diagnostics {
+	attrValue, diags := c.getAttributeValue(ctx, path)
+
+	if attrValue == nil {
+		return diags
+	}
+
+	valueAsDiags := ValueAs(ctx, attrValue, target)
+
+	// ValueAs does not have path information for its Diagnostics.
+	for idx, valueAsDiag := range valueAsDiags {
+		if valueAsDiag.Severity() == diag.SeverityError {
+			valueAsDiags[idx] = diag.NewAttributeErrorDiagnostic(
+				path,
+				valueAsDiag.Summary(),
+				valueAsDiag.Detail(),
+			)
+		} else if valueAsDiag.Severity() == diag.SeverityWarning {
+			valueAsDiags[idx] = diag.NewAttributeWarningDiagnostic(
+				path,
+				valueAsDiag.Summary(),
+				valueAsDiag.Detail(),
+			)
+		}
+	}
+
+	diags.Append(valueAsDiags...)
+
+	return diags
+}
+
+// getAttribute retrieves the attribute found at `path` and returns it as an
 // attr.Value. Consumers should assert the type of the returned value with the
 // desired attr.Type.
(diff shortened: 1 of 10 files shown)


tfsdk/config.go
```diff
@@ -21,10 +21,43 @@ func (c Config) Get(ctx context.Context, target interface{}) diag.Diagnostics {
 	return reflect.Into(ctx, c.Schema.AttributeType(), c.Raw, target, reflect.Options{})
 }
 
-// GetAttribute retrieves the attribute found at `path` and returns it as an
+// GetAttribute retrieves the attribute found at `path` and populates the
+// `target` with the value.
+func (c Config) GetAttribute(ctx context.Context, path *tftypes.AttributePath, target interface{}) diag.Diagnostics {
+	attrValue, diags := c.getAttributeValue(ctx, path)
+
+	if attrValue == nil {
```
paddycarver, line 29: Would we not want `diags.HasError()` here instead?

tfsdk/config.go
```diff
@@ -21,10 +21,43 @@ func (c Config) Get(ctx context.Context, target interface{}) diag.Diagnostics {
...
+				valueAsDiag.Summary(),
+				valueAsDiag.Detail(),
+			)
+		} else if valueAsDiag.Severity() == diag.SeverityWarning {
+			valueAsDiags[idx] = diag.NewAttributeWarningDiagnostic(
+				path,
+				valueAsDiag.Summary(),
+				valueAsDiag.Detail(),
+			)
+		}
```
paddycarver, line 49: Eek. Can we add an issue at least to have `diag.Diagnostic` have a `SetPath` method to avoid this kind of stuff?

tfsdk/config.go
bflad: You're right we should check for error diagnostics, although for safety we'll probably want both checks. I think the reason I chose `attrValue == nil` is because it will otherwise panic in `ValueAs()`... or something... but I forgot to add the error check as well.

tfsdk/config.go
bflad: I created https://github.com/hashicorp/terraform-plugin-framework/issues/169, but it might require some refactoring to the `diag` package.

tfsdk/config.go
```diff
@@ -21,10 +21,45 @@ func (c Config) Get(ctx context.Context, target interface{}) diag.Diagnostics {
 	return reflect.Into(ctx, c.Schema.AttributeType(), c.Raw, target, reflect.Options{})
 }
 
-// GetAttribute retrieves the attribute found at `path` and returns it as an
+// GetAttribute retrieves the attribute found at `path` and populates the
+// `target` with the value.
+func (c Config) GetAttribute(ctx context.Context, path *tftypes.AttributePath, target interface{}) diag.Diagnostics {
+	attrValue, diags := c.getAttributeValue(ctx, path)
+
+	if attrValue == nil || diags.HasError() {
```
paddycarver, line 29: I'm still 🤔 about the case where `attrValue` is `nil`. If error diagnostics are returned, the provider will think that `target` has been set, when it has not. If error diagnostics _are_ returned, then the check is unnecessary. The only case in which the check helps us avoids a panic but is really hard to debug.

Do we want to catch a `nil` `attrValue` and return an error diagnostic for it, instead?

bflad:
This may require some light rebase work depending on #185

bflad:
Rebased after #185

PR #179, opened 2021-09-24
go get github.com/hashicorp/terraform-plugin-go@v0.4.0

bflad:
And of course: go mod tidy


bflad:
This was approved out of band yesterday, but we somehow missed clicking buttons in here. Merging this now so its included with the next release.

PR #185, opened 2021-09-29
tfsdk: Return null value when missing from Config, Plan, and State GetAttribute() method

### LINKED ISSUE #150: "step cannot be applied to this value" failure after before update if nested attribute was previously null
When running `terraform plan` on a file with a previously created resource, any optional properties (and their child properties, even if their parent was null) that were null in the state fail with `AttributeName("...") still remains in the path: step cannot be applied to this value`.

The initial create happens without any problems. When I run `terraform plan` again, the refresh also seems to succeed. But before the update handler is run, it throws out the "State Read Error"s. It seems to be related to #112, but happening at a different point in the lifecycle.

I managed to recreate this in the hashicups demo (after making all the diagnostics changes to match v0.3.0). The error looks like this:

```
╷
│ Error: State Read Error
│ 
│   with hashicups_order.edu,
│   on main.tf line 15, in resource "hashicups_order" "edu":
│   15: resource "hashicups_order" "edu" {
│ 
│ An unexpected error was encountered trying to read an attribute from the state. This is always an error in the provider. Please report the following to the
│ provider developer:
│ 
│ AttributeName("inner") still remains in the path: step cannot be applied to this value
╵
```

### Module version

```
github.com/hashicorp/terraform-plugin-framework v0.3.0
```

```go
func (r resourceType) GetSchema(_ context.Context) (tfsdk.Schema, diag.Diagnostics) {
	return tfsdk.Schema{
			"something": {
				Optional: true,
					"inner": {
						Type:     types.StringType,
						Optional: true,
					},
				}),
			},
			...
```

```go
// Create a new resource
func (r resourceOrder) Create(ctx context.Context, req tfsdk.CreateResourceRequest, resp *tfsdk.CreateResourceResponse) {
	...
	// Generate resource state struct
	var result = Order{
		ID:          types.String{Value: strconv.Itoa(order.ID)},
		Items:       ois,
		Something:   types.Object{Null: true}, // Creating "Something" with an initial null value
		LastUpdated: types.String{Value: string(time.Now().Format(time.RFC850))},
	}
	...
```

```hcl
// Initial create
resource "hashicups_order" "edu" {
  # something = {}
  items = [...]
}

// And then change it to this and run "terraform plan"
resource "hashicups_order" "edu" {
  something = {}
  items = [...]
}
```

### Debug Output

https://gist.github.com/avinashbot/def1208ed818aa1af246f190e28ebe63

### Expected Behavior

To not error out on optional resources that have null in the state.

### Actual Behavior

The printed error message.

### Steps to Reproduce

1. `terraform apply` to create resources
2. `terraform plan` to refresh/update

### References

- #112 has the same error message.

bflad:
Closes #150

Includes more exhaustive unit testing `Config`, `Plan`, and `State` `GetAttribute()` method, which is copied between all of them. Also left TODOs for #186 to follow this up with unknown parent checking.

### DIFF
--- a/.changelog/185.txt
+++ b/.changelog/185.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+tfsdk: Fetch null values from valid missing `Config`, `Plan`, and `State` paths in `GetAttribute()` method
+```
--- a/tfsdk/attribute_test.go
+++ b/tfsdk/attribute_test.go
@@ -708,15 +708,15 @@ func TestAttributeModifyPlan(t *testing.T) {
 				Config: Config{
 					Raw: tftypes.NewValue(tftypes.Object{
 						AttributeTypes: map[string]tftypes.Type{
-							"nottest": tftypes.String,
+							"test": tftypes.String,
 						},
 					}, map[string]tftypes.Value{
-						"nottest": tftypes.NewValue(tftypes.String, "testvalue"),
+						"test": tftypes.NewValue(tftypes.String, "testvalue"),
 					}),
 					Schema: Schema{
 						Attributes: map[string]Attribute{
 							"test": {
-								Type:     types.StringType,
+								Type:     types.ListType{ElemType: types.StringType},
 								Required: true,
 							},
 						},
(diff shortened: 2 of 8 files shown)


kmoe:
:+1: agree with extending this approach to Plan and State

kmoe (approved):
LGTM

PR #188, opened 2021-09-29
tfsdk: Support for List and Set Blocks

### LINKED ISSUE #85: Consider adding (deprecated) support for blocks 😢
### Module version

```
v0.2.0
```

### Use-cases

We decided to model nested fields as nested attributes, because they're better in basically every way™, and I think that was a good decision.

However, for providers that currently exist and want to adopt the framework using mux, this presents two problems:

1. Provider blocks need to match perfectly when muxing. But SDKv2 has no way to model nested attributes, and the framework has no way to model blocks, meaning if you have nested fields of any kind in your provider configuration, you just can't mux across SDKv2 and the framework. This is suboptimal, and makes migration to the framework much, much more difficult.
2. Changing a group of nested fields from a block to an attribute is a _breaking change_, because it requires practitioners to add the `=` to their configuration. Meaning piecemeal migration of resources is not possible, as you can't model blocks in the framework, and need to migrate to nested attributes at that point, which is a practitioner-visible change.

Both of these problems severely limit how palatable our migration story can be.

### Attempted Solutions

I thought about just relaxing mux to consider blocks and nested attributes the same for considering whether provider blocks are "equivalent" or not, as they'll both show up as objects in the config and can be accessed the same way. But that seems brittle, and would lead to weird situations where things aren't treated as sensitive properly, or where things can be expressed in the framework that SDKv2 doesn't know how to handle... it just feels like a really leaky abstraction and lossy mapping. It also doesn't address point 2 basically at all.

### Proposal

I'm proposing that we add support to the framework for modeling blocks, and make it clear from its introduction that it is deprecated, we recommend everyone use nested attributes, and that it's only supported to ease migration from SDKv2. This isn't ideal, but I don't see a way to get a clean, piecemeal, non-practitioner-impacting migration strategy if we don't.

bflad:
Closes #85

The primary purpose for supporting blocks is to allow previously existing schemas defined by the older Terraform Plugin SDK to not require practitioner breaking changes upon migrating to the framework (except the protocol version and therefore the minimum Terraform CLI version unless a tfprotov5 server implementation is added to this framework as well). This also allows this framework to be muxed with the older framework since the provider schema must match. It is expected that over time any schema definitions including `Blocks` will migrate to `Attributes`.

Provider developers should always opt for `Attributes` in new schema definitions.


tfsdk/nested_attributes.go
```diff
@@ -8,9 +8,14 @@ import (
...
-// They can be a list, a set, or a map (with string keys), or they can be
-// nested directly, like an object.
+// NestingMode is an enum type of the ways nested attributes can be nested in
+// an attribute or a block. They can be a list, a set, a map (with string
+// keys), or they can be nested directly, like an object.
+//
+// While the protocol and theoretically Terraform itself support map, single,
+// and group nesting modes, this framework intentionally only supports list
+// and set for blocks as those other modes were not typically implemented or
+// tested since the older Terraform Plugin SDK did not support them.
```
bflad, line 18: I think we might prefer a separate type (e.g. `BlockNestedMode`) so this is handled at compile-time instead of overloading the existing enum, but I left this as-is for now.

tfsdk/attribute.go
```diff
@@ -275,142 +300,117 @@ func (a Attribute) validate(ctx context.Context, req ValidateAttributeRequest, r
...
-			}
+	nm := a.Attributes.GetNestingMode()
+	switch nm {
+	case NestingModeList:
+		l, ok := req.AttributeConfig.(types.List)
 
-			for _, value := range s.Elems {
-				tfValueRaw, err := value.ToTerraformValue(ctx)
+		if !ok {
+			err := fmt.Errorf("unknown attribute value type (%T) for nesting mode (%T) at path: %s", req.AttributeConfig, nm, req.AttributePath)
```
paddycarver, line 340: I know you're not the one creating this message, but because we're here anyways, do we want to use the `attr.Type.String` method here instead of `%T`, to honor the provider developer's choices about how their type is represented?

tfsdk/block.go
```diff
@@ -0,0 +1,412 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-go/tfprotov6"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+)
+
+// Block defines the constraints and behaviors of a single structural field in a
+// schema.
+type Block struct {
+	tftypes.AttributePathStepper
```
paddycarver, line 18: Does this work? What does it do, to embed an interface into a struct? 🤔

tfsdk/block.go
```diff
@@ -0,0 +1,412 @@
...
+		}
+
+		return schemaNestedBlock.Block.BlockTypes[i].TypeName < schemaNestedBlock.Block.BlockTypes[j].TypeName
+	})
+
+	return schemaNestedBlock, nil
+}
+
+// validate performs all Block validation.
+func (b Block) validate(ctx context.Context, req ValidateAttributeRequest, resp *ValidateAttributeResponse) {
```
paddycarver, line 231: If I'm reading this right, we're opting _not_ to have block-level validation available. Is that the case? Is that by design?

tfsdk/nested_attributes.go
paddycarver: I think I'd agree with that, the cost/benefit tradeoff seems reasonable to me.

paddycarver (commented):
Start of a review, getting distracted so going to leave the comments I have, but still have to look through a lot of this, sorry.

tfsdk/block.go
bflad: We can certainly add it, it's just not included yet. I don't think there's anything preventing its inclusion other than time/effort. If this is looking okay so far, I'll spend that time adding it.

(Aside: I also need to at least passthrough attribute plan modification through the block support still.)

tfsdk/block.go
bflad: Good catch. I thought it added compile-time verification that the interface was actually satisfied by the implementation, but It certainly does not. I'll replace it with the `var` approach. 👍

tfsdk/attribute.go
bflad: That sounds reasonable to me!

tfsdk/block.go
bflad: Addressed 👍

bflad:
I've addressed the first round of feedback. I've created #222 for plan modification so we can either try to include that in this first round of support or later on since that's another large code chunk.

tfsdk/block.go
```diff
@@ -0,0 +1,406 @@
...
+	resp.Diagnostics.Append(diags...)
+
+	if diags.HasError() {
+		return
+	}
+
+	req.AttributeConfig = attributeConfig
+
+	for _, validator := range b.Validators {
+		validator.Validate(ctx, req, resp)
```
paddycarver, line 242: I forget offhand, but do we need to return early here if `resp.Diagnostics.HasError`?

tfsdk/schema.go
```diff
@@ -15,21 +15,49 @@ import (
...
+	// set the value. Blocks are instead repeated as necessary, or require
+	// the use of dynamic block expressions. See also:
+	//   https://www.terraform.io/docs/language/syntax/configuration.html
+	//   https://www.terraform.io/docs/language/expressions/dynamic-blocks.html
+	//
+	// Attributes are strongly preferred over Blocks. Blocks should only be
+	// used for configuration compatibility with previously existing schemas
+	// from an older Terraform Plugin SDK. Efforts should be made to convert
+	// Blocks to Attributes as a breaking change for practitioners.
+	Blocks map[string]Block
```
paddycarver, line 59: Because we know we want to remove this eventually, do we want to put a `// Deprecated:` comment on it so that linters pick up usage of it and people can have automated guidance in moving away?

tfsdk/schema.go
```diff
@@ -75,40 +111,32 @@ func (s Schema) AttributeTypeAtPath(path *tftypes.AttributePath) (attr.Type, err
...
+		return typ.AttributeType(), nil
+	case nestedBlock:
+		return typ.Block.attributeType(), nil
+	case Attribute:
+		return typ.attributeType(), nil
+	case Block:
+		return typ.attributeType(), nil
+	case Schema:
+		return typ.AttributeType(), nil
+	default:
```
paddycarver, line 127: This whole block makes me think we're missing a chance here for an interface. I don't think that refactoring needs to happen as part of this PR. Just calling it out as something to think on, and at most maybe add a TODO about.

paddycarver (commented):
Some final minor comments, but on the whole looks good to me.

A lot going on here, and it's really bringing home for me how badly #215 is needed. But I think I'm as confident as reasonably possible in the changes.

I think there's a lingering question in my mind of whether we want blocks to match nested attributes in using enum functions (e.g. [`tfsdk.ListNestedAttributes`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/tfsdk#ListNestedAttributes) but `ListNestedBlocks` instead) just for uniformity, but I'm 🤔 about that design and whether we want to keep it, anyways, so I think that's always something we can tackle later, if we need to.

Awesome work.

tfsdk/block.go
bflad: No worries! The current implementation of `(Attribute).validate()` loops through `Validators` without skipping early:

https://github.com/hashicorp/terraform-plugin-framework/blob/16188d2b2de27dc006107f5d637aa9d7e44b7a37/tfsdk/attribute.go#L274-L276

The rationale that I can recall is that since the configuration value is not changing during this process, that all validators should return their errors as they likely will represent different expectations, e.g.

```go
Validators: []AttributeValidator{
  stringLengthAtMost(8),
  stringBeginsWith("tf"),
},
```

So if `some-long-string-without-tf-prefix!` was passed in, both validation error diagnostics would be returned. 👍

tfsdk/schema.go
bflad: Ah yes, thank you, will add that.

tfsdk/schema.go
bflad: Created https://github.com/hashicorp/terraform-plugin-framework/issues/223 👍

tfsdk/block.go
paddycarver: That seems reasonable to me!

tfsdk/schema.go
paddycarver: Thanks!

paddycarver (approved):
Let's do it 😎

Issue #172, opened 2021-09-21
Reexamine where we convert from Terraform values into `attr.Value`s.


paddycarver:
Currently, the framework is following a model of keeping everything as a `tftypes.Value` all the way through the framework, until the config, state, or plan tries to retrieve it, then converts back from an `attr.Value` into a `tftypes.Value` when setting the config, state, or plan. That means that for all the framework code, we're more likely to be using the `tftypes.Value` abstraction than the `attr.Value` abstraction provider developers are using and expect.

I'm suspicious that we can clean up a lot of code, handle a lot of edge cases, and make more things work intuitively the way provider developers expect if we instead convert from `tftypes.Value`s into `attr.Value`s when handling protocol requests, and then passing around the `attr.Value`s.

bflad:
I think this pairs well with https://github.com/hashicorp/terraform-plugin-framework/issues/215 👍

bflad:
Created bespoke issues for tracking specific pieces of framework updates:

- #588 
- #589 
- #590

If there are other places which need updates, we can create other new issues.

PR #194, opened 2021-10-01
attr: Added `IsNull()` and `IsUnknown()` methods to the `Value` interface type

### LINKED ISSUE #193: attr: Consider Adding IsNull() and IsUnknown() methods to Value
### Module version

```
v0.4.2
```

### Use-cases

When writing attribute plan modifiers, it can be necessary to check if the configuration or state `attr.Value` are null or unknown. It feels a little tedious to have to know the underlying value type for comparison and how to properly create the conditional.

### Attempted Solutions

```go
"function_name": {
			// Property: FunctionName
			// CloudFormation resource type schema:
			// {
			//   "description": "The name of the Lambda function, up to 64 characters in length. If you don't specify a name, AWS CloudFormation generates one.",
			//   "minLength": 1,
			//   "type": "string"
			// }
			Description: "The name of the Lambda function, up to 64 characters in length. If you don't specify a name, AWS CloudFormation generates one.",
			Optional:    true,
			Computed:    true,
				validate.StringLenAtLeast(1),
			},
				tfsdk.RequiresReplaceIf(func(ctx context.Context, state, config attr.Value, path *tftypes.AttributePath) (bool, diag.Diagnostics) {
					if config.Equal(state) || config.Equal(types.String{Null: true}) {
						return false, nil
					}

					return true, nil
				}, "require replacement if configuration value changes", "require replacement if configuration value changes"),
			},
		},
```

### Proposal

```go
type Value interface {
	// ... existing methods ...

	// IsNull returns true if the value is null.
	IsNull() bool

	// IsUnknown returns true if the value is unknown.
	IsUnknown() bool
}
```

For any `attr.Value` that potentially do not understand null/unknown, they can always return false for both methods. The implementation above then becomes:

```go
				tfsdk.RequiresReplaceIf(func(ctx context.Context, state, config attr.Value, path *tftypes.AttributePath) (bool, diag.Diagnostics) {
					if config.Equal(state) || config.IsNull() {
						return false, nil
					}

					return true, nil
				}, "require replacement if configuration value changes", "require replacement if configuration value changes"),
			},
```

bflad:
Closes #193


kmoe (approved):
Needs review by @paddycarver but I'm on board with this as you [know](https://github.com/hashicorp/terraform-plugin-framework/blob/main/docs/design/plan-modification.md#requiresreplaceifset)!

paddycarver:
Arguments for methods:

* Discoverability: because standalone functions aren't tied to the `attr.Value` type, they're not surfaced by the language server, godoc, etc. as things you can do.

paddycarver:
We decided we wanted to wait on this momentarily until we can tackle #215, #172, and other refactoring issues, and refactor our package boundaries a little. Right now, it's a little hard to understand how large the discoverability tradeoff is going to be, just because we're guessing at what package architectures may be available in the future. I'm still hopeful we'll be able to have an `attr` package that wraps all this up neatly, but we can't today because we get import cycles. We'll return to this once we figure out what package boundaries we can draw after refactoring a bit.

bflad:
I'm going to close this out for now given some other recent/upcoming changes. Can revisit this again in the future, if it makes sense.

PR #195, opened 2021-10-01
docs/design: Initial schema blocks support documentation

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/85
See also: https://github.com/hashicorp/terraform-plugin-framework/pull/188

### DIFF
--- a/docs/design/blocks.md
+++ b/docs/design/blocks.md
@@ -1,0 +1,645 @@
+# Schema Blocks
+
+The framework currently implements nested attribute support in schemas that is only available in the recent Terraform Plugin Protocol version 6. Prior versions of the protocol and SDK implemented support for blocks. Provider developers may have previously existing resources from the older Terraform Plugin SDK that they wish to mux with or bring into the new framework without introducing breaking changes for practitioners.
+
+This design documentation will walk through and recommend options for schema block handling in the framework.
+
+## Background
+
+### Terraform Configuration Language
+
+Terraform's configuration language is built on top of [HashiCorp Configuration Language (HCL)](https://github.com/hashicorp/hcl) and conceptually supports two syntaxes for declaring configuration as described in the [Configuration Syntax documentation](https://www.terraform.io/docs/language/syntax/configuration.html):
+
+- **Attributes** (**Arguments**): Assign a value to a particular name.
+- **Blocks**: Container for other content.
+
+With the attributes using an equals sign (`=`), such as:
+
+```terraform
+resource "example_thing" "example" {
+  some_attr = "abc123" # Example attribute/argument
+}
+```
+
+While blocks are structural and defined without an equals sign (`=`), such as:
+
+```terraform
+resource "example_thing" "example" {
+  some_block {
+    # ... attributes/arguments ...
+  }
+}
+```
+
+Blocks can be repeated multiple times if the underlying schema type allows multiple declarations, such as:
+
+```terraform
+resource "example_thing" "example" {
+  some_block {
+    # ... attributes/arguments ...
+  }
+
+  some_block {
+    # ... attributes/arguments ...
+  }
+}
+```
+
+To handle situations where the block configuration is not static, the configuration language added support for [`dynamic` block expressions](https://www.terraform.io/docs/language/expressions/dynamic-blocks.html). In contrast, working with attributes is relatively simpler as there are many expressions that can operate directly on any collection value, such as [`for` expressions](https://www.terraform.io/docs/language/expressions/for.html).
+
+Blocks also have other limitations in Terraform, such as not supporting sensitivity. Instead, each nested attribute must implement sensitivity as appropriate.
+
+### Terraform Plugin Protocol
+
+In terms of the protocol, all configuration information from the plugin is referred to as the schema. Prior to the recent version 6 of the protocol, only blocks were supported implementing complex schema types.
+
+Version 6 of the protocol defines the following implementation details for schema:
+
+```protobuf
+message Schema {
+    message Block {
+        int64 version = 1;
+        repeated Attribute attributes = 2;
+        repeated NestedBlock block_types = 3;
+        string description = 4;
+        StringKind description_kind = 5;
+        bool deprecated = 6;
+    }
+
+    message Attribute {
+        string name = 1;
+        bytes type = 2;
+        Object nested_type = 10;
+        string description = 3;
+        bool required = 4;
+        bool optional = 5;
+        bool computed = 6;
+        bool sensitive = 7;
+        StringKind description_kind = 8;
+        bool deprecated = 9;
+    }
+
+    message NestedBlock {
+        enum NestingMode {
+            INVALID = 0;
+            SINGLE = 1;
+            LIST = 2;
+            SET = 3;
+            MAP = 4;
+            GROUP = 5;
+        }
+
+        string type_name = 1;
+        Block block = 2;
+        NestingMode nesting = 3;
+        int64 min_items = 4;
+        int64 max_items = 5;
+    }
+
+    message Object {
+        enum NestingMode {
+            INVALID = 0;
+            SINGLE = 1;
+            LIST = 2;
+            SET = 3;
+            MAP = 4;
+        }
+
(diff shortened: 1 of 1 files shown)


docs/design/blocks.md
```diff
@@ -0,0 +1,645 @@
...
+// worry only about CRUD operations while off-loading validation, diff
+// generation, etc. to this higher level library.
+//
+// In spite of the name, this struct is not used only for terraform resources,
+// but also for data sources. In the case of data sources, the Create,
+// Update and Delete functions must not be provided.
+type Resource struct {
+    // ... other fields ...
+
+    // Schema is the schema for the configuration of this resource.
```
metadave, line 302: the whitespace seems off here

docs/design/blocks.md
```diff
@@ -0,0 +1,645 @@
...
+}
+```
+
+To handle situations where the block configuration is not static, the configuration language added support for [`dynamic` block expressions](https://www.terraform.io/docs/language/expressions/dynamic-blocks.html). In contrast, working with attributes is relatively simpler as there are many expressions that can operate directly on any collection value, such as [`for` expressions](https://www.terraform.io/docs/language/expressions/for.html).
+
+Blocks also have other limitations in Terraform, such as not supporting sensitivity. Instead, each nested attribute must implement sensitivity as appropriate.
+
+### Terraform Plugin Protocol
+
+In terms of the protocol, all configuration information from the plugin is referred to as the schema. Prior to the recent version 6 of the protocol, only blocks were supported implementing complex schema types.
```
kmoe, line 54:
```suggestion
In terms of the protocol, all configuration information from the plugin is referred to as the schema. Prior to the recent version 6 of the protocol, implementing complex schema types could only be done using blocks.
```

Perhaps also add a brief sentence pointing out the addition of `nested_type` to `Attribute` and what exactly this enables.

bflad: Thanks! Pulled in that change and added a quick sentence there about `nested_type`.

docs/design/blocks.md
```diff
@@ -0,0 +1,645 @@
...
+	Blocks: ListNestedBlocks(map[string]Attribute{
+		"example_attribute": {
+			Type:     types.StringType,
+			Required: true,
+		},
+	}, ListNestedBlocksOptions{}),
+},
+```
+
+This allows the framework to differentiate between block and nested attributes by fields.
```
paddycarver, line 598: My question here is that I think this could allow providers to define incorrect schemas that we can't model with the protocol. For example, a ListNestedAttributes could contain a `tfsdk.Attribute` that has a `Blocks` property on it, which, if memory serves, we can't actually model in the protocol; attributes can only have attributes under them.

I think allowing this in the type system when it's not supported in the protocol may be confusing.

paddycarver (commented):
I think I disagree with the recommendation on this one, explained in the comment I left. I think in my head, I envisioned the "blocks as a separate type" approach, because it more closely aligns with the abstractions we're surfacing.

It may be worth noting that our ultimate goal is removing support for blocks again.

I'm interested in the assertion that the recommended approach will allow for easier conversion from blocks to attributes in the future. Is it just because it can be done via find-and-replace instead of moving a field between two maps? Can we mitigate that advantage through automation?

bflad:
> I'm interested in the assertion that the recommended approach will allow for easier conversion from blocks to attributes in the future. Is it just because it can be done via find-and-replace instead of moving a field between two maps? Can we mitigate that advantage through automation?

Correct. Once you move beyond very simple code editing operations such as find-replace, at least with tooling that I've worked with like `go/analysis`, `semgrep` and `rf`, it becomes a bit of a fools errand to enumerate all the possible cases to introduce automation that catches it all. If you are familiar with other tooling, frameworks, or concepts that make this easier, would certainly like to hear more.

For example, its possible to declare things like this:

```go
func (t exampleResourceType) GetSchema(ctx context.Context) (tfsdk.Schema, diag.Diagnostics) {
	return tfsdk.Schema{
		Blocks: map[string]tfsdk.Block{
			"example_block": {
					"example_attr": {
						Type:     types.StringType,
						Required: true,
					},
				},
				Nesting: tfsdk.BlockNestingList,
			},
		},
	}, nil
}
```

Or:

```go
var exampleBlock := tfsdk.Block{
		"example_attr": {
			Type:     types.StringType,
			Required: true,
		},
	},
	Nesting: tfsdk.BlockNestingList,
}

func (t exampleResourceType) GetSchema(ctx context.Context) (tfsdk.Schema, diag.Diagnostics) {
	return tfsdk.Schema{
		Blocks: map[string]tfsdk.Block{
			"example_block": exampleBlock,
		},
	}, nil
}
```

Or:

```go
func exampleBlock() tfsdk.Block {
	return tfsdk.Block{
			"example_attr": {
				Type:     types.StringType,
				Required: true,
			},
		},
		Nesting: tfsdk.BlockNestingList,
	}
}

func (t exampleResourceType) GetSchema(ctx context.Context) (tfsdk.Schema, diag.Diagnostics) {
	return tfsdk.Schema{
		Blocks: map[string]tfsdk.Block{
			"example_block": exampleBlock(),
		},
	}, nil
}
```

Plus others... Now the automation needs to worry about converting multiple types and fields that might be spread across the AST. `go/analysis` can pick up some of these cases with effort, but rewriting code in that framework generally needs to be able to compile between passes (in my experience).

There's also code like this that even further complicates things:

```go
var listBlockMode := tfsdk.BlockNestingList
var exampleBlock := tfsdk.Block{
		"example_attr": {
			Type:     types.StringType,
			Required: true,
		},
	},
	NestingMode: listBlockMode,
}

func (t exampleResourceType) GetSchema(ctx context.Context) (tfsdk.Schema, diag.Diagnostics) {
	return tfsdk.Schema{
		Blocks: map[string]tfsdk.Block{
			"example_block": exampleBlock,
		},
	}, nil
}
```

The tooling then needs to know to introspect the value of the variable to extract its value. Worse would be a function call, which could get into SSA-land. Experimental tooling like `rf` might be able to do some things that are harder in `go/analysis`, but then you will likely need to mix and match tools, which is its own complexity.

To summarize the differences in terms of potential operations:

Blocks as New Attribute Field:

- Rename `Blocks` field identifier to `Attributes`
- Rename `ListNestedBlocks`/`SetNestedBlocks` call expressions to `ListNestedAttributes`/`SetNestedAttributes`
- Rename `ListNestedBlocksOptions`/`SetNestedBlocksOptions` declarations to `ListNestedAttributesOptions`/`SetNestedBlocksOptions`
- Either default to adding `Required`/`Optional`/`Computed` field identifier(s) and `true` value(s) or its a manual step for provider developers

Blocks as a Separate Type:

- Convert `map[string]Block` declaration to `map[stringAttribute`
- Convert `Block` declaration to `Attribute`
    - Wrap `Attributes` with call expression (e.g. `ListNestedAttributes`) based on `NestingMode`
        - Migrate `MinItems` and `MaxItems` to options fields (e.g. `ListNestedAttributesOptions`) or `Validators`
    - Remove `NestingMode` field identifier and value
- Rename `Blocks` field identifier to `Attributes`
    - If `Attributes` exists already, append the block's attributes to existing `Attributes` map (yikes)
- Either default to adding `Required`/`Optional`/`Computed` field identifier(s) and `true` value(s) or its a manual step for provider developers

All these differences are also reflected in the complexity required to work with both concepts in a provider, meaning multiple mental models are necessary to work with code using both. Whether that's a good or bad thing, I think is debatable, since they are conceptually different in many regards. However, this is all predicated on my presumption that since block support is deprecated, we do not need to worry about drastic changes to the protocol regarding them, which would in turn make the abstraction difficult to maintain over time. If that presumption is wrong, would love to know that.

---

> My question here is that I think this could allow providers to define incorrect schemas that we can't model with the protocol. ... I think allowing this in the type system when it's not supported in the protocol may be confusing.

This can be solved with an additional schema validation, but it is annoying that it wouldn't be caught at compile time.

---

All this said, the current recommendation is not a hill I'm looking to defend strongly. I do, however, worry much about the necessary complexity for provider developers working with both `Attributes` and `Blocks` and any future migration strategies we want to employ.

PR #197, opened 2021-10-04
Update info in README about supported features

paddycarver:
We have a lot more supported features now! Also, let's point people to the page helping them decide which SDK to use.


bflad (approved):
Sorry, forgot about this note here in addition to the website.

Issue #180, opened 2021-09-24
Consider Adding Attribute State Copying Plan Modifiers


bflad:
### Module version

```
v0.4.0
```

### Use-cases

Due to the proposed behavior changes of #175, it is likely that an upcoming version of the framework will begin marking all `Computed`-only attributes as unknown in plans. The side effect is that provider developers may explicitly know that the attribute's state value should be copied to the plan as it is expected to not change during update (and therefore be hidden from plan output).

### Attempted Solutions

Provider developers are able to manually create plan modifiers which copy the prior state value into the plan.

### Proposal

This is likely to be a common use case for providers to improve plan outputs. The framework should likely provide two helper plan modifiers for this case:

* Always copy known state value to plan for the attribute (attribute is never expected to change after creation)
* Conditionally copy known state value to plan for the attribute (attribute _may_ change due to other updates)

These likely would live in the `tfsdk` package, similar to [`RequiresReplace`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/tfsdk#RequiresReplace) and [`RequiresReplaceIf`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/tfsdk#RequiresReplaceIf)

### References

- https://github.com/hashicorp/terraform-plugin-framework/issues/175

bflad:
I was helping a colleague write one of these for their provider today and here is an example implementation:

```go
type ComputedValueDoesNotChangeModifier struct {}

func (m ComputedValueDoesNotChangeModifier) Description(ctx context.Context) string {
	return "Value does not change after creation."
}

func (m ComputedValueDoesNotChangeModifier) MarkdownDescription(ctx context.Context) string {
	return m.Description()
}

func (m ComputedValueDoesNotChangeModifier) Modify(ctx context.Context, req tfsdk.ModifyAttributePlanRequest, resp *tfsdk.ModifyAttributePlanResponse) {
	// Do nothing for Create. It will automatically be marked as unknown.
	if req.AttributeState == nil || req.AttributeState.IsNull() {
		return
	}

	resp.AttributePlan = req.AttributeState
}
```

PR #204, opened 2021-10-07
Add state preserving plan modifiers

kmoe:
Closes https://github.com/hashicorp/terraform-plugin-framework/issues/180

Needs careful testing.


tfsdk/attribute_plan_modification.go
```diff
@@ -118,6 +119,85 @@ func (r RequiresReplaceIfModifier) MarkdownDescription(ctx context.Context) stri
...
+}
+
+// PreserveStateModifier is an AttributePlanModifier that copies the prior state
+// value for an attribute into that attribute's plan, if that state is non-null.
+//
+// Computed attributes without the PreserveState attribute plan modifier will
+// have their value set to Unknown in the plan, so their value always will be
+// displayed as "(known after apply)" in the CLI plan output.
+// If this plan modifier is used, the prior state value will be displayed in
+// the plan instead.
```
bflad, line 134: To match the comment from the testing.

```suggestion
// the plan instead unless a prior plan modifier adjusts the value.
```

tfsdk/attribute_plan_modification.go
```diff
@@ -118,6 +119,85 @@ func (r RequiresReplaceIfModifier) MarkdownDescription(ctx context.Context) stri
 	return r.markdownDescription
 }
 
+// PreserveState returns a PreserveStateModifier.
+func PreserveState() AttributePlanModifier {
```
bflad, line 123: Naming nit (I know the worst kind! 😦 ): seeing this in a [provider implementation](https://github.com/hashicorp/terraform-provider-corner/pull/32) and with my new provider developer hat on, I initially was left wondering "What are the other attributes doing then, if they aren't preserving state? Is data being deleted? Is this scary?" As a framework developer, I happen to know this is described by the [Terraform Resource Instance Change Lifecycle](https://github.com/hashicorp/terraform/blob/main/docs/resource-instance-change-lifecycle.md), but I'm wondering if there is anything minor we could or should do to further drive home when it might be used. Maybe `PreserveComputedState()` is what I'm reaching for naming-wise?

bflad (approved):
I'm on board with this -- just some non-blocking items to consider. 🚀

tfsdk/attribute_plan_modification.go
paddycarver: How do you feel about `tfsdk.UseStateForUnknown`?

paddycarver:
Think I addressed the comments. Still feel good about it, @bflad?

bflad (approved):
Excellent 🚀

PR #208, opened 2021-10-12
types: Prevent panics when ValueFromTerraform received nil values

### LINKED ISSUE #207: tfsdk: GetAttribute panic for nested object in missing/null attribute
### Module version

```
github.com/hashicorp/terraform-plugin-framework v0.4.3-0.20211006203044-2d173e3470f6
```

I discovered this issue in a resource with this (partial) schema:

```go
return tfsdk.Schema{
        "patches": {
            Optional: true,
            Attributes: tfsdk.ListNestedAttributes(map[string]tfsdk.Attribute{
                "content": {
                    Type:     types.StringType,
                    Required: true,
                },
                "target": {
                    Optional: true,
                        "name": {
                            Type:     types.StringType,
                            Optional: true,
                        },
                        // more optional attributes...
                    }),
                },
            }, tfsdk.ListNestedAttributesOptions{}),
        },
    },
}
```

I can include additional provider code if needed, but the panic stack is entirely in the framework so hopefully the unit test reproduction included below is sufficient.

```hcl
resource "myprovider_resource" "test" {
  patches = [
    {
      content = file("/path/to/file.patch"),
      target = {
        name = "test"
      }
    }
  ]
}
```

The resource exists in state and was initially created without the `patches` property. Adding `patches` and planning triggers the crash.

### Expected Behavior

Terraform executes a plan without crashing, showing `patches` changing from empty/unset to the content above.

### Actual Behavior

The provider panics:

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x11e2385]

goroutine 25 [running]:
github.com/hashicorp/terraform-plugin-framework/types.ObjectType.ValueFromTerraform(0xc000174c60, 0x18fe2b0, 0xc00079e140, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/Users/bkeyes/code/my-provider/vendor/github.com/hashicorp/terraform-plugin-framework/types/object.go:55 +0x65
github.com/hashicorp/terraform-plugin-framework/tfsdk.State.GetAttribute(0x1903f28, 0xc00064e6f0, 0x1737b00, 0xc00017bd10, 0xc000156330, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/Users/bkeyes/code/my-provider/vendor/github.com/hashicorp/terraform-plugin-framework/tfsdk/state.go:74 +0x624
github.com/hashicorp/terraform-plugin-framework/tfsdk.Attribute.modifyPlan(0x0, 0x0, 0x1904d78, 0xc000156420, 0x0, 0x0, 0x0, 0x0, 0x100, 0x0, ...)
	/Users/bkeyes/code/my-provider/vendor/github.com/hashicorp/terraform-plugin-framework/tfsdk/attribute.go:458 +0x19a
github.com/hashicorp/terraform-plugin-framework/tfsdk.modifyAttributesPlans(0x18fe2b0, 0xc00079e140, 0xc0001563f0, 0xc0004ad368, 0x1903f28, 0xc000121470, 0x1737b00, 0xc000156d50, 0xc000156330, 0x0, ...)
	/Users/bkeyes/code/my-provider/vendor/github.com/hashicorp/terraform-plugin-framework/tfsdk/schema.go:245 +0x528
github.com/hashicorp/terraform-plugin-framework/tfsdk.modifyAttributesPlans(0x18fe2b0, 0xc00079e140, 0xc000156330, 0xc0001450c0, 0x1903f28, 0xc000121470, 0x1737b00, 0xc000156d50, 0xc000156330, 0x0, ...)
	/Users/bkeyes/code/my-provider/vendor/github.com/hashicorp/terraform-plugin-framework/tfsdk/schema.go:275 +0xae8
github.com/hashicorp/terraform-plugin-framework/tfsdk.Schema.modifyAttributePlans(...)
	/Users/bkeyes/code/my-provider/vendor/github.com/hashicorp/terraform-plugin-framework/tfsdk/schema.go:222
github.com/hashicorp/terraform-plugin-framework/tfsdk.(*server).planResourceChange(0xc00046a5a0, 0x18fe2b0, 0xc00079e140, 0xc0005d2460, 0xc000145a90)
	/Users/bkeyes/code/my-provider/vendor/github.com/hashicorp/terraform-plugin-framework/tfsdk/serve.go:810 +0x1848
github.com/hashicorp/terraform-plugin-framework/tfsdk.(*server).PlanResourceChange(0xc00046a5a0, 0x18fe2b0, 0xc00079e100, 0xc0005d2460, 0xc00079e100, 0x17e40a0, 0xc000156000)
	/Users/bkeyes/code/my-provider/vendor/github.com/hashicorp/terraform-plugin-framework/tfsdk/serve.go:635 +0xc8
github.com/hashicorp/terraform-plugin-go/tfprotov6/tf6server.(*server).PlanResourceChange(0xc00023ea00, 0x18fe358, 0xc00079e100, 0xc00045e000, 0xc00023ea00, 0xc000156030, 0xc000089ba0)
	/Users/bkeyes/code/my-provider/vendor/github.com/hashicorp/terraform-plugin-go/tfprotov6/tf6server/server.go:314 +0xb5
github.com/hashicorp/terraform-plugin-go/tfprotov6/internal/tfplugin6._Provider_PlanResourceChange_Handler(0x17e40a0, 0xc00023ea00, 0x18fe358, 0xc000156030, 0xc000280300, 0x0, 0x18fe358, 0xc000156030, 0xc000168000, 0x42a6)
	/Users/bkeyes/code/my-provider/vendor/github.com/hashicorp/terraform-plugin-go/tfprotov6/internal/tfplugin6/tfplugin6_grpc.pb.go:363 +0x214
google.golang.org/grpc.(*Server).processUnaryRPC(0xc0003a3180, 0x19052b8, 0xc000603200, 0xc0000c8200, 0xc00046a690, 0x1cf0de8, 0x0, 0x0, 0x0)
	/Users/bkeyes/code/my-provider/vendor/google.golang.org/grpc/server.go:1194 +0x52b
google.golang.org/grpc.(*Server).handleStream(0xc0003a3180, 0x19052b8, 0xc000603200, 0xc0000c8200, 0x0)
	/Users/bkeyes/code/my-provider/vendor/google.golang.org/grpc/server.go:1517 +0xd05
google.golang.org/grpc.(*Server).serveStreams.func1.2(0xc000038290, 0xc0003a3180, 0x19052b8, 0xc000603200, 0xc0000c8200)
	/Users/bkeyes/code/my-provider/vendor/google.golang.org/grpc/server.go:859 +0xab
created by google.golang.org/grpc.(*Server).serveStreams.func1
	/Users/bkeyes/code/my-provider/vendor/google.golang.org/grpc/server.go:857 +0x1fd
```

### Steps to Reproduce

Add this test to `StateGetAttributeTest` in `tfsdk/state_test.go` and run the tests:

```go
"WithAttributeName-ListNestedAttributes-null-WithElementKeyInt-WithAttributeName-object": {
    state: State{
        Raw: tftypes.NewValue(tftypes.Object{
                "test": tftypes.List{
                    ElementType: tftypes.Object{
                            "sub_test": tftypes.Object{
                                    "value": tftypes.String,
                                },
                            },
                        },
                    },
                },
                "other": tftypes.Bool,
            },
        }, map[string]tftypes.Value{
            "test": tftypes.NewValue(tftypes.List{
                ElementType: tftypes.Object{
                        "sub_test": tftypes.Object{
                                "value": tftypes.String,
                            },
                        },
                    },
                },
            }, nil),
            "other": tftypes.NewValue(tftypes.Bool, nil),
        }),
        Schema: Schema{
                "test": {
                    Attributes: ListNestedAttributes(map[string]Attribute{
                        "sub_test": {
                            Attributes: SingleNestedAttributes(map[string]Attribute{
                                "value": {
                                    Type:     types.StringType,
                                    Optional: true,
                                },
                            }),
                            Optional: true,
                        },
                    }, ListNestedAttributesOptions{}),
                    Optional: true,
                },
                "other": {
                    Type:     types.BoolType,
                    Optional: true,
                },
            },
        },
    },
    path: tftypes.NewAttributePath().WithAttributeName("test").WithElementKeyInt(0).WithAttributeName("sub_test"),
    expected: types.Object{
        Null:      true,
        AttrTypes: map[string]attr.Type{"value": types.StringType},
    },
},
```

### References

- I originally encountered the problem from #150 and updated to the latest development commit, which led to this panic
- The fix from #185 creates a situation where a `tftypes.Value` with a `nil` type is passed to `types.Object#ValueFromTerraform`.

### Other Thoughts

In `types.Object#ValueFromTerraform`, I can fix the failing test by either:

1. Adding a check that `in.Type()` is not `nil` before calling `Equal`
2. Moving the `IsKnown` and `IsNull` checks before the type check

I can submit a PR with the test and a fix, but I'm not sure which option is best, or if there are any other similar cases that need fixes.

bluekeyes:
When getting an object under a missing (null) attribute, `GetAttribute` would panic instead of returning null. Also add a unit test for this case.

Fixes #207.


bflad (changes requested):
Hi @bluekeyes 👋  Thank you for submitting this. Fixing the panic certainly seems ideal in this situation.

In terms of implementation, the preference should likely be the following for the type implementations:

- Check for `nil` first, return null type in that situation
- Perform the type check; types should ensure they align correctly upfront
- Perform the null/unknown checks

This class of panic fix should be performed in all framework defined types in the `types` package and a quick `nil` input unit test should be included for each type's `ValueFromTerraform` such as `TestObjectTypeValueFromTerraform`. The `TestStateGetAttribute` unit testing addition should likely also get replicated across `Config` and `Plan` to try and prevent future regressions should those implementations diverge for whatever reason.

Please let us know if you have time to make these changes and thanks again for reporting this and contributing! 😄

bflad:
Oh! Also feel free to create a CHANGELOG entry as well for this by creating a `.changelog/208.txt` file with contents such as:

``````markdown
```release-note:bug
types: Prevented panics when `ValueFromTerraform` received `nil` values
```
``````

File: types/list_test.go [hunk 1]
```diff
@@ -153,20 +153,48 @@ func TestListTypeValueFromTerraform(t *testing.T) {
 				},
 			},
 		},
+		"wrong-type": {
```
bluekeyes, line 156: Unrelated to the fix here, but I added these tests for consistency with objects and maps, since they were missing from lists and sets.

tfsdk/config_test.go
```diff
@@ -329,6 +329,66 @@ func TestConfigGetAttribute(t *testing.T) {
 			path:     tftypes.NewAttributePath().WithAttributeName("test").WithElementKeyInt(0).WithAttributeName("sub_test"),
 			expected: types.String{Null: true},
 		},
+		"WithAttributeName-ListNestedAttributes-null-WithElementKeyInt-WithAttributeName-Object": {
```
bluekeyes, line 332: Looking at this now, I think the original problem was also possible if the nested item here was a list, map, or set instead of an object. Do you think it's worth testing those cases explicitly too or should the unit tests on the value types cover that?

types/object.go
```diff
@@ -52,6 +52,10 @@ func (o ObjectType) ValueFromTerraform(ctx context.Context, in tftypes.Value) (a
 	object := Object{
 		AttrTypes: o.AttrTypes,
 	}
+	if in.Type() == nil {
+		object.Null = true
+		return object, nil
```
bluekeyes, line 57: This could possibly also return `nil, nil`, but I wasn't sure if everything that called `ValueFromTerraform` was prepared to check for a `nil` value if the error was also `nil`.

It seemed reasonable that if converting a value with no type to a value with a type the result would be a null value with the target type, but I could see arguments for other conventions too.

bluekeyes (commented):
Thanks for the review - I think I've addressed your main concerns. I did not add the `nil` type check to any of the primitive types, as they don't do any explicit type checks right now, deferring to `tftypes.Value.As` instead, but I could go through and do this if you'd like. It felt a bit weird without any other existing type checks.

types/object.go
bflad: I think returning the correct type with a null value is probably best here. 👍

types/list_test.go
bflad: Thank you!

tfsdk/config_test.go
bflad: If you're inclined, the additional unit testing to ensure we testing in-depth would be appreciated, but I wouldn't want to hold up these fixes on it since its covered closer to the source.

bflad (approved):
Thank you again, @bluekeyes, looks good to me 🚀

PR #213, opened 2021-10-27
Complicate our RequiresReplace(If) logic.

### LINKED ISSUE #187: Strange behavior with `Computed` attributes after v0.4.2 upgrade
I upgraded to `terraform-plugin-framework@v0.4.2` and am now seeing some strange behavior.
A resource has an attribute

```go
"function_name": {
	Description: "The name of the Lambda function, up to 64 characters in length. If you don't specify a name, AWS CloudFormation generates one.",
	Optional:    true,
	Computed:    true,
		validate.StringLenAtLeast(1),
	},
		tfsdk.RequiresReplace(),
	},
},
```

When `terraform plan` shows a difference in another attribute (`description` in this case which is _Optional_ and is now getting a value):

```
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
-/+ destroy and then create replacement

  # cloud_service_thing.test must be replaced
-/+ resource "cloud_service_thing" "test" {
      ~ arn                 = "..." -> (known after apply)
      + description         = "test"
      ~ function_name       = "1WlhknThPCtZOiZjwt283Olmq-0A8jrKEtMBpJ" -> (known after apply) # forces replacement
      ~ id                  = "1WlhknThPCtZOiZjwt283Olmq-0A8jrKEtMBpJ" -> (known after apply)
        # (9 unchanged attributes hidden)
    }

Plan: 1 to add, 0 to change, 1 to destroy.
```

The attribute `function_name` (which hasn't changed) is showing a change (and a replacement as it's _ForceNew_).
`arn` and `id` are also both _Computed_ attributes and their values have not changed.
This behavior was not observed with `v0.4.0`.

paddycarver:
Fixes #187, supercedes #203.

This changes the logic on our RequiresReplace function in the following
ways:

* if the attribute being modified has not changed, RequiresReplace isn't
  changed. This prevents the resource from being destroyed and recreated
  when _any_ attribute changes, limiting it only to the attribute with
  RequiresReplace set on it.
* if the attribute being modified is computed and is null in the config,
  RequiresReplace isn't changed. This prevents the resource from being
  destroyed and recreated when the provider returns an unknown value in
  the plan, or even a new, provider-controlled value, because it's
  unlikely the practitioners expect their resource to be destroyed and
  recreated when an attribute out of their control changes. This has the
  unfortunate side-effect that practitioners removing an
  Optional+Computed field from their config will not prompt the default
  RequiresReplace behavior, but providers can specify their own, special
  plan modifier that has a better understanding of practitioner intent
  in that case.
* if the resource is being created or destroyed, RequiresReplace isn't
  changed. This prevents the resource from being marked for destruction
  and recreation when it's just now being created or destroyed, which
  would be nonsensical.

RequiresReplaceIf gets all these changes and an additional change that
it will only set RequiresReplace to true, never to false, as its Modify
method's GoDoc indicated. This means that RequiresReplaceIf functions
can be chained, and that RequiresReplaceIf won't overwrite previous plan
modifiers' determinations on whether or not the resource should be
destroyed and recreated.

Tests were added for RequiresReplace and RequiresReplaceIf to test these
invariants.

### DIFF
--- a/tfsdk/attribute_plan_modification.go
+++ b/tfsdk/attribute_plan_modification.go
@@ -199,9 +199,9 @@ func (r RequiresReplaceIfModifier) Modify(ctx context.Context, req ModifyAttribu
 	res, diags := r.f(ctx, req.AttributeState, req.AttributeConfig, req.AttributePath)
 	resp.Diagnostics.Append(diags...)
 
-	// if the function says to require replacing, we require replacing
-	// if the function says not to, we don't change the value that other
-	// plan modifiers may have set
+	// If the function says to require replacing, we require replacing.
+	// If the function says not to, we don't change the value that prior
+	// plan modifiers may have set.
 	if res {
 		resp.RequiresReplace = true
 	}
--- a/tfsdk/attribute_plan_modification.go
+++ b/tfsdk/attribute_plan_modification.go
@@ -204,6 +204,8 @@ func (r RequiresReplaceIfModifier) Modify(ctx context.Context, req ModifyAttribu
 	// plan modifiers may have set.
 	if res {
 		resp.RequiresReplace = true
+	} else if resp.RequiresReplace {
+		// TODO: log that we didn't override the result
 	}
 }
 
--- a/tfsdk/attribute_plan_modification.go
+++ b/tfsdk/attribute_plan_modification.go
@@ -65,6 +65,52 @@ type RequiresReplaceModifier struct{}
 
 // Modify sets RequiresReplace on the response to true.
 func (r RequiresReplaceModifier) Modify(ctx context.Context, req ModifyAttributePlanRequest, resp *ModifyAttributePlanResponse) {
+	if req.AttributeConfig == nil || req.AttributePlan == nil || req.AttributeState == nil {
+		// shouldn't happen, but let's not panic if it does
+		return
+	}
+
+	if req.State.Raw.IsNull() {
+		// if we're creating the resource, no need to delete and
+		// recreate it
+		return
+	}
+
+	if req.Plan.Raw.IsNull() {
+		// if we're deleting the resource, no need to delete and
+		// recreate it
+		return
+	}
+
+	attrSchema, err := req.State.Schema.AttributeAtPath(req.AttributePath)
+	if err != nil {
+		resp.Diagnostics.AddAttributeError(req.AttributePath,
+			"Error finding attribute schema",
+			fmt.Sprintf("An unexpected error was encountered retrieving the schema for this attribute. This is always a bug in the provider.\n\nError: %s", err),
+		)
+		return
+	}
+
+	configRaw, err := req.AttributeConfig.ToTerraformValue(ctx)
+	if err != nil {
+		resp.Diagnostics.AddAttributeError(req.AttributePath,
+			"Error converting config value",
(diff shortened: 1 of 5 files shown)


tfsdk/attribute_plan_modification.go
```diff
@@ -103,9 +150,61 @@ type RequiresReplaceIfModifier struct {
...
+		return
+	}
+
 	res, diags := r.f(ctx, req.AttributeState, req.AttributeConfig, req.AttributePath)
 	resp.Diagnostics.Append(diags...)
-	resp.RequiresReplace = res
+
+	// if the function says to require replacing, we require replacing
+	// if the function says not to, we don't change the value that other
+	// plan modifiers may have set
```
bflad, line 204: I'm finding the comment wording here confusing, was this the intended meaning?

```suggestion
	// If the function says to require replacing, we require replacing.
	// If the function says not to, we don't change the value that prior
	// plan modifiers may have set.
```

That said, it also seems like we may want warning diagnostics or logging in the converse situation so provider developers can get some indication of the behavior that occurred. Of course, that probably means logging in a lot more places, so not sure how best to capture that as a TODO.

tfsdk/attribute_plan_modification_test.go
```diff
@@ -0,0 +1,682 @@
...
+		plan         Plan
+		config       Config
+		path         *tftypes.AttributePath
+		expectedPlan attr.Value
+		expectedRR   bool
+	}
+
+	schema := Schema{
+		Attributes: map[string]Attribute{
+			"a": {
```
bflad, line 28: Nit: It could be clearer down in the test cases if these names are more descriptive, e.g. `optional-computed` and `optional`

tfsdk/attribute_plan_modification_test.go
```diff
@@ -0,0 +1,682 @@
...
+				State:           tc.state,
+				Plan:            tc.plan,
+				AttributeConfig: attrConfig,
+				AttributeState:  attrState,
+				AttributePlan:   attrPlan,
+				ProviderMeta:    Config{},
+			}
+			resp := &ModifyAttributePlanResponse{
+				AttributePlan: req.AttributePlan,
+			}
```
bflad, line 665: To capture the final conditional, should we augment this to include `RequiresReplace: tc.priorRR` and add a unit testing that has `priorRR` that matches the `tc.path`? Then we should be able to verify prior RR is preserved when `tc.ifReturn` is false.

tfsdk/attribute_plan_modification.go
```diff
@@ -64,6 +65,52 @@ type RequiresReplaceModifier struct{}
 
 // Modify sets RequiresReplace on the response to true.
```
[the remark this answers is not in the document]
bflad: If/when this documentation is updated, we can probably close out https://github.com/hashicorp/terraform-plugin-framework/issues/189 as well. 👍

tfsdk/attribute_plan_modification.go
```diff
@@ -103,9 +150,61 @@ type RequiresReplaceIfModifier struct {
 // Modify sets RequiresReplace on the response to true if the conditional
 // RequiresReplaceIfFunc returns true.
```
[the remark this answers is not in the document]
bflad: If/when this documentation is updated, we can probably close out https://github.com/hashicorp/terraform-plugin-framework/issues/189 as well. 👍

paddycarver:
Think I addressed all the comments. Still look good, @bflad?

PR #224, opened 2021-11-11
tfsdk: Support block plan modification

### LINKED ISSUE #222: Support Block Plan Modification
### Module version

```
v0.4.2
```

### Use-cases

The initial block support of #85 may not contain plan modification support for the block itself or underlying attributes to reduce the complexity of first round code changes. This support should be added for parity with attributes and to prevent provider developer confusion.

### Proposal

Add `PlanModifiers` field to the `Block` type and wire with schema plan modification. Ensure each `Attribute` under `Block.Attributes` properly handles their `PlanModifiers` field.

bflad:
Closes #222

Adds `PlanModifiers` within `Block`s and ensures any children attribute and block plan modifiers of a `Block` are executed.

No changelog entry is necessary unless there is a framework release prior to this support.

### DIFF
--- a/tfsdk/attribute.go
+++ b/tfsdk/attribute.go
@@ -478,10 +478,10 @@ func (a Attribute) validateAttributes(ctx context.Context, req ValidateAttribute
 }
 
 // modifyPlan runs all AttributePlanModifiers
-func (a Attribute) modifyPlan(ctx context.Context, req ModifyAttributePlanRequest, resp *ModifyAttributePlanResponse, schemaResp *ModifySchemaPlanResponse) {
+func (a Attribute) modifyPlan(ctx context.Context, req ModifyAttributePlanRequest, resp *ModifySchemaPlanResponse) {
 	attrConfig, diags := req.Config.getAttributeValue(ctx, req.AttributePath)
 	resp.Diagnostics.Append(diags...)
-	schemaResp.Diagnostics = resp.Diagnostics
+
 	// Only on new errors.
 	if diags.HasError() {
 		return
@@ -490,7 +490,7 @@ func (a Attribute) modifyPlan(ctx context.Context, req ModifyAttributePlanReques
 
 	attrState, diags := req.State.getAttributeValue(ctx, req.AttributePath)
 	resp.Diagnostics.Append(diags...)
-	schemaResp.Diagnostics = resp.Diagnostics
+
 	// Only on new errors.
 	if diags.HasError() {
 		return
@@ -499,27 +499,25 @@ func (a Attribute) modifyPlan(ctx context.Context, req ModifyAttributePlanReques
 
 	attrPlan, diags := req.Plan.getAttributeValue(ctx, req.AttributePath)
 	resp.Diagnostics.Append(diags...)
-	schemaResp.Diagnostics = resp.Diagnostics
+
 	// Only on new errors.
 	if diags.HasError() {
 		return
 	}
 	req.AttributePlan = attrPlan
-	resp.AttributePlan = attrPlan
 
+	var requiresReplace bool
 	for _, planModifier := range a.PlanModifiers {
 		modifyResp := &ModifyAttributePlanResponse{
-			AttributePlan:   resp.AttributePlan,
-			RequiresReplace: resp.RequiresReplace,
+			AttributePlan:   req.AttributePlan,
+			RequiresReplace: requiresReplace,
 		}
 
 		planModifier.Modify(ctx, req, modifyResp)
 
 		req.AttributePlan = modifyResp.AttributePlan
-		resp.AttributePlan = modifyResp.AttributePlan
 		resp.Diagnostics.Append(modifyResp.Diagnostics...)
-		resp.RequiresReplace = modifyResp.RequiresReplace
-		schemaResp.Diagnostics = resp.Diagnostics
+		requiresReplace = modifyResp.RequiresReplace
 
 		// Only on new errors.
 		if modifyResp.Diagnostics.HasError() {
@@ -527,13 +525,12 @@ func (a Attribute) modifyPlan(ctx context.Context, req ModifyAttributePlanReques
 		}
 	}
 
-	if resp.RequiresReplace {
-		schemaResp.RequiresReplace = append(schemaResp.RequiresReplace, req.AttributePath)
+	if requiresReplace {
+		resp.RequiresReplace = append(resp.RequiresReplace, req.AttributePath)
 	}
 
-	setAttrDiags := schemaResp.Plan.SetAttribute(ctx, req.AttributePath, resp.AttributePlan)
+	setAttrDiags := resp.Plan.SetAttribute(ctx, req.AttributePath, req.AttributePlan)
 	resp.Diagnostics.Append(setAttrDiags...)
-	schemaResp.Diagnostics = resp.Diagnostics
 
 	if setAttrDiags.HasError() {
 		return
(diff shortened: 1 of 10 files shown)


tfsdk/serve_resource_attribute_plan_modifiers_test.go
```diff
@@ -18,12 +18,11 @@ func (rt testServeResourceTypeAttributePlanModifiers) GetSchema(_ context.Contex
 			"name": {
 				Required: true,
 				Type:     types.StringType,
-				// For the purposes of testing, these plan modifiers behave
-				// differently for certain values of the attribute.
-				// By default, they do nothing.
 				PlanModifiers: []AttributePlanModifier{
 					testWarningDiagModifier{},
-					testErrorDiagModifier{},
```
paddycarver, line 26: Is there a reason we're dropping `testErrorDiagModifier{}`? I see stuff getting shuffled around and cleaned up, which is great, but this looks like a behavior change that I'm not understanding the rationale of.

tfsdk/attribute.go
```diff
@@ -477,9 +478,10 @@ func (a Attribute) validateAttributes(ctx context.Context, req ValidateAttribute
 }
 
 // modifyPlan runs all AttributePlanModifiers
-func (a Attribute) modifyPlan(ctx context.Context, req ModifyAttributePlanRequest, resp *ModifyAttributePlanResponse) {
+func (a Attribute) modifyPlan(ctx context.Context, req ModifyAttributePlanRequest, resp *ModifyAttributePlanResponse, schemaResp *ModifySchemaPlanResponse) {
```
paddycarver, line 481: Two things on this:

1. Could we avoid having the double responses by just refactoring this to be at the schema level, instead of at the attribute level? Meaning it gets a `ModifySchemaPlanRequest`, a `*tftypes.AttributePath` and a `*ModifySchemaPlanResponse` and just updates them for that attribute? That would avoid the double bookkeeping, right?
2. I think in other places we have decided that the children should have their plan modifiers run before the parents, e.g. the elements before the list, so that the list could assume the plan for the elements was accurate when its plan modifier ran. This looks like it's running on the parent first, then on the elements. Is there a reason it needs to be this way, or can we talk about doing the children first, then the parent?

tfsdk/serve_resource_attribute_plan_modifiers_test.go
bflad: The original implementation of this test was dependent on the value being equal to the `TESTDIAG` string to conditionally perform the behavior of returning an error diagnostic or not. When the `testWarningDiagModifer` and `testErrorDiagModifier` were modified to return a warning/error unconditionally to make them easier to use in other tests (such as list block plan modifier testing), it had the side-effect of causing _this_ testing to always fail with returning that error rather than happening conditionally based on other parts of the test setup.

Happy to change this however.

tfsdk/attribute.go
bflad: 1. We can probably drop the `ModifyAttributePlanResponse` double bookkeeping, I think that should just work. 👍  For the request, if we're going to need to pass extra information (path) to be able to use `ModifySchemaPlanRequest`, I think sticking with `ModifyAttributePlanRequest` which already has that field or creating a separate type might be best.
2. If I'm reading the current logic correct, it does the current attribute plan modifiers before children: https://github.com/hashicorp/terraform-plugin-framework/blob/33080f200a6d9f58eb9b8a9e8e9c5e0a4483e27c/tfsdk/schema.go#L311-L343 -- if we want to change this, its probably worth thinking about the design some more because there are likely arguments either which way on which to run first.

bflad: https://github.com/hashicorp/terraform-plugin-framework/pull/224/commits/386ac38a8a7b2449fe38a89d809d3305c976cb0e drops the double response in preference of just `ModifySchemaPlanResponse`.

paddycarver: Oops! Sorry. My bad re: 2.

I do think the doing it children first, then parent makes more sense, just because then the parent can see the modified children for, e.g., comparing for equality, checking values, etc. If we do parent first, then children, we're either working with out of date children (in which case, what happens if both modify the child in conflicting ways? Who wins?) or we get the updated children, but no way to influence the parent on the results of the final child values.

I dunno if that makes sense or not, but that's my two cents.

tfsdk/block_test.go
```diff
@@ -5,13 +5,765 @@ import (
...
+								"nested_attr": tftypes.NewValue(tftypes.String, nestedAttrValue),
+							},
+						),
+					},
+				),
+			},
+		)
+	}
+
+	var schemaNullTfValue tftypes.Value = tftypes.NewValue(
```
paddycarver, line 76: Tiny nitpick:

if `schemaTfValue` accepted a `tftypes.Value` that it used as the value of `nested_attr`, we could avoid having two of these helpers while still avoiding repeating large, large blocks of verbose object construction.

Not a big deal, if you don't like it as much we don't need to do it, just mentioning it.

tfsdk/block_test.go
```diff
@@ -5,13 +5,765 @@ import (
...
+							"nested_attr": tftypes.String,
+						},
+					},
+				},
+				nil,
+			),
+		},
+	)
+
+	modifyAttributePlanRequest := func(attrPath *tftypes.AttributePath, schema Schema, configValue, planValue, stateValue string) ModifyAttributePlanRequest {
```
paddycarver, line 102: not a big deal, but if we make config, plan, and state members of a struct  type (e.g.

```go
type values struct {
config string
plan string
state string
}
```
) it may be a little easier to follow inside tests what the values are getting set to, because we can use what are essentially named arguments. I keep finding myself scrolling back up here to see what order they're in (especially since my brain keeps thinking it's state, config, plan).

paddycarver (approved):
Some minor comments. I think this looks good as is, really awesome work.

We can either switch the order of nested plan modifiers here, or we can split that out into another PR so this one can land, if you'd like.

Great tests, I love the functions-for-schema-and-values thing and think we could probably cut down on our test LoC significantly if we embraced that a bit more. I also really like the cleanup of moving `modifyPlan` to methods on the attribute and block types.

tfsdk/attribute.go
bflad: Created https://github.com/hashicorp/terraform-plugin-framework/issues/225 for future noodling! 🍝

tfsdk/attribute.go
```diff
@@ -477,9 +478,10 @@ func (a Attribute) validateAttributes(ctx context.Context, req ValidateAttribute
 }
 
 // modifyPlan runs all AttributePlanModifiers
-func (a Attribute) modifyPlan(ctx context.Context, req ModifyAttributePlanRequest, resp *ModifyAttributePlanResponse) {
+func (a Attribute) modifyPlan(ctx context.Context, req ModifyAttributePlanRequest, resp *ModifySchemaPlanResponse) {
```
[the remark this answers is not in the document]
bflad: Thank you for the earlier suggestion!

tfsdk/block_test.go
bflad: Sorry just to clarify, in `schemaTfValue` are you thinking that we should pass in the list value for each invocation? e.g. this part

```go
					[]tftypes.Value{
						tftypes.NewValue(
							tftypes.Object{
									"nested_attr": tftypes.String,
								},
							},
							map[string]tftypes.Value{
								"nested_attr": tftypes.NewValue(tftypes.String, nestedAttrValue),
							},
						),
					},
```

It's already accepting the string value for `nested_attr`. The `schemaNullTfValue` helper is handling where that list value is `nil`. Just trying to make sure I understand, thanks!

paddycarver: You understand, I'm just confused.

I apparently briefly thought the `tftypes.String` in `nested_attr` was getting set to `null`, which is not what's happening here at all. I'm just being wrong.

Sorry, ignore this.

tfsdk/block_test.go
bflad: I agree! It is less than ideal to not have that field identifier context for knowing which value is for which data. I like this approach and will followup with the test refactoring.

paddycarver (approved):
Still looks great to me :D

PR #226, opened 2021-11-19
Update verbiage to technical preview.

paddycarver:
We're in a way less experimental state now, and are more accurately
classified as a technical preview.


bflad (approved):
Looks good to me 🚀

PR #231, opened 2021-12-03
Refactor ToTerraformValue to return a tftypes.Value.

### LINKED ISSUE #171: Consider making `attr.Value.ToTerraformValue` return a `tftypes.Value`
It's honestly pretty bizarre that [`attr.Value.ToTerraformValue`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/attr#Value.ToTerraformValue) wants type developers to return an `interface{}` that needs to be one of the types accepted by [`tftypes.NewValue`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-go/tftypes#NewValue). It means that nothing ever happens with those values except calling `tftypes.NewValue` and including them, which just makes code dealing with that unnecessarily verbose. It also introduces a weird, murky integration point that seems at odds with the rest of the framework in that it doesn't really take advantage of any compile-time checks that what you're doing is valid and there's no real reason it can't.

I think the reason I originally pitched it like that was because `attr.Type` was returning the `tftypes.Type`, and so rather than making the type developer specify the `tftypes.Type` in both places, I figured the framework could just do it, because it already had that information. But that's not really a good enough justification for how bizarre that integration point is.

I think we should probably consider just asking for a `tftypes.Value` from that method, instead.

paddycarver:
Refactor the `attr.Value.ToTerraformValue` to return a `tftypes.Value`,
which simplifies a bunch of code and makes the interface safer by
allowing it to be type checked by the compiler. It did require quite a
bit of refactoring of internal code, however.

This also simplifies checking whether an arbitrary `attr.Value` is null
or unknown considerably.

Fixes #171.

### DIFF
--- a/types/number.go
+++ b/types/number.go
@@ -58,7 +58,7 @@ func (n Number) ToTerraformValue(_ context.Context) (tftypes.Value, error) {
 		return tftypes.NewValue(tftypes.Number, nil), nil
 	}
 	if err := tftypes.ValidateValue(tftypes.Number, n.Value); err != nil {
-		return tftypes.NewValue(tftypes.Number, tftypes.UnknownValue), nil
+		return tftypes.NewValue(tftypes.Number, tftypes.UnknownValue), err
 	}
 	return tftypes.NewValue(tftypes.Number, n.Value), nil
 }
--- a/types/number.go
+++ b/types/number.go
@@ -45,20 +45,22 @@ func (n Number) Type(_ context.Context) attr.Type {
 	return NumberType
 }
 
-// ToTerraformValue returns the data contained in the *Number as a *big.Float.
-// If Unknown is true, it returns a tftypes.UnknownValue. If Null is true, it
-// returns nil.
-func (n Number) ToTerraformValue(_ context.Context) (interface{}, error) {
+// ToTerraformValue returns the data contained in the *Number as a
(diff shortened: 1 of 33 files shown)


types/number.go
```diff
@@ -45,20 +45,22 @@ func (n Number) Type(_ context.Context) attr.Type {
...
-		return tftypes.UnknownValue, nil
+		return tftypes.NewValue(tftypes.Number, tftypes.UnknownValue), nil
 	}
 	if n.Value == nil {
-		return nil, nil
+		return tftypes.NewValue(tftypes.Number, nil), nil
 	}
-	return n.Value, nil
+	if err := tftypes.ValidateValue(tftypes.Number, n.Value); err != nil {
+		return tftypes.NewValue(tftypes.Number, tftypes.UnknownValue), nil
```
bflad, line 61: I believe this should be returning the `ValidateValue()` error 😄 

```suggestion
		return tftypes.NewValue(tftypes.Number, tftypes.UnknownValue), err
```

bflad (approved):
Much cleaner! Two things, otherwise, looks good to me 🚀

paddycarver:
Fixes look good?

bflad (approved):
Looks good to me 🚀

Issue #221, opened 2021-10-29
nil TypeMap permadiff 


appilon:
### Module version

```
0.4.2
```

### Relevant provider source code

```go
"permissions": {
				Type:        types.MapType{ElemType: types.BoolType},
				Optional:    true,
			},
```

### Terraform Configuration Files

```hcl
resource "sample_profile" "test" {
  name            = "test"
  description     = "test update"
  # permissions = {}
}
```

### Expected Behavior

After applying the initial config, I expect a subsequent plan to be noop.
### Actual Behavior

Terraform seems to think the current state has an empty set (but not nil) which causes a diff with the plan (nil).
```
- permissions     = {} -> null
```

appilon:
I attempted to massage the plan to match
```
type fixEmptyMap struct {
	emptyDescriptions
}

func (fixEmptyMap) Modify(_ context.Context, req tfsdk.ModifyAttributePlanRequest, resp *tfsdk.ModifyAttributePlanResponse) {
	if req.AttributeState == nil {
		return
	}
	state := req.AttributeState.(types.Map)
	plan := req.AttributePlan.(types.Map)
	if len(state.Elems) == 0 && !state.Null && plan.Null {
		plan.Null = false
		plan.Elems = make(map[string]attr.Value, 0)
		resp.AttributePlan = plan
	}
}
```
But got the following error on plan
```
╷
│ Error: Provider produced invalid plan
│
│ Provider "registry.terraform.io/hashicorp/salesforce" planned an invalid value for salesforce_profile.test.permissions: planned value cty.MapValEmpty(cty.Bool) for a non-computed attribute.
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
```

bflad:
Would you be able to show how the attribute is being set into state (if it is) during `Create` and/or `Read`?

appilon:
The permissions go through a bit of an expansion from the API response but the logic is the following:
```
        // expand permissions
	permissions := make(map[string]attr.Value, len(includePermissions))
	for _, k := range includePermissions {
		v, ok := p[k]
		trimmedKey := strings.TrimPrefix(k, "Permissions")
		if ok {
			permissions[trimmedKey] = types.Bool{Value: v.(bool)}
		} else {
			// set to unknown, maybe we should panic?
			permissions[trimmedKey] = types.Bool{Unknown: true}
		}
	}
	if len(permissions) > 0 {
[...]

Issue #201, opened 2021-10-05
Unset Primitive Types on Optional Attributes with (State).Set() Generate Errors


bflad:
### Module version

```
v0.4.2
```

### Relevant provider source code

```go
package provider

import (
	"context"

	"github.com/hashicorp/terraform-plugin-framework/diag"
	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
	"github.com/hashicorp/terraform-plugin-framework/types"
)

type optionalTypesResourceType struct{}

func (t optionalTypesResourceType) GetSchema(ctx context.Context) (tfsdk.Schema, diag.Diagnostics) {
	return tfsdk.Schema{
		Attributes: map[string]tfsdk.Attribute{
			"id": {
				Type:     types.StringType,
				Computed: true,
			},
			"optional_types_bool": {
				Type:     types.BoolType,
				Optional: true,
			},
			"optional_types_float64": {
				Type:     types.Float64Type,
				Optional: true,
			},
			"optional_types_int64": {
				Type:     types.Int64Type,
				Optional: true,
			},
			"optional_types_string": {
				Type:     types.StringType,
				Optional: true,
			},
		},
	}, nil
}

func (t optionalTypesResourceType) NewResource(ctx context.Context, p tfsdk.Provider) (tfsdk.Resource, diag.Diagnostics) {
	return optionalTypesResource{}, nil
}

type optionalTypesResourceTypeData struct {
	ID                   string        `tfsdk:"id"`
	OptionalTypesBool    types.Bool    `tfsdk:"optional_types_bool"`
	OptionalTypesFloat64 types.Float64 `tfsdk:"optional_types_float64"`
	OptionalTypesInt64   types.Int64   `tfsdk:"optional_types_int64"`
	OptionalTypesString  types.String  `tfsdk:"optional_types_string"`
}

type optionalTypesResource struct{}

func (r optionalTypesResource) Create(ctx context.Context, req tfsdk.CreateResourceRequest, resp *tfsdk.CreateResourceResponse) {
	data := optionalTypesResourceTypeData{
		ID: "testing",
	}
	resp.Diagnostics.Append(resp.State.Set(ctx, data)...)
}

func (r optionalTypesResource) Read(ctx context.Context, req tfsdk.ReadResourceRequest, resp *tfsdk.ReadResourceResponse) {
	// Intentionally blank for testing.
}

[...]

PR #232, opened 2021-12-03
Allow using ValueAs into attr.Value.

### LINKED ISSUE #230: Allow "generic" use of `State.GetAttribute`
### Module version

```
v0.5.0
```

### Use-cases

The changes made in #167 prevent the "generic" use of `State.GetAttribute` to be used to copy a value, for example:

https://github.com/hashicorp/terraform-provider-awscc/blob/3cdc5b061801143a621f3d571c53e0e36b22ab96/internal/generic/state.go#L12-L26

```go
// CopyValueAtPath copies the value at a specified path from source State to destination State.
func CopyValueAtPath(ctx context.Context, dst, src *tfsdk.State, path *tftypes.AttributePath) error {
	val, diags := src.GetAttribute(ctx, path)

	if diags.HasError() {
		return tfresource.DiagsError(diags)
	}

	diags = dst.SetAttribute(ctx, path, val)

	if diags.HasError() {
		return tfresource.DiagsError(diags)
	}

	return nil
}
```

I have tried using a pointer to `attr.Value`:

```go
func CopyValueAtPath(ctx context.Context, dst, src *tfsdk.State, path *tftypes.AttributePath) error {
	var val attr.Value
	diags := src.GetAttribute(ctx, path, &val)

	if diags.HasError() {
		return tfresource.DiagsError(diags)
	}

	diags = dst.SetAttribute(ctx, path, val)

	if diags.HasError() {
		return tfresource.DiagsError(diags)
	}

	return nil
}
```

but this fails:

```
    state_test.go:471: unexpected error from CopyValueAtPath: 1 error occurred:
        	* An unexpected error was encountered trying to convert into a Terraform value. This is always an error in the provider. Please report the following to the provider developer:
        
        Cannot use attr.Value attr.Value, only types.Number is supported because types.primitive is the type in the schema
```

paddycarver:
Allow for generic handling of attr.Values by special-casing ValueAs
(and, consequently, GetAttribute) to bypass reflection when we want to
just obtain the raw attr.Value.

Fixes #230.

### DIFF
--- a/.changelog/232.txt
+++ b/.changelog/232.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+Added the ability to get an attribute as a generic `attr.Value` when using `GetAttribute`.
+```
--- a/internal/reflect/generic_attr_value.go
+++ b/internal/reflect/generic_attr_value.go
@@ -1,0 +1,12 @@
+package reflect
+
+import (
+	"context"
+	"reflect"
+
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+)
+
+func IsGenericAttrValue(ctx context.Context, target interface{}) bool {
+	return reflect.TypeOf((*attr.Value)(nil)) == reflect.TypeOf(target)
+}
--- a/tfsdk/value_as.go
+++ b/tfsdk/value_as.go
@@ -14,6 +14,10 @@ import (
 // the contents of `val`, using the reflection rules
 // defined for `Get` and `GetAttribute`.
 func ValueAs(ctx context.Context, val attr.Value, target interface{}) diag.Diagnostics {
+	if reflect.IsGenericAttrValue(ctx, target) {
+		*(target.(*attr.Value)) = val
+		return nil
+	}
 	raw, err := val.ToTerraformValue(ctx)
 	if err != nil {
 		return diag.Diagnostics{diag.NewErrorDiagnostic("Error converting value",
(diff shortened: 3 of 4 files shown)


bflad (approved):
Looks good to me 🚀 

I'd love to unit test this to match the CHANGELOG entry since we have expectations for what is to occur "at a distance", but it would not be a trivial inclusion. e.g. adding the below test case to `TestStateGetAttribute`

```go
		"attr.Value": {
			state: State{
				Raw: tftypes.NewValue(tftypes.Object{
						"name": tftypes.String,
					},
				}, map[string]tftypes.Value{
					"name": tftypes.NewValue(tftypes.String, "namevalue"),
				}),
				Schema: Schema{
						"name": {
							Type:     types.StringType,
							Required: true,
						},
					},
				},
			},
			target:   new(attr.Value),
			expected: &types.String{Value: "namevalue"},
		},
```

Yields:

```
--- FAIL: TestStateGetAttribute (0.00s)
    --- FAIL: TestStateGetAttribute/attr.Value (0.00s)
        /Users/bflad/src/github.com/hashicorp/terraform-plugin-framework/tfsdk/state_test.go:1070: unexpected value (+wanted, -got):   interface{}(
            - 	&⟪0xc00020ffb0⟫"github.com/hashicorp/terraform-plugin-framework/types".String{Value: "namevalue"},
            + 	&⟪0xc00020c7f8⟫"github.com/hashicorp/terraform-plugin-framework/types".String{Value: "namevalue"},
              )
```

Because the `types.String` `Equal` method is implemented as a value receiver and not as a pointer receiver. Semi-relates to proposal 2 of #201 and #221.

paddycarver:
Yeah, I think that's what I found. Which is a bummer. Need to think about how to do that properly.... :(

PR #234, opened 2021-12-13
Introduce some logging.

### LINKED ISSUE #63: Integration with terraform-plugin-log
As [`terraform-plugin-log`](https://github.com/hashicorp/terraform-plugin-log) is the recommended method for logging from new providers, `terraform-plugin-framework` should provide the functionality described in the documentation:

> For most plugin authors, building on an SDK or framework, the SDK or framework will take care of injecting a logger using New.

paddycarver:
Start logging using tflog. This largely just inserts logging at decision
points in the serve.go file of tfsdk, which is where the most invisible,
impactful decisions are made. Over time, logging should be introduced in
more places on an as-needed basis or as we touch those parts of the
code.

Fixes #63.


paddycarver:
Happy with that changelog?

bflad (approved):

Issue #215, opened 2021-10-27
Refactor how we're serving providers


paddycarver:
Currently, a lot of logic is living in the `tfsdk/serve.go` file. It's basically gluing together all the abstractions and helpers in the framework, and it's where the magic happens. It also is what surfaces a gRPC server that Terraform can connect to.

I think having those two concerns colocated is probably a bit of technical debt. Doing so means that testing the magic needs to be tested at the gRPC level. It also means that code to mutate values and call helpers is living right next to code that converts between the framework's types and the gRPC server's.

I think we could make the methods on the server thin wrappers around calls with more narrow function signatures and requirements, and that would allow us to write more tests more easily, because we wouldn't need to implement an entire provider to integration test against. The server methods would just be responsible for converting from gRPC types into framework types, calling these functions, and converting back. The functions would just be responsible for the mutation and calling other helpers/calling into the framework injection points.

I think this would make it a little easier to navigate, test, and understand the codebase.

bflad:
I think this pairs well with https://github.com/hashicorp/terraform-plugin-framework/issues/172 👍

Issue #84, opened 2021-07-28
Optionally relax requirements of Config.Get, Plan.Get, and State.Get


paddycarver:
### Module version

```
v0.2.0
```

### Use-cases

`Config.Get`, `Plan.Get`, and `State.Get` all require for unknown and null values to be explicitly handled, and will return an error if they're not. This can be annoying in certain situations, where unknown or null values don't matter--for example, computed-only fields will always be unknown in the plan, it's annoying to handle them explicitly instead of defaulting to the empty value, because there's no confusion about whether they were set or not.

### Attempted Solutions

We already actually have support for this built into the reflection code powering `Get`, it just isn't surfaced in `Get` right now.

### Proposal

Let's add `GetConfigOption`, `GetStateOption`, and `GetPlanOption` types that can adjust how `Get` works on those types, and add a third, variadic argument to `Get` that accepts those options. That allows for the default cause to continue working today, allows for the optional specification of the "relax, please" flag, and allows us room to add more flags in the future if it turns out we want other options.

bflad:
Over in #498 there is a proposal for the opposite approach to this issue since we have learned quite a bit about default recommendations we can suggest to prevent issues. Since this issue was raised, there has also been the introduction of `attr.Value` conversion helpers such as `tfsdk.ConvertValue()` and `tfsdk.ValueAs()`.

There is a sense that aligning every schema attribute type (`attr.Type`) with an associated value type (`attr.Value) will prevent confusion, rather than having runtime errors raised to practitioners based on a complex set of rules that a provider developer might not be familiar with, or introducing a plethora of data handling options which provider developers must discover and understand. If using a non-framework-defined Go type is important for a use case, then custom provider-defined types and the conversion helpers are there. If fetching zero-values for null or unknown values is an important use case, then `attr.Value` implementations can already enable that.

We can of course revisit this in the future, but for now I'm going to close this issue.

Aside: In #447 we are considering making the framework-defined `types` value types zero-value null instead of an "invalid" value (which tends to be a zero-value of the underlying storage type). This is two-fold to prevent the creation of these "invalid" value types while also enabling use case where an API never sends a value, so making it null by default will play nicely with Terraform's data handling.

Issue #66, opened 2021-07-12
`Plan` and `State` `SetAttribute` methods do not support setting a Null value


ewbankkit:
### Module version

```console
% go list -m github.com/hashicorp/terraform-plugin-framework/...
github.com/hashicorp/terraform-plugin-framework v0.1.0
```

### Actual Behavior

The [`Plan.SetAttribute`](https://github.com/hashicorp/terraform-plugin-framework/blob/1a7927fec93459115be87f283dd1ee7941b30578/tfsdk/plan.go#L62) and [`State.SetAttribute`](https://github.com/hashicorp/terraform-plugin-framework/blob/1a7927fec93459115be87f283dd1ee7941b30578/tfsdk/state.go#L65) methods do not currently support setting a _Null_ value via `state.SetAttribute(ctx, path, nil)`. I get the error

```
error setting value at AttributeName("identifier"): error creating new state value: don't know how to construct attr.Type from <nil> (invalid)
```

### Expected Behavior

To be able to set a _Null_ value (of the type specified at `path`) in `State` (and `Plan`, although I am not currently concerned with that case).

paddycarver:
Fundamentally, this is a problem with how the reflection code is written. We currently infer some information about what the provider tries to set in state, then compare that to the schema to ensure that it matches what the schema expects. But we start with the type information of the data passed in.

In your example, you're using an untyped `nil`, which breaks that assumption because it has no type, so it falls through the switch statement. Wrapping it in a type assertion for an appropriate type for that attribute (e.g. `*string(nil)`) should, in theory, work. As should `var val *string`.

This isn't intuitive and is annoyingly verbose, though. And there's no reason it has to be this way.

One way to resolve this is to just check, in our reflection code, if the value is `nil` and return a null value if it is, short-circuiting the type reflection entirely. But I'm suspicious that untyped constants (string literals, numerical literals, etc.) will have the same problem.

A cleaner way may just be to flip our reflection code on it's head, starting with the type the schema defines and switching on that, checking that the data fits in it on a case-by-case basis, which should let us break down the untyped constants problem a little more cleanly.

PR #237, opened 2021-12-20
Massive refactor

paddycarver:
A massive refactoring effort to:

* Change how we're serving providers (#215)
* Change how we're structuring reflection (#66, #84)
* Change where we convert from `tftypes.Value` to `attr.Value` (#172)

and clean up a few other things along the way. It turns out when you start poking at one of these, the whole ball of yarn starts unraveling. Especially since this is an offshoot of the work I was doing in #52.


paddycarver:
Oh, also, this refactors Plan/Config/State to use new types, hopefully cutting down on the amount of code needed. I deleted the plan/state/config tests, but they'll be replaced with non-duplicated tests for the `ReadOnlyData` and `*Data` types. I largely lumped this in because the thought of updating those 10,000 lines of tests made me sad.

tfsdk/data.go
```diff
@@ -0,0 +1,131 @@
...
+func setAttributeValue(ctx context.Context, path *tftypes.AttributePath, object types.Object, val attr.Value) (types.Object, diag.Diagnostics) {
+	// TODO: update object with val at path
+	return types.Object{}, nil
+}
+
+func (d *Data) Set(ctx context.Context, val interface{}, options ...SetDataOption) diag.Diagnostics {
+	return nil
+}
+
+func (d *Data) RemoveResource(ctx context.Context) {
```
bflad, line 111: Drive-by note: In principle, these new "data" handling types feel appropriate to centralize the data logic of the Config/Plan/State handling. However, I'm not sure that we should necessarily get away from having `Config`, `Plan`, and `State` types. This `RemoveResource` method is a good example of something that should only be implemented with `State` and not `Plan`.

I'm wondering if `Config`, `Plan`, and `State` should still exist with encapsulating the underlying data types, e.g. briefly

```go
type Config struct {
  ReadOnlyData
}

type Plan struct {
  *Data
}

type State struct {
  *Data
}

func (s State) RemoveResource(ctx context.Context) {/* ... */}
```

I'm also wondering if this particular change could be done outside the larger refactoring.

tfsdk/serve_config.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-go/tfprotov6"
+)
+
+// parseConfig turns a *tfprotov6.DynamicValue and a Schema into a Config.
+func parseConfig(ctx context.Context, dv *tfprotov6.DynamicValue, schema Schema) (ReadOnlyData, diag.Diagnostics) {
```
bflad, line 11: Drive-by nit: I'd be tempted to name this `newReadOnlyData` for clarity, unless its updated to return a `Config` (which I'd still opt for the new/New prefix).

tfsdk/serve_resource.go
```diff
@@ -0,0 +1,50 @@
+package tfsdk
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+)
+
+func serveGetResourceType(ctx context.Context, provider Provider, typ string) (ResourceType, diag.Diagnostics) {
```
bflad, line 10: Drive-by question: Is there a particular reason these aren't unexported methods on `Provider`? e.g.

```go
func (p Provider) getResourceType(ctx context.Context, typ string) (ResourceType, diag.Diagnostics) {/* ... */}
```

bflad (commented):
Just leaving some drive-by notes while this is in draft. 👍  It'd be super helpful to understand the desired end state (of course a little difficult while prototyping 🙁 ) and which pieces we can potentially be split out to reduce design review burden.

bflad:
There's been _dozens_ of pull requests to accomplish even part of this. 😛 This has really served as a valuable reference for various ideas though, so thank you.

PR #243, opened 2022-01-04
tfsdk: Added Debug field to ServeOpts for running providers via debugger and testing processes

### LINKED ISSUE #239: Debugger-Debugging mode
### Use-cases
When developing a provider, using just "log based" debugging may not be the best tool to use.

### Attempted Solutions
Using `.Debug` instead of `.Serve` or using the `--debug` parameter. Looking into the code it seems like it is planned but not implemented:

https://github.com/hashicorp/terraform-plugin-framework/blob/574e52cdb9f8015c92cbe28a65dd008456a0726c/tfsdk/serve.go#L47

### Proposal
Implementing the `--debug` is ideal, but if a workaround is provided for the meantime to be able to connect the debugger an an IDE it would be appreciated.

### References
none

bflad:
Closes #239
Reference: https://www.terraform.io/plugin/sdkv2/debugging

Similar to Terraform Plugin SDK, providers need to be executable via debugger and testing processes for instrumenting the provider code. In this workflow, the caller manages the provider lifecycle, rather than Terraform CLI. The Terraform CLI reattach configuration is output to stdout.

Verified in a framework-based provider by:

- `go mod edit -replace=github.com/hashicorp/terraform-plugin-framework=/Users/bflad/src/github.com/hashicorp/terraform-plugin-framework` (use this worktree)
- Updated provider `main` function code to 

```go
func main() {
	var debug bool

	flag.BoolVar(&debug, "debug", false, "set to true to run the provider with support for debuggers like delve")
	flag.Parse()

	var err error
	ctx := context.Background()
	serveOpts := tfsdk.ServeOpts{
		Name: "registry.terraform.io/bflad/framework",
	}

	if debug {
		err = tfsdk.Debug(ctx, provider.New, serveOpts)
	} else {
		err = tfsdk.Serve(ctx, provider.New, serveOpts)
	}

	if err != nil {
		fmt.Printf("Error serving provider: %s", err)
		os.Exit(1)
	}
}
```

- Added provider VSCode configuration to `.vscode/launch.json` (refer also to terraform-provider-scaffolding)

```json
        {
            "name": "Debug - Attach External CLI",
            "type": "go",
            "request": "launch",
            "mode": "debug",
            // this assumes your workspace is the root of the repo
            "program": "${workspaceFolder}",
            "env": {},
            "args": [
                // pass the debug flag for reattaching
                "-debug",
            ],
        }
```

- In provider code, set a breakpoint
- Launched debug process
- Copied TF_REATTACH_PROVIDERS string from Outputs > Go Debug
- Exported environment variable and executed Terraform CLI
- Breakpoint hit and halted further execution

### DIFF
--- a/.changelog/243.txt
+++ b/.changelog/243.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+tfsdk: Added `Debug` field to `ServeOpts` for running providers via debugger and testing processes
+```
--- a/tfsdk/serve.go
+++ b/tfsdk/serve.go
@@ -28,6 +28,12 @@ type ServeOpts struct {
 	// Name is the name of the provider, in full address form. For example:
 	// registry.terraform.io/hashicorp/random.
 	Name string
+
+	// Debug runs the provider in a mode acceptable for debugging and testing
+	// processes, such as delve, by managing the process lifecycle. Information
+	// needed for Terraform CLI to connect to the provider is output to stdout.
+	// os.Interrupt (Ctrl-c) can be used to stop the provider.
+	Debug bool
 }
 
 // NewProtocol6Server returns a tfprotov6.ProviderServer implementation based
@@ -39,12 +45,18 @@ func NewProtocol6Server(p Provider) tfprotov6.ProviderServer {
 }
 
 // Serve serves a provider, blocking until the context is canceled.
-func Serve(ctx context.Context, factory func() Provider, opts ServeOpts) error {
+func Serve(ctx context.Context, providerFunc func() Provider, opts ServeOpts) error {
+	var tf6serverOpts []tf6server.ServeOpt
+
+	if opts.Debug {
+		tf6serverOpts = append(tf6serverOpts, tf6server.WithManagedDebug())
+	}
+
 	return tf6server.Serve(opts.Name, func() tfprotov6.ProviderServer {
 		return &server{
-			p: factory(),
+			p: providerFunc(),
 		}
-	}) // TODO: set up debug serving if the --debug flag is passed
+	}, tf6serverOpts...)
 }
 
 func (s *server) registerContext(in context.Context) context.Context {


go.mod
```diff
@@ -4,6 +4,7 @@ go 1.17
 
 require (
 	github.com/google/go-cmp v0.5.6
+	github.com/hashicorp/go-plugin v1.4.3
```
bflad, line 7: Note: This is temporary until terraform-plugin-go is updated: https://github.com/hashicorp/terraform-plugin-go/pull/123

tfsdk/debug.go
```diff
@@ -0,0 +1,164 @@
...
+			String:  pluginReattachConfig.Addr.String(),
+		},
+	}, closeCh, nil
+}
+
+// Debug runs the provider in a mode acceptable for debugging and testing
+// processes, such as delve, by managing the process lifecycle. Information
+// needed for Terraform CLI to connect to the provider is output to stdout.
+// os.Interrupt (Ctrl-c) can be used to stop the provider.
+func Debug(ctx context.Context, providerFunc func() Provider, opts ServeOpts) error {
```
bflad, line 108: Aside: Instead of introducing an exported function, we could also implement this by adding a `Debug bool` field to `ServeOpts`, then triggering this logic via an unexported function. Thinking about that again, maybe that would be a cleaner implementation for providers rather than forcing them to implement the conditional between serving functions.

bflad: Opted to go with the `ServeOpts` approach!

bflad:
Switched this to using a `Debug bool` field in `ServeOpts` and its cleaner for provider implementations (and not bloating the `tfsdk` function surface area further). 👍 

Verified with this main function:

```go
func main() {
	var debug bool

	flag.BoolVar(&debug, "debug", false, "set to true to run the provider with support for debuggers like delve")
	flag.Parse()

	ctx := context.Background()
	serveOpts := tfsdk.ServeOpts{
		Debug: debug,
		Name:  "registry.terraform.io/bflad/framework",
	}

	err := tfsdk.Serve(ctx, provider.New, serveOpts)

	if err != nil {
		fmt.Printf("Error serving provider: %s", err)
		os.Exit(1)
	}
}
```

ewbankkit:
Will this work with a [muxed](https://github.com/hashicorp/terraform-plugin-mux) provider?

bflad:
terraform-plugin-mux would likely require its own debug functionality, since providers are served using the terraform-plugin-go server implementations, rather than the SDK or framework implementations. I'll open an issue to investigate that further there, thanks for the call out!

bflad:
Almost all of this code could be removed via an upstream implementation such as: https://github.com/hashicorp/terraform-plugin-go/pull/137

```go
func Serve(ctx context.Context, providerFunc func() Provider, opts ServeOpts) error {
	var tf6serverOpts []tf6server.ServeOpt

	if opts.Debug {
		tf6serverOpts = append(tf6serverOpts, tf6server.WithManagedDebug())
	}

	return tf6server.Serve(opts.Name, func() tfprotov6.ProviderServer {
		return &server{
			p: providerFunc(),
		}
	}, tf6serverOpts...)
}
```

It may be worth getting that in before recreating the implementation here.

bflad:
Rebased on terraform-plugin-go@v0.6.0 and ready for review. 👍

ewbankkit:
LGTM :rocket:.

PR #253, opened 2022-01-25
Update Contributing Guide with additional unit testing information

### LINKED ISSUE #135: Add testing practices to the contributing guide
@bflad pointed out that we don't really write down anywhere our preferences around testing. We should, in the CONTRIBUTING.md guide, write down our practices of using table-driven tests and calling `t.Parallel` (or documenting why a test can't be run in parallel, when it absolutely can't be avoided).

bflad:
Closes #135


bookshelfdave (approved):
lgtm :+1:

PR #259, opened 2022-02-07
Skip validation of a values missing in config

gzigzigzeo:
**Problem**

I have the following schema:

```
"spec": {
		"options": {
				"lock": {
					Description: "Lock specifies the locking mode (strict|best_effort) to be applied with  the role.",
					Optional:    true,
				},
				"max_sessions": {
					Description: "MaxSessions defines the maximum number of  concurrent sessions per connection.",
					Optional:    true,
					Computed:    true,
					Type:        types.Int64Type,
				},
			}),
			Computed:    true,
			Description: "Options is for OpenSSH options like agent forwarding.",
			Optional:    true,
			PlanModifiers: []tfsdk.AttributePlanModifier{tfsdk.UseStateForUnknown()},
		},
```

And the following config:

```
spec {}
```

`"spec"` is the base object, `"options"` is the child computed object, and `"max_sessions"` is the bottommost Int64 value, which, in turn, is also computed.

`options` might either be missing in my configuration at all, set partially or set completely, including `max_sessions` attribute.

When `options` section is missing in the configuration completely, Terraform crashes:

```
github.com/hashicorp/terraform-plugin-framework/types.int64Validate({0x103337498, 0x1400030b600}, {{0x0, 0x0}, {0x0, 0x0}}, 0x14000683ce0)
	/Users/gzigzigzeo/go/src/github.com/gravitational/teleport-plugins/vendor/github.com/hashicorp/terraform-plugin-framework/types/int64.go:16 +0x44
github.com/hashicorp/terraform-plugin-framework/types.primitive.Validate(0x3, {0x103337498, 0x1400030b600}, {{0x0, 0x0}, {0x0, 0x0}}, 0x14000683ce0)
	/Users/gzigzigzeo/go/src/github.com/gravitational/teleport-plugins/vendor/github.com/hashicorp/terraform-plugin-framework/types/primitive.go:121 +0xd8
github.com/hashicorp/terraform-plugin-framework/tfsdk.Config.getAttributeValue({{{0x103341820, 0x140004f8690}, {0x1031eb6a0, 0x140004f41e0}}, {0x14000558bd0, 0x0, 0x0, {0x0, 0x0}, {0x0, ...}, ...}}, ...)
```

**Solution**

That happens because Terraform Framework tries to validate the type of `spec.options.max_sessions` config value, which is missing in the config, hence, can't have a type to validate. It must be skipped.

### DIFF
--- a/tfsdk/config.go
+++ b/tfsdk/config.go
@@ -91,11 +91,13 @@ func (c Config) getAttributeValue(ctx context.Context, path *tftypes.AttributePa
 	//       If found, convert this value to an unknown value.
 	// Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/186
 
-	if attrTypeWithValidate, ok := attrType.(attr.TypeWithValidate); ok {
-		diags.Append(attrTypeWithValidate.Validate(ctx, tfValue, path)...)
+	if err == nil {
+		if attrTypeWithValidate, ok := attrType.(attr.TypeWithValidate); ok {
+			diags.Append(attrTypeWithValidate.Validate(ctx, tfValue, path)...)
 
-		if diags.HasError() {
-			return nil, diags
+			if diags.HasError() {
+				return nil, diags
+			}
 		}
 	}
 
--- a/tfsdk/config.go
+++ b/tfsdk/config.go
@@ -90,14 +90,11 @@ func (c Config) getAttributeValue(ctx context.Context, path *tftypes.AttributePa
 	// TODO: If ErrInvalidStep, check parent paths for unknown value.
 	//       If found, convert this value to an unknown value.
 	// Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/186
+	if attrTypeWithValidate, ok := attrType.(attr.TypeWithValidate); ok {
+		diags.Append(attrTypeWithValidate.Validate(ctx, tfValue, path)...)
 
-	if err == nil {
-		if attrTypeWithValidate, ok := attrType.(attr.TypeWithValidate); ok {
-			diags.Append(attrTypeWithValidate.Validate(ctx, tfValue, path)...)
-
-			if diags.HasError() {
-				return nil, diags
-			}
+		if diags.HasError() {
+			return nil, diags
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


ewbankkit:
Similar crash reported in https://github.com/hashicorp/terraform-provider-awscc/issues/382.
In that case the offending code is https://github.com/hashicorp/terraform-plugin-framework/blob/5f18d804a22bb86da3cfadff07a58e419b9b0f73/tfsdk/state.go#L80-L100

gzigzigzeo:
@ewbankkit Thanks for pointing this out, provided failed test and fixed it as well.

bflad:
Hi @gzigzigzeo 👋  Thanks for submitting this and sorry you ran into trouble here.

In general, the preference for this framework should be performing attribute validation in all cases. This allows provider developers to check against `null` values explicitly, rather than the framework making any assumptions about what might be the correct behavior. To that end, this likely means that some of the built-in framework types need to handle this class of `nil` `tftypes.Value` issue to prevent panics.

My recommendation here would be to revert the general validation logic and instead perform a `nil` check before the problematic call into `in.Type()`, e.g.

```go
if in == nil {
  return diags
}

if !in.Type().Equal(tftypes.Number) { // ...
```

Please reach out if you have any questions (although it may take a week for a next response).

gzigzigzeo:
Hey, @bflad! Thank you for the clarification! In fact, reverting of this logic was my initial thought, and performing it in float64.go and int64.go fixed the issue. At that moment, I thought that it was a bit implicit and does not guarantee a custom user validations from failing. I now see why your approach is better and framework should not guarantee anything here. I am going to fix my implementation accordingly, add a counterpart tests for other built-in types and push it to this PR. Thanks again.

gzigzigzeo:
@bflad Done!

bflad (commented):
Thanks for the updates, @gzigzigzeo. Since I didn't want to break your `main` branch by updating it, this will get pulled in as a separate commit. I will also fix the `Set` validation (#261) as part of pulling this in.

bflad:
Merged as c584343531cb3bf1b46dad2e87631c75812593c6

Issue #42, opened 2021-06-04
Support upgrading resource state


paddycarver:
### Module version

```
N/A
```

### Use-cases

Sometimes resource schemas change over time and need massaging to be converted to the new format. This isn't usually needed; a refresh should update everything nicely. But if some values can't be populated by refresh, or refresh can't run without some values, and those values need to be migrate to a new data type or format, then provider-assisted state upgrades are necessary.

We should offer an optional method on resource types to support this callback.

### Attempted Solutions

N/A

### Proposal

Add an optional method on the Resource interface that lets providers modify the state when Terraform calls UpgradeResourceState.

PR #263, opened 2022-02-17
tfsdk: Support protocol version 5 and verify valid resource type in UpgradeResourceState RPC

### LINKED ISSUE #262: tfsdk: Send UpgradeResourceStateResponse.UpgradedState.Msgpack for Terraform 0.12-0.14 Compatibility
### Module version

```
v0.5.0
```

https://github.com/hashicorp/terraform-provider-corner/pull/44

### Expected Behavior

While the framework has not actually implemented `UpgradeResourceState` handling (refer to #42), there is currently a passthrough implementation there since the RPC is always invoked. When passing a framework provider through terraform-plugin-mux `tf6to5server` for Terraform CLI 0.12+ compatibility, there should be no errors either.

### Actual Behavior

Terraform CLI 0.12 through 0.14 return the following (less than helpful) error:

```
Error: EOF
```

This is because those versions of Terraform CLI requires the `Msgpack` field to be populated. This will require converting from RawState JSON and flatmap to MessagePack.

### Steps to Reproduce

Using any framework provider and `tf6to5server`:

1. `terraform apply`
2. `terraform apply`

Using https://github.com/hashicorp/terraform-provider-corner/pull/44:

1. `go test ./internal/tf6to5provider`

### References

- https://github.com/hashicorp/terraform-plugin-framework/blob/bf0240fa43e10ac4c3ac454338f75fe928e9f048/tfsdk/serve.go#L493-L497
- #22

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/42
Closes #262
Reference: https://github.com/hashicorp/terraform-provider-corner/pull/44

Terraform CLI version 0.12 through 0.14 support protocol version 5 and require the `UpgradeResourceStateResponse` type `UpgradedState` type `Msgpack` field. To support downgraded framework providers via terraform-plugin-mux `tf6to5server`, the `RawState` type `JSON` field cannot simply be passed through like the previous implementation. This change will parse the `RawState` and output it via the `tfprotov6.NewDynamicValue()` function, which will always set the `Msgpack` field.

Similar logic would have already been required as part of the larger effort to support resource versioning and upgrading resource state over time.

Passthrough of the same state of an invalid resource type should not have been previously allowed, since the provider did not implement the resource type. Regardless, Terraform CLI should not have reached this point since `GetProviderSchema` would not have returned the resource type anyways for other resource operations.

Verified via integration testing of terraform-plugin-mux `tf6to5server`, which previously returned a non-descript `EOF` error.

### DIFF
--- a/.changelog/pending.txt
+++ b/.changelog/pending.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+tfsdk: Support protocol version 5 and verify valid resource type in `UpgradeResourceState` RPC
+```
--- a/tfsdk/serve.go
+++ b/tfsdk/serve.go
@@ -485,16 +485,100 @@ func (s *server) validateResourceConfig(ctx context.Context, req *tfprotov6.Vali
 	resp.Diagnostics = validateSchemaResp.Diagnostics
 }
 
-func (s *server) UpgradeResourceState(ctx context.Context, req *tfprotov6.UpgradeResourceStateRequest) (*tfprotov6.UpgradeResourceStateResponse, error) {
-	// uncomment when we implement this function
-	//ctx = s.registerContext(ctx)
+// upgradeResourceStateResponse is a thin abstraction to allow native
+// Diagnostics usage.
+type upgradeResourceStateResponse struct {
+	Diagnostics   diag.Diagnostics
+	UpgradedState *tfprotov6.DynamicValue
+}
 
-	// TODO: support state upgrades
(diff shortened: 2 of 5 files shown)


detro (approved):
I have to admit, I don't fully follow this code (yet, hopefully!).
But the general idea (provide a controlled, programmatic way to support a provider-supported upgrade) makes perfect sense. And it plugs a "hole" where previously the behaviour was, essentially, undefined.

LGTM

PR #264, opened 2022-03-07
Updating golang doc for `tfsdk.Attribute` `.Computed` and `.PlanModifiers` fields

detro:
This is aimed to help developers what's necessary to implement a "default value" for attributes.

NOTE: I'm not an english-native speaker, so this is just a starting point. I only care we document this, but the words can totally be changed.

### DIFF
--- a/tfsdk/attribute.go
+++ b/tfsdk/attribute.go
@@ -56,6 +56,15 @@ type Attribute struct {
 	// Required and Optional are both false, Computed must be true, and the
 	// attribute will be considered "read only" for the practitioner, with
 	// only the provider able to set its value.
+	//
+	// To define an attribute that is Optional and uses PlanModifiers
+	// to set a "default value" when none is provided by configuration,
+	// Computed must also be set: this is necessary because default
+	// values are, in effect, set by the provider.
+	// Failing to do so, will return the error like:
+	//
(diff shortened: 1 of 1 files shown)


tfsdk/attribute.go
```diff
@@ -49,13 +49,25 @@ type Attribute struct {
...
 	// Computed indicates whether the provider may return its own value for
-	// this attribute or not. Required and Computed cannot both be true. If
+	// this Attribute or not. Required and Computed cannot both be true. If
 	// Required and Optional are both false, Computed must be true, and the
 	// attribute will be considered "read only" for the practitioner, with
 	// only the provider able to set its value.
+	//
+	// As mentioned above for Optional, when defining an Attribute
+	// that has "default value", Computed must also be set to true.
+	// Failing to do so, will return the error like:
```
bflad, line 67: Nit: "Above" only makes sense in the context of reading the code or browsing on pkg.go.dev; for folks coming to this documentation via the Go language server, only this field's documentation is shown. I think we can just drop the beginning part of this sentence.

```suggestion
	// When defining an Optional Attribute that has "default value",
	// Computed must also be set to true. Otherwise, Terraform
	// will return an error like:
```

bflad (approved):
One nit, otherwise great to have this documented. 🚀

PR #228, opened 2021-11-29
docs/design: Upgrade Resource State

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/42

PR #292, opened 2022-04-18
tfsdk: Initial ResourceWithUpgradeState implementation

### LINKED ISSUE #42: Support upgrading resource state

bflad:
Closes #42
Reference: https://github.com/hashicorp/terraform-plugin-framework/pull/228

Support provider defined `UpgradeResourceState` RPC handling, by introducing an optional `ResourceWithUpgradeState` interface type, with an `UpgradeState` method. Each underlying state version upgrade implementation is expected to consume the prior state, perform any necessary data manipulations, then respond with the upgraded state.

This framework implementation differs from the terraform-plugin-sdk implementation:

- State upgraders are specified via a mapping, rather than a slice with underlying version field. This should prevent certain classes of coding issues.
- State upgraders must be wholly contained from the prior state version to the current schema version. The framework does not loop through each successive version because attempting to recreate the `tfprotov6.RawState` for each intermediate version request would be very problematic. For example, terraform-plugin-go does not implement functionality for marshalling a `RawState`. Provider developers can use their own coding techniques to reduce code duplications when multiple versions need the same logic.
- Specifying the full prior schema is now an optional implementation detail. Working with the lower level data types is more challenging, however this has been a repeated feature request.

There are some quirks and potential future enhancements to the framework `UpgradeResourceState` handling:

- Past and current versions Terraform CLI will call `UpgradeResourceState` even if the state version matches the current schema version. This implementation keeps the framework's prior logic to roundtrip the existing state into the upgraded state. It may be possible to stop this Terraform CLI behavior with protocol version 6, although the logic would need to remain for backwards compatibility.
- It may be possible to help provider developers simplify logic by attempting to automatically populate compatible parts of the upgraded state from the prior state. This can potentially be done at a later time.


tfsdk/resource_upgrade_state.go
```diff
@@ -0,0 +1,121 @@
...
+// the UpgradeResourceState RPC. This RPC is automatically called by Terraform
+// when the current Schema type Version field is greater than the stored state.
+// Terraform does not store previous Schema information, so any breaking
+// changes to state data types must be handled by providers.
+//
+// Terraform CLI can execute the UpgradeResourceState RPC even when the prior
+// state version matches the current schema version. The framework will
+// automatically intercept this request and attempt to respond with the
+// existing state. In this situation the framework will not execute any
+// provider defined logic, so declaring it for this version is extraneous.
```
detro, line 20: Probably not strictly caused by FW, but why "...CLI can execute UpgradeResourceState RPC even when prior state verson matches current schema version"?

And with "extraneous" you mean "pointless" or "superfluous", because the FW will intercept it and so, if a provider hasn't changed schema version, it's dead code?

bflad: > CLI can execute UpgradeResourceState RPC even when prior state verson matches current schema version

terraform-plugin-sdk used to perform some flatmap -> JSON handling due to changes in Terraform 0.12. Why this was done on the provider side versus the CLI side, I’m not exactly sure. Alas, we must support this odd case until the next major version of the protocol.

> And with "extraneous" you mean "pointless" or "superfluous", because the FW will intercept it and so, if a provider hasn't changed schema version, it's dead code?

Correct, it'll be dead code, similar to declaring state upgraders for versions above the current schema version. I'm just imagining that this might be more of a common gotcha where provider developers might create a state upgrade specifying the current schema version rather than the prior one. Hopefully some unit testing bits can discover that issue and report back to the provider developer about the issue.

detro (approved):

bflad:
Associated terraform.io update: https://github.com/hashicorp/terraform-website/pull/2255

PR #294, opened 2022-04-19
tfsdk: Introduce new ProviderServer functions, deprecate existing

### LINKED ISSUE #22: Support muxing with SDKv2
### Module version

```
N/A
```

### Use-cases

Mux was born out of the desire to make sure that this new framework could be adopted piecemeal, that moving over wasn't a Big Bang rewrite project. That one resource at a time could be rewritten into or implemented in the new framework, and older SDKv2 code would continue to work.

### Attempted Solutions

The fundamental problem is that SDKv2 uses protocol version 5, and this framework is being built on protocol version 6. On the face of it, that's not that big a deal; hashicorp/terraform-plugin-go-contrib#7 talks about a way to transparently upgrade providers built on version 5 of the protocol to version 6  of the protocol, without any loss of fidelity. So in theory an SDKv2 provider could be just wrapped in that helper, and then muxed with a provider using this framework as two protov6 providers. No big deal.

And this should, in theory, work fine.

As long as the provider configuration (the schema for the provider itself, not the resources) doesn't use any nested blocks or nested attributes.

Because both muxed providers need to respond to the same ConfigureProvider RPC to get their configuration, they need to agree on what the schema for that call is. SDKv2 is on protocol version 5, which has no concept of nested attributes. This framework is choosing not to expose sub-blocks, because we don't think they're needed anymore and are confusing for provider developers, so we're steering people away from them. This means there exist a great number of provider configuration schemas that can't be modeled by both SDKv2 and the framework, meaning those providers would not be able to be muxed.

We have a few options:

1. We could relax the schema requirement; it's artificially imposed in mux. We _could_ say "nested block and nested attribute are equivalent" and then just paper over how they're presented. This seems tricky, sub-optimal, and like it could have a bunch of bugs lurking.
2. We could expose nested attributes in SDKv2. That would require it to move to protocol version 6, meaning it would only be compatible with Terraform 0.15 and later. I don't think provider developers would like that.
3. We could expose nested blocks in this framework as part of the schema work in #12 /  #11. This would be a shame, because it would carry forward a concept that we're trying to bury and guide people away from, and it may take us a while to get rid of it. It also lets them use blocks _anywhere_, not just in the provider schema, where they're needed.
4. We could add a special, optional callback (`GetMuxedProviderSchema` or something) to the provider that allowed it to return a schema of a different type, with support for nested blocks. This would limit the impact to only the area needed for compatibility, and would let providers who don't need that compatibility (new providers, completely rewritten providers, etc.) avoid the cognitive overhead entirely.

### Proposal

I think my personal preference is number 4, but I think that's a lightly held preference at this point. I think either 3 or 4 is the way to solve this, honestly.

bflad:
Closes #22 (its technically solved already, this is just cherry on top)

Provider developers have a few reasons to directly need a `tfprotov6.ProviderServer` implementation:

- When implementing acceptance testing with terraform-plugin-sdk helper/resource.
- When using terraform-plugin-mux tf6muxserver.NewMuxServer().
- When using terraform-plugin-mux tf6to5server.DowngradeServer().

The current `NewProtocol6Server(Provider)` function enables these use cases, however the implementation is less than ergonomic for provider developers as a wrapping function is required in all cases:

```go
// helper/resource acceptance testing

resource.TestCase{
	ProtoV6ProviderFactories: map[string]func() (tfprotov6.ProviderServer, error){
		"example": func() (tfprotov6.ProviderServer, error) {
			return tfsdk.NewProtocol6Server(New()), nil
		},
	},
 // ...
}

// tf6muxserver.NewMuxServer()

providers := []func() tfprotov6.ProviderServer{
	func() tfprotov6.ProviderServer {
		return tfsdk.NewProtocol6Server(provider1.New())
	},
	func() tfprotov6.ProviderServer {
		return tfsdk.NewProtocol6Server(provider2.New())
	},
}

muxServer, err := tf6muxserver.NewMuxServer(ctx, providers...)

// tf6to5server.DowngradeServer()

downgradeServer, err := tf6to5server.DowngradeServer(ctx, func() tfprotov6.ProviderServer {
	return tfsdk.NewProtocol6Server(provider.New())
})
```

The new functions simplify these implementations and get provider developers closer to not directly importing terraform-plugin-go (with a few other targeted changes in other projects, such as type aliases), e.g.

```go
// helper/resource acceptance testing

resource.TestCase{
	ProtoV6ProviderFactories: map[string]func() (tfprotov6.ProviderServer, error){
		"example": tfsdk.NewProtocol6ProviderServerWithError(New()),
	},
 // ...
}

// tf6muxserver.NewMuxServer()

providers := []func() tfprotov6.ProviderServer{
	tfsdk.NewProtocol6ProviderServer(provider1.New()),
  tfsdk.NewProtocol6ProviderServer(provider2.New()),
}

muxServer, err := tf6muxserver.NewMuxServer(ctx, providers...)

// tf6to5server.DowngradeServer()

downgradeServer, err := tf6to5server.DowngradeServer(ctx, tfsdk.NewProtocol6ProviderServer(provider.New()))
```

This change prefers deprecation over straight replacement to give some lead time for various documentation updates across multiple projects.

The naming is less than ideal, however it feels necessary to remain generic instead of picking any particular details related to their current usage (e.g. "NewTestServer") as this can/will change over time. The "Protocol6" naming is important should a new major protocol version 7 be released, which this framework must support.

### DIFF
--- a/tfsdk/serve.go
+++ b/tfsdk/serve.go
@@ -38,10 +38,37 @@ type ServeOpts struct {
 
 // NewProtocol6Server returns a tfprotov6.ProviderServer implementation based
 // on the passed Provider implementation.
+//
+// Deprecated: Use NewProtocol6ProviderServer instead. This will be removed
+// in an upcoming minor version before v1.0.0.
 func NewProtocol6Server(p Provider) tfprotov6.ProviderServer {
 	return &server{
 		p: p,
 	}
+}
+
+// Returns a protocol version 6 ProviderServer implementation based on the
+// given Provider and suitable for usage with the terraform-plugin-go
+// tf6server.Serve() function and various terraform-plugin-mux functions.
+func NewProtocol6ProviderServer(p Provider) func() tfprotov6.ProviderServer {
+	return func() tfprotov6.ProviderServer {
+		return &server{
+			p: p,
+		}
+	}
+}
+
+// Returns a protocol version 6 ProviderServer implementation based on the
+// given Provider and suitable for usage with the terraform-plugin-sdk
+// acceptance testing helper/resource.TestCase.ProtoV6ProviderFactories.
+//
+// The error return is not currently used, but it may be in the future.
+func NewProtocol6ProviderServerWithError(p Provider) (func() tfprotov6.ProviderServer, error) {
+	return func() tfprotov6.ProviderServer {
+		return &server{
+			p: p,
+		}
+	}, nil
 }
 
 // Serve serves a provider, blocking until the context is canceled.
--- a/.changelog/pending.txt
+++ b/.changelog/pending.txt
@@ -1,0 +1,11 @@
+```release-note:note
+tfsdk: The `NewProtocol6Server()` function has been deprecated in preference of `NewProtocol6ProviderServer()` and `NewProtocol6ProviderServerWithError()` functions, which will simplify muxing and testing implementations. The `NewProtocol6Server()` function will be removed in an upcoming minor release before v1.0.0.
+```
+
+```release-note:enhancement
+tfsdk: Added `NewProtocol6ProviderServer()` function, which can be directly used with terraform-plugin-go and terraform-plugin-mux.
+```
+
+```release-note:enhancement
+tfsdk: Added `NewProtocol6ProviderServerWithError()` function, which can be directly used with terraform-plugin-sdk acceptance testing.
+```
--- a/tfsdk/serve_test.go
+++ b/tfsdk/serve_test.go
@@ -16,6 +16,39 @@ import (
 	"github.com/hashicorp/terraform-plugin-go/tftypes"
 )
 
+func TestNewProtocol6ProviderServer(t *testing.T) {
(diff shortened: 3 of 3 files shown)


tfsdk/serve.go
```diff
@@ -38,12 +38,39 @@ type ServeOpts struct {
 
 // NewProtocol6Server returns a tfprotov6.ProviderServer implementation based
 // on the passed Provider implementation.
+//
+// Deprecated: Use NewProtocol6ProviderServer instead. This will be removed
+// in an upcoming minor version before v1.0.0.
```
detro, line 43: I was going to ask if we really need to do this deprecation phase, but I did a quick GitHub global code search, and while it's in the order of ~70 usages, and it includes the HashiCorp repos, it's still 70 usages.

Yep, deprecation first it is.

detro (approved):
LGTM

bflad:
Associated terraform.io update: https://github.com/hashicorp/terraform-website/pull/2256

Issue #160, opened 2021-09-14
Resource Import Documentation Handling


bflad:
### Module version

```
v0.4.0 (not yet released)
```

### Use-cases

Provider developers will want to document whether a resource supports import and what type of import identifier(s) are supported. Tooling such as [`terraform-plugin-docs`](https://github.com/hashicorp/terraform-plugin-docs) (or similar) should be able to determine the implementation and automatically generate import documentation.

For example, most simple resources follow this simple type of documentation pattern for "passthrough" import:

- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/instance#import
- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/rds_cluster#import
- https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/windows_virtual_machine#import
- https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/mssql_database#import

Some resources may need to document more complicated cases, such as "complex" identifiers that get split into multiple attributes:

- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_role_policy#import
- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_provisioned_concurrency_config#import

The most advanced cases require documenting multiple example import identifiers (sometimes to make the process easier for practitioners, other times to workaround incomplete Read APIs):

- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route53_record#import
- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route53_zone_association#import
- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/security_group_rule#import
- https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/cloudfunctions_function#import
- https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_instance#import

### Attempted Solutions

Version 0.4.0 of the framework supports resource import via a _required_ `ImportState` method on the `Resource` type. This makes it difficult for tooling to determine if import is supported or not as it is reliant on specific implementation details, which may or may not be consistently implemented (such as the recommended `ResourceImportStateNotImplemented()` helper).

There is no current methodology for tooling to know import examples, except by describing it manually in resource descriptions.

### Proposal

Consider migrating the `ImportState` method onto its own interface type, e.g.

```go
type ResourceWithImportState interface {
[...]

PR #297, opened 2022-04-21
tfsdk: Move Resource ImportState method to optional ResourceWithImportState interface

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/160

Due to provider developer feedback, it has been suggested to make the current `Resource` interface `ImportState` method optional. This change accomplishes that by moving the existing method signature to a new `ResourceWithImportState` interface. This has the add-on effect of allowing future documentation generation easily determine whether resource import support is implemented, by type checking against the new interface.

This also deprecates the `ResourceImportStateNotImplemented()` function. Providers can now signal that a resource does not support import by omitting the `ImportState` method and the framework will automatically return an error diagnostic. From a quick GitHub search, it appeared there was only one usage of a custom error message outside the framework testing. However, it was only being used to include the resource type in the message and was of no real utility over the generic messaging. A code comment is left with a suggested implementation, should there be a feature request for customized messaging.


tfsdk/serve_import.go
```diff
@@ -85,6 +86,26 @@ func (s *server) importResourceState(ctx context.Context, req *tfprotov6.ImportR
...
+		// if the resource implements the ResourceWithImportState interface.
+		// Instead, a separate "ResourceWithoutImportState" interface could be
+		// created with a method such as:
+		//    ImportNotImplementedMessage(context.Context) string.
+		resp.Diagnostics.AddError(
+			"Resource Import Not Implemented",
+			"This resource does not support import. Please contact the provider developer for additional information.",
+		)
+		return
+	}
```
detro, line 107: 👨‍🍳 💋

detro (approved):
This change makes for a neat cleanup.

tfsdk/resource_import.go
```diff
@@ -6,10 +6,26 @@ import (
 	"github.com/hashicorp/terraform-plugin-go/tftypes"
 )
 
+// Optional interface on top of Resource that enables provider control over
+// the ImportResourceState RPC. This RPC is called by Terraform when the
+// `terraform import` command is executed. Afterwards, the ReadResource RPC
+// is executed to allow providers to fully populate the resource state.
+type ResourceWithImportState interface {
```
tombuildsstuff, line 13: perhaps nit-picking, but shouldn't this interface contain `Resource`, which would allow for validation that an implementation is fully implementing both using a single line? e.g.

```
var _ ResourceWithImportState = MyResource{}
# which also implies
# var _ Resource = MyResource{}
```

PR #300, opened 2022-04-22
tfsdk: Added logging for all framework handoffs to provider defined logic

bflad:
This will help practitioners and provider developers further understand when and where logic handoffs occur. In general, the framework and underlying terraform-plugin-go use the `DEBUG` log level for project code boundaries and major logic milestones while the `TRACE` log level is used for more lower level logic details.


detro (approved):
I read your description about why using `DEBUG` for these new logs instead of `TRACE`: the reasoning makes sense, but I'm concerned how much more logging this is going to create.

Yes, the fact that user can filter by sub-system seems the way out of it, but... it's still a lot more logging.

I can't raise an argument against it more than "are you sure?", because codewise this LGTM.

PR #301, opened 2022-04-22
tfsdk: Added automatic RemoveResource() call after Resource type Delete method execution

### LINKED ISSUE #100: Consider Automatically Calling State.RemoveResource() After Successful Resource Deletion
### Module version

```
v0.2.0
```

### Use-cases

Currently the framework requires all resource implementations to include this type of logic in the `Delete()` method:

```go
func (r exampleResource) Delete(ctx context.Context, req tfsdk.DeleteResourceRequest, resp *tfsdk.DeleteResourceResponse) {
	// logic to delete resource

	resp.State.RemoveResource(ctx)
}
```

Is there ever a time this would not be desired on successful deletion? Maybe we can save a few keystrokes.

### Proposal

Implement `State.RemoveResource(ctx)` in the calling code on `Delete()` success so it is not necessary for all implementations. From the previous framework:

https://github.com/hashicorp/terraform-plugin-sdk/blob/c455248338544ae42b6449eae44279c0c82d7987/helper/schema/resource.go#L428-L434

bflad:
Closes #100

Due to provider developer feedback, the framework will now perform automatic removal of state on successful (no error diagnostic) `Resource` type `Delete` executions. This matches the behavior of terraform-plugin-sdk.

Providers can still explicitly call `(State).RemoveResource()` or leave existing implementations, however it is extraneous for most `Delete` logic now.


bookshelfdave (approved):
that was quick :-) lgtm!

PR #303, opened 2022-04-24
correct NewProtocol6ProviderServerWithError to match the sdkv2 defini…

iwarapter:
…tion for ProtoV6ProviderFactories

The SDKv2 `TestCase` `ProtoV6ProviderFactories` expects a slightly different function signature, which results in test provider definitions like:

```
"example": func() (tfprotov6.ProviderServer, error) {
		srv, err := tfsdk.NewProtocol6ProviderServerWithError(New())
		return srv(), err
	},
```

https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/resource/testing.go#L343-L348

This change aligns the function signature to work as below:

```
	"example": tfsdk.NewProtocol6ProviderServerWithError(New("test")),
```

I have done some simple integration tests with the corner provider.

### DIFF
--- a/.changelog/303.txt
+++ b/.changelog/303.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+tfsdk: Correct `NewProtocol6ProviderServerWithError` function signature to align to sdk v2 `TestCase.ProtoV6ProviderFactories`
+```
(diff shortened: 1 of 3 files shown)


bflad (approved):
Hey @iwarapter 👋  Thanks for catching this. That was indeed the intention here. Since this hasn't been released yet, I'm going to drop the extra changelog entry so it doesn't confuse anyone, but I really appreciate you catching this and fixing it before it was released.

Aside: I'm not sure yet, but as part of #215, we may actually move these to a separate package (along with `Serve()` and `ServeOpts`), so we can help reduce the size of `tfsdk` and clarify the abstractions better as we move closer to a v1.0.0.

PR #308, opened 2022-04-27
Introduce providerserver package, deprecate tfsdk server functionality

bflad:
Continues: https://github.com/hashicorp/terraform-plugin-framework/pull/294
Continues: https://github.com/hashicorp/terraform-plugin-framework/pull/296

This change represents the first half of the work necessary to extract the `tfprotov6.ProviderServer` implementation and helper functions out of the `tfsdk` package and into separate packages. The `providerserver` package will be the provider developer facing functionality, while the next iteration of this refactoring will move the actual server implementation into a separate internal package. Once in that separate internal package, efforts can be made to make that code handle terraform-plugin-go type conversions "at the edge" better.


tfsdk/serve.go
```diff
@@ -15,50 +15,34 @@ import (
...
 func NewProtocol6Server(p Provider) tfprotov6.ProviderServer {
-	return &server{
-		p: p,
-	}
-}
-
-// Returns a protocol version 6 ProviderServer implementation based on the
-// given Provider and suitable for usage with the terraform-plugin-go
-// tf6server.Serve() function and various terraform-plugin-mux functions.
-func NewProtocol6ProviderServer(p Provider) func() tfprotov6.ProviderServer {
```
bflad, line 40: This is safe, as this function was not released as part of this package yet.

tfsdk/serve.go
```diff
@@ -15,50 +15,34 @@ import (
...
-		}
-	}
-}
-
-// Returns a protocol version 6 ProviderServer implementation based on the
-// given Provider and suitable for usage with the terraform-plugin-sdk
-// acceptance testing helper/resource.TestCase.ProtoV6ProviderFactories.
-//
-// The error return is not currently used, but it may be in the future.
-func NewProtocol6ProviderServerWithError(p Provider) func() (tfprotov6.ProviderServer, error) {
```
bflad, line 53: This is safe, as this function was not released as part of this package yet.

providerserver/providerserver.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
+	"github.com/hashicorp/terraform-plugin-go/tfprotov6"
+	"github.com/hashicorp/terraform-plugin-go/tfprotov6/tf6server"
+)
+
+// Returns a protocol version 6 ProviderServer implementation based on the
```
bendbennett, line 12: Perhaps prefix with `NewProtocol6 `?

bendbennett (approved):

providerserver/providerserver.go
bflad: Updated in [81cba2c](https://github.com/hashicorp/terraform-plugin-framework/pull/308/commits/81cba2c2c215b865eeafd7f013ebed402c5b05a5) -- we should figure out what linter is triggering the warnings so we can either explicitly enable or disable it per team preference.

PR #316, opened 2022-05-04
internal: Introduce logging unit testing

bflad:
Creates `internal/testing/emptyprovider` package so it can be shared for multiple packages, then uses it to verify the logging setup with `(internal/proto6server.Server).GetProviderSchema()`. This could be expanded to other RPC handlers, if desired and the testing setup is trivial enough. Also added unit testing within internal/logging to ensure it is fully covered.


bendbennett (approved):
LGTM

PR #319, opened 2022-05-06
Introduce internal/fwserver package and migrate GetProviderSchema

bflad:

This step of the provider server refactoring introduces a new `internal/fwserver` package that will contain the provider server implementation for any protocol version, there only being a version 6 implementation at the moment. This package will contain only framework-native types that aren't already defined by the `tfsdk` package that are used by provider developers today, which is necessary to prevent import cycles. The protocol specific implementations will then wrap this framework server implementation and handle any necessary conversion of protocol specific types to framework specific types and back. Eventually, the existing `internal/proto6server` implementation will make no reference to any framework native types except the framework server and protocol type conversions.

For now, only the `GetProviderSchema` RPC is migrated to show what this refactoring will look like for the rest of the RPCs.

This has a few benefits which can be seen here including clear abstractions for protocol specific handling versus generic framework handling and unit testing at those abstraction boundaries.

_Please note that much of the "end-to-end" unit testing is already in the `internal/proto6server` package, which shows no changes in behavior, and therefore can help gain confidence in this change not introducing potential regressions._

### DIFF
--- a/internal/fromproto6/doc.go
+++ b/internal/fromproto6/doc.go
@@ -1,0 +1,3 @@
+// Package fromproto6 contains functions to convert from protocol version 6
+// (tfprotov6) types to framework types.
+package fromproto6
--- a/internal/fromproto6/getproviderschema.go
+++ b/internal/fromproto6/getproviderschema.go
@@ -1,0 +1,20 @@
+package fromproto6
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-framework/internal/fwserver"
+	"github.com/hashicorp/terraform-plugin-go/tfprotov6"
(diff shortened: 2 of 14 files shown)


bendbennett (approved):
I really like the direction this is going in, a distinct separation of concerns for protocol-specific and framework-specific handling is a great idea and makes logical sense. I have to be honest though and confess that I didn’t look at [internal/toproto6/getproviderschema_test.go](https://github.com/hashicorp/terraform-plugin-framework/pull/319/files#diff-ed38ae77bb15681d5823546305836e5e315f32fe05d7489ae79b1bdfaafd9a6c). Great work Brian!

Issue #299, opened 2022-04-22
tfsdk: Consider Caching Provider Defined Resources/Data Sources in Server


bflad:
### Module version

```
v0.6.1
```

### Use-cases

The current provider server logic always calls back into the provider defined `GetResources()` and `GetDataSources()` methods when trying to find each `ResourceType` or `DataSourceType` associated with every RPC. While not a biggest performance deal, it does mean that it is repeatedly fetching the maps and executing any provider logic (if any). Terraform CLI and the framework currently expect the type mapping to be consistent for providers after `GetProviderSchema`. The server could cache these maps on first use.

### Proposal

- Add `dataSourceTypes map[string]DataSourceType` and `resourceTypes map[string]ResourceType` fields to the `server` type
- In `(*server).GetProviderSchema()`, set `s.dataSourceTypes = dataSourceSchemas` and `s.resourceTypes = resourceSchemas`
- Update `(*server).getDataSourceType()` and `(*server).getResourceType()` methods to check `s.dataSourceTypes` and `s.resourceTypes` fields instead of calling the provider defined `GetDataSources()` and `GetResources()` methods

bflad:
Similar caching could also be done for each DataSourceType and ResourceType GetSchema.

bflad:
(Commit message wording accidentally closed this)

PR #320, opened 2022-05-09
internal: Migrate ValidateProviderConfig RPC to fromproto6, fwserver, and toproto6

bflad:

One particular quirk that this implementation shows is that it is not entirely possible to avoid working with `tfsdk`/`fwserver` bits in `proto6server` as the current data types within the framework (`Config`, `Plan`, `State`) require the `Schema` so data can be appropriately converted from the protocol to the framework. To support this effort and prevent repetitively executing the provider defined `GetSchema` method, a `fwserver` method is introduced that caches the `Schema` and `Diagnostics` results on first use. This methodology can also be applied in the future to resolve the existing https://github.com/hashicorp/terraform-plugin-framework/issues/299.


internal/fwserver/server.go
```diff
@@ -9,4 +14,41 @@ import (
...
+
+// ProviderSchema returns the Schema associated with the Provider. The Schema
+// and Diagnostics are cached on first use.
+func (s *Server) ProviderSchema(ctx context.Context) (*tfsdk.Schema, diag.Diagnostics) {
+	logging.FrameworkTrace(ctx, "Checking ProviderSchema lock")
+	s.providerSchemaMutex.Lock()
+
+	if s.providerSchema != nil {
+		s.providerSchemaMutex.Unlock()
+		return s.providerSchema, nil
```
detro, line 41: Should this return also `s.ProviderSchemaDiags`?

Unless the intention is to report potential diag-reported warnings/errors, only the first time?

internal/fwserver/server.go
```diff
@@ -9,4 +14,41 @@ import (
...
+}
+
+// ProviderSchema returns the Schema associated with the Provider. The Schema
+// and Diagnostics are cached on first use.
+func (s *Server) ProviderSchema(ctx context.Context) (*tfsdk.Schema, diag.Diagnostics) {
+	logging.FrameworkTrace(ctx, "Checking ProviderSchema lock")
+	s.providerSchemaMutex.Lock()
+
+	if s.providerSchema != nil {
+		s.providerSchemaMutex.Unlock()
```
detro, line 40: To ensure future code doesn't break the needs of the mutex to be unlocked, I have seen a slightly different usage pattern (that I like). Something like:

```
s.ProviderSchemaMutex.Lock()
defer s.ProviderSchemaMutex.Unlock()
...
```

This way, whatever branching of the logic we take (the memoized one or not), the locking bit is handled safely and consistently.

detro (approved):
Just a suggestion around the way the mutex gets unlocked.

Mostly is moving things to the new structure, so LGTM.

internal/fwserver/server.go
bflad: Thank you for this suggestion, it's a great one and one I'll implement!

internal/fwserver/server.go
bflad: Yes, it should, thanks for noticing!

PR #324, opened 2022-05-11
internal: Migrate ValidateDataSourceConfig RPC to fromproto6, fwserver, and toproto6

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/299

This introduces framework server caching for data source types and schemas, similar to the provider schema cache, which is needed with other data source RPCs. The same methodology will be applied for resource types and schemas, when the `ValidateResourceConfig` RPC is migrated.


bendbennett (approved):
LGTM

Issue #329, opened 2022-05-16
Thoughts about block deprecation


lemaitre:
Hi everyone.

I just want to add some food for thought on the subject of blocks. One aspect of blocks that I think have been missed on those discussion is how much more declarative they are. Indeed, Blocks of different types can be interleaved without altering the semantics.
For example, adding a port forwarding to a container does not need to be tied to the other ports forwarded on this container.

I will take a provider that I started to write using the plugin framework : a provider to create custom resources from shell commands.

Here is what I currently have:

```terraform
resource "cmd_local" "foo" {
  inputs = {
    a = /* ... */
    b = /* ... */
  }
  create {
    cmd = <<-EOT
      echo "$INPUT_a" > a
      echo "$INPUT_b" > b
    EOT
  }
  destroy {
    cmd = <<-EOT
      rm a b
    EOT
  }
  
  # a variable
  reload {
    name = "a"
    cmd = "cat a"
  }
  update {
    triggers = ["a"]
    cmd = <<-EOT
      echo "$INPUT_a" > a
    EOT
  }

  # b variable
  reload {
    name = "b"
    cmd = "cat b"
  }
  update {
    triggers = ["b"]
    cmd = <<-EOT
      echo "$INPUT_b" > b
    EOT
  }

  # update fallback
  update {
    triggers = ["a", "b"]
    cmd = <<-EOT
      echo "$INPUT_a" > a
      echo "$INPUT_b" > b
    EOT
  }
}
```

With blocks, I can nicely group those however I want (here, I have grouped together how `b` is reloaded, and what update to do when `b` is updated. For those arguing that I should just make my data model grouping those directly, then I would argue that an update is not tied to a single variable, therefore, there is no nice way to make it naturally grouping without making cumbersome.
By the way, this example would even be simpler with map blocks as I could make the `reload` block a map.

bflad:
The `Schema` type `Blocks` field Go documentation deprecation has been removed, which will reflect in the next release. If you have additional expectations from raising this issue, please reach out and thank you for submitting this.

lemaitre:
Hi @bflad, 

> If the ask is to remove the "Deprecated:" comment on the `Block` Go types, then you will be happy to know that this will likely be done shortly.

Yes, that was my main point, and I'm glad to hear that.

> > By the way, this example would even be simpler with map blocks
> 
> We explicitly are not enabling map block support at the moment because it is introduces some interesting problems

[...]

PR #330, opened 2022-05-17
tfsdk: Remove Go documentation deprecation for Schema type Blocks field

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/329

While attributes will be recommended over blocks going forward, we do not want to alarm or cause confusion with provider developers migrating existing schemas from terraform-plugin-sdk. Terraform also does not provide native functionality to help practitioners switch syntax in configurations, so this deprecation notice is considered premature.


bendbennett (approved):
LGTM

detro (approved):
:)

PR #331, opened 2022-05-19
internal: Migrate ValidateResourceConfig RPC to fromproto6, fwserver, and toproto6

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/299

This migration is almost identical to what was done for `ValidateDataSourceConfig`. This introduces framework server caching for resource types and schemas, similar to the data source schema, data source type, and provider schema caches, which is needed with other resource RPCs.


bendbennett (approved):
LGTM

PR #333, opened 2022-05-19
internal/proto6server: Always use data source and resource caches, remove intermediate Provider field

### LINKED ISSUE #299: tfsdk: Consider Caching Provider Defined Resources/Data Sources in Server

bflad:
Closes #299

This completes the migration to the cached data source and resource schema/type handling in the new `internal/fwserver` package. This also removes the previous `Provider` field so that implementation detail is now wholly within the `internal/fwserver` to remove any confusion.

### DIFF
--- a/internal/proto6server/serve.go
+++ b/internal/proto6server/serve.go
@@ -24,8 +24,6 @@ var _ tfprotov6.ProviderServer = &Server{}
 type Server struct {
 	FrameworkServer fwserver.Server
 
-	// Provider will be migrated to FrameworkServer over time.
-	Provider         tfsdk.Provider
 	contextCancels   []context.CancelFunc
 	contextCancelsMu sync.Mutex
 }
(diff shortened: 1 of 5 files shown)


bendbennett (approved):
LGTM

PR #335, opened 2022-05-19
Proposal: adding `IsNull()` and `IsUnknown()` methods to `attr.Value` interface

detro:
_**NOTE:** This comes after a verbal conversation about this._

_**NOTE:** This is a draft, to check a few things about the approach I'm taking. It's only applied to `Bool`, so it's not the final PR._

## Background

As part of the work to port https://github.com/hashicorp/terraform-provider-tls/ to `terraform-plugin-framework`, a need arises as soon as I attempted to implement `AttributeValidator` and `AttributePlanModifier`.

### About plan modifiers

For plan modifiers, it's possible to implement a `defaultValueAttributePlanModifier` that when used looks like:
```golang
  defaultValue(types.Bool{Value: false}),
},
```

And the implementation looks like:
```golang
type defaultValueAttributePlanModifier struct {
	DefaultValue attr.Value
}

var _ tfsdk.AttributePlanModifier = (*defaultValueAttributePlanModifier)(nil)

func defaultValue(v attr.Value) tfsdk.AttributePlanModifier {
	return &defaultValueAttributePlanModifier{v}
}

func (apm defaultValueAttributePlanModifier) Modify(ctx context.Context, req tfsdk.ModifyAttributePlanRequest, res *tfsdk.ModifyAttributePlanResponse) {
	// First, check the attribute configuration
	attrConfigValue, err := req.AttributeConfig.ToTerraformValue(ctx)
	if err != nil {
		res.Diagnostics.AddAttributeError(
			req.AttributePath,
			fmt.Sprintf("Unexpected error converting config value for %q", pathToString(req.AttributePath)),
			fmt.Sprintf("This is always a bug with the provider. Error: %s", err),
		)
		return
	}

	// If the attribute configuration is not null, we are done here
	if !attrConfigValue.IsNull() {
		return
	}

	// Second, check the attribute plan (so far):
	// this is necessary as other plan modifier might have already been applied, and we don't
	// want to interfere.
	attrPlanValue, err := req.AttributePlan.ToTerraformValue(ctx)
	if err != nil {
		res.Diagnostics.AddAttributeError(
			req.AttributePath,
			fmt.Sprintf("Unexpected error converting plan value for %q", pathToString(req.AttributePath)),
			fmt.Sprintf("This is always a bug with the provider. Error: %s", err),
		)
		return
	}

	// If the attribute plan is not null, or is fully known (objects), we are done here
	if !attrPlanValue.IsNull() || attrPlanValue.IsFullyKnown() {
		return
	}

	res.AttributePlan = apm.DefaultValue
}
```

If you look closely, we are having to convert back to TF types to check for null-ness. That feels sub-optimal (_code smell?_).

### About validators

Similarly to the above, I wanted to implement `AttributeValidator` that can be used to verify that an attribute was/wasn't set, in respect to other fields. This is a very useful declarative feature of `terraform-plugin-sdk/v2`:
```golang
type Schema struct {
  // ...
  ConflictsWith [][string](https://pkg.go.dev/builtin#string)
  ExactlyOneOf  [][string](https://pkg.go.dev/builtin#string)
  AtLeastOneOf  [][string](https://pkg.go.dev/builtin#string)
  RequiredWith  [][string](https://pkg.go.dev/builtin#string)
  // ...
}
```

This can be achieved in the framework by leveraging `tfsdk.ValidateAttributeReques.Config`, that allows to interrogate the entire configuration while validating a single attribute.

For example, this is how the equivalent of `RequiredWith` looks like:
```golang
type requiredWithAttributeValidator struct {
	requiredAttributes []*tftypes.AttributePath
}

var _ tfsdk.AttributeValidator = (*requiredWithAttributeValidator)(nil)

func requiredWith(attributePaths ...*tftypes.AttributePath) tfsdk.AttributeValidator {
	return &requiredWithAttributeValidator{attributePaths}
}

func (av requiredWithAttributeValidator) Validate(ctx context.Context, req tfsdk.ValidateAttributeRequest, res *tfsdk.ValidateAttributeResponse) {
	tflog.Debug(ctx, "Validating attribute is set together with other required attributes", map[string]interface{}{
		"attribute":          pathToString(req.AttributePath),
		"requiredAttributes": av.requiredAttributes,
	})

	var v attr.Value
	diags := tfsdk.ValueAs(ctx, req.AttributeConfig, &v)
	if diags.HasError() {
		res.Diagnostics.Append(diags...)
		return
	}

	for _, path := range av.requiredAttributes {
		var o attr.Value
		req.Config.GetAttribute(ctx, path, &o)

		isVNull, err := isAttrValueNull(ctx, v)
		if err != nil {
			res.Diagnostics.AddAttributeError(
				req.AttributePath,
				fmt.Sprintf("Unexpected error converting attribute value for %q", pathToString(req.AttributePath)),
				fmt.Sprintf("This is always a bug with the provider. Error: %s", err),
			)
		}

		isONull, err := isAttrValueNull(ctx, o)
		if err != nil {
			res.Diagnostics.AddAttributeError(
				path,
				fmt.Sprintf("Unexpected error converting attribute value for %q", pathToString(path)),
				fmt.Sprintf("This is always a bug with the provider. Error: %s", err),
			)
		}

		if !isVNull && isONull {
			res.Diagnostics.AddAttributeError(
				req.AttributePath,
				fmt.Sprintf("Attribute %q missing", pathToString(path)),
				fmt.Sprintf("%q must be specified when %q is specified", pathToString(path), pathToString(req.AttributePath)),
			)
			return
		}
	}
}
```

Unfortunately, like for plan modifiers, the implementations needs to convert back to TF types to interrogate null-ness. Writing `switch` blocks and use type assertions could work, but it would then break for custom types, that is a feature of the new framework.

## Proposal

`attr.Value`, the interface implemented by every `types.*`, should contain extra methods:

```golang
type Value interface {
  //...
  // IsNull returns true if the Value is not set, or is explicitly set to null.
  IsNull() bool

  // IsUnknown returns true if the value is not yet known.
  IsUnknown() bool
}
```

This would allow implementing key features like the above, way simpler/cleaner.

## Future considerations

Follows a list of things we will not be addressing in this proposal PR, following the initial team consultation.

### Un-export `Unknown` and `Null` fields on concrete types

It's a bit awkward to have on the concrete types both `.Unknown` fields, as well as `.IsUnknown()` methods exported. We should probably look to ways to move the boolean field away from the public interface of those types. Maybe `.unknown` and `.null`?

### Constructor helpers

By virtue of the previous point, it would become necessary to provide some helper constructors to be able to create concrete types that set `.unknown` and `.null` for us.
For the `types.Bool` type, they could maybe look like:

```golang
func NewBool(v bool) Bool {}

func UnknownBool() Bool {}

func NullBool() Bool {}
```

### `tftypes.Value.IsFullyKnown()`

TF types offer an additional method called `IsFullyKnown()`: this is the same as `IsKnown()` for _primitive types_, but it's useful for structured types like `Objects`, `Maps` etc. I could be beneficial for our structured `types.*`?


types/bool.go
```diff
@@ -30,29 +31,44 @@ var _ attr.Value = Bool{}
...
+	null bool
 
-	// Value contains the set value, as long as Unknown and Null are both
+	// Value contains the set value, as long as unknown and null are both
 	// false.
 	Value bool
 }
 
+// NewBool returns a Bool initialized with the given value.
+func NewBool(value bool) Bool {
```
bendbennett, line 47: Nit: In terms of naming, do we want to consider `NewBoolUnknown()`, `NewBoolNull()` or `BoolUnknown()`, `BoolNull()`?

bflad: I think we should treat any of these potential helpers as a separate proposal issue, so we can hash through the details separate from potentially changing the `attr.Value` interface.

attr/value.go
```diff
@@ -13,11 +13,16 @@ type Value interface {
...
-	// a tftypes.Value.
+	// ToTerraformValue returns the data contained in the Value as a tftypes.Value.
 	ToTerraformValue(context.Context) (tftypes.Value, error)
 
 	// Equal must return true if the Value is considered semantically equal
 	// to the Value passed as an argument.
 	Equal(Value) bool
+
+	// Unknown returns true if the Value is not known.
+	Unknown() bool
```
bflad, line 24: I'd recommend naming these with an `Is` prefix, so the existing fields can remain (at least for now).

types/bool.go
detro: Yeah, the need arised primarily from the fact that I was renaming `Unknown` to `unknown` etc.

I can punt on that.

attr/value.go
detro: OK.

But follow up: in regards to `IsFullyKnown` and `IsKnown` on TF types, they use the _positive_ connotation. While the FW types use the negative `Unknown`.

While it's of course trivial to obtain the opposite for `Unknown / Known`, it's not the same for `FullyKnown / FullyUnknown`? Because something that is _not_ fully known, can be fully or partially unknown.

I wouldn't care at all about `FullyKnown`-ness, but we do have the type `Object` to deal with. What's your take?

bflad: We can add fully known handling later, if it is necessary. 😄  For now, I think its decent to only worry about directly checking a specific value.

detro: Even by adding it later, I think it's important to decide what we do about `Unknown vs Known`. Because if we add it later and we stick for now with `Unknown`, we would end up with `Unknown()` and `FullyKnown()`? Or with `Unknown()` and `FullyUnknown()`?

I appreciate that this might seem superfluous right now, but around future-proofing naming I get jittery.

.gitignore
```diff
@@ -1,2 +1,7 @@
+# Website generation artifacts
 node_modules
-website-preview
\ No newline at end of file
+website-preview
+
+# Jetbrains IDEs
+.idea/
+*.iws
```
detro, line 7: 🙏 I just need to stop worrying about me fat-fingering it and open a PR with a bunch of Jetbrains files in it.

internal/testing/types/bool.go
```diff
@@ -4,13 +4,15 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
```
bflad, line 7: Do you happen to have `goimports` enabled in your editor on save? It should revert this to the way it was.

internal/testing/types/bool.go
```diff
@@ -4,13 +4,15 @@ import (
...
+
 	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/types"
-	"github.com/hashicorp/terraform-plugin-go/tftypes"
 )
 
 var (
-	_ attr.Type = BoolType{}
+	_ attr.Type  = BoolType{}
+	_ attr.Value = Bool{}
```
bflad, line 15: Thank you for adding these. 👍

bflad (approved):
Looks good to me 🚀 

Let's ensure there is an associated changelog entry, e.g. `.changelog/335.txt`

``````markdown
```release-note:breaking-change
attr: The `Value` interface now requires the `IsNull()` and `IsUnknown()` methods
```

```release-note:enhancement
types: Added `IsNull()` and `IsUnknown()` methods to all types
```
``````

bendbennett (approved):
LGTM. Nice!

internal/testing/types/bool.go
detro: I have checked, and both Goland and `goimports` do the same thing: they seem to be putting "standard", then "3rd party", then "local" imports, separated by a space.

I thought that was the "Go way" to do imports, no?

bflad: https://github.com/golang/go/wiki/CodeReviewComments#imports tends to be my jam. Not sure why certain things would default to splitting out the "local" imports group versus not.

PR #336, opened 2022-05-19
internal: Add ProviderMeta handling to fwserver

bflad:
This introduces provider meta schema handling, similar to provider schema handling. This will ensure a smoother migration for RPCs that require this information, encapsulate the interface type handling within `internal/fwserver`, and prevent extra executions against the `Provider` `GetMetaSchema()` method.


bendbennett (approved):
LGTM

detro (approved):
So many calls to `GetMetaSchema` saved ;) 

LGTM

PR #343, opened 2022-05-23
Remove unnecessary conversion to `tftypes` in `AttributePlanModifier`s implementations

detro:
Small follow up from #335


bflad (approved):
So fresh and so clean 🚀

bendbennett (approved):
Nice!

Issue #326, opened 2022-05-13
Missing NestingMode in GetSchema makes Terraform crash


abergmeier:
### Module version

```
github.com/hashicorp/terraform-plugin-framework v0.8.0
```

### Relevant provider source code

```go

var (
	outputDockerAttributes = map[string]tfsdk.Attribute{}
	outputBlocks                             = map[string]tfsdk.Block{
		"docker": {
			Attributes: outputDockerAttributes,
			MaxItems:   1,
			PlanModifiers: tfsdk.AttributePlanModifiers{
				tfsdk.RequiresReplace(),
			},
		},
	}
)

func (t builtResourceType) GetSchema(ctx context.Context) (tfsdk.Schema, diag.Diagnostics) {
	return tfsdk.Schema{
		Blocks: map[string]tfsdk.Block{
			"output": {
				MinItems:            1,
				MaxItems:            1,
				Blocks:              outputBlocks,
				Description:         "Output destination. For entries of type gha, GitHub Action credentials are automatically added to attrs.",
				MarkdownDescription: "Output destination. For entries of `type` `gha`, GitHub Action credentials are automatically added to attrs.",
				PlanModifiers: tfsdk.AttributePlanModifiers{
					tfsdk.RequiresReplace(),
				},
			},
		},
	}
}
```

### Debug Output

See https://gist.github.com/abergmeier/fe55ad498453c30c85dde63aca2aa54b

### Expected Behavior

framework should have validation in place to not allow "invalid" Schemas.

### Actual Behavior

Make Terraform crash.

### Steps to Reproduce

terraform apply

abergmeier:
Introducing `NestingMode: ` seems to fix the problem - so maybe validate whether `NestingMode` is unknown?

bflad:
Hi @abergmeier 👋  Thank you for raising this and sorry you ran into trouble here.

Glad we have a lead on what caused the issue. We should also ensure there is a bug report raised in the [Terraform CLI issue tracker](https://github.com/hashicorp/terraform/issues), because they should likely raise a proper error instead of crashing. I'm guessing this is on the latest Terraform CLI version?

In terms of resolution here, we likely cannot raise a compiler error without making the schema code more complicated, however we do have plans in #113 to add provider unit testing functionality, similar to how you could get unit testing failures with terraform-plugin-sdk.

Do you also think the documentation could be improved anywhere to help avoid this? Your perspective would be super helpful. Thanks!

[...]

PR #348, opened 2022-05-24
internal: Introduce testprovider package and implement fwserver GetProviderSchema unit testing

bflad:

Introduces a declarative `tfsdk.Provider` for easier unit testing implementations. Over time this can replace the monolithic `proto6server.testServeProvider`, while enabling unit testing across both `fwserver`, `proto6server`, and potentially a `proto5server` implementation. Unit testing for all possible attribute/block types is already handled via `toproto6` package unit testing such as `TestBlock` and `TestSchemaAttribute`. Specific error handling cases such as https://github.com/hashicorp/terraform-plugin-framework/issues/326 can be added for confirmation afterwards.


bendbennett (approved):
LGTM

PR #354, opened 2022-05-27
Prevent `ToTerraformValue` panic when the underlying `attr.Value` types doesn't have type information for its fields

detro:
`ToTerraformValue` in all the non _primitive_ types of `attr.Value`, relies on it containing `ElemType` or `AttrsType` to do the conversion to TF types.

But if by mistake one omits setting that, the framework was throwing a panic as it wasn't checking for those fields being not null. Now it does.

### DIFF
--- a/types/list.go
+++ b/types/list.go
@@ -163,7 +163,7 @@ func (l List) Type(ctx context.Context) attr.Type {
 // a tftypes.Value.
 func (l List) ToTerraformValue(ctx context.Context) (tftypes.Value, error) {
 	if l.ElemType == nil {
-		return tftypes.Value{}, fmt.Errorf("cannot convert List to Terraform Value if ElemType field is null")
+		return tftypes.Value{}, fmt.Errorf("cannot convert List to tftypes.Value if ElemType field is not set")
 	}
 	listType := tftypes.List{ElementType: l.ElemType.TerraformType(ctx)}
 	if l.Unknown {
--- a/types/list.go
+++ b/types/list.go
@@ -162,6 +162,9 @@ func (l List) Type(ctx context.Context) attr.Type {
 // ToTerraformValue returns the data contained in the AttributeValue as
 // a tftypes.Value.
 func (l List) ToTerraformValue(ctx context.Context) (tftypes.Value, error) {
+	if l.ElemType == nil {
+		return tftypes.Value{}, fmt.Errorf("cannot convert List to Terraform Value if ElemType field is null")
+	}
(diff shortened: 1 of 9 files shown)


bendbennett (approved):
LGTM

types/list_test.go
```diff
@@ -361,16 +362,32 @@ func TestListToTerraformValue(t *testing.T) {
...
 		name, test := name, test
 		t.Run(name, func(t *testing.T) {
 			t.Parallel()
 
-			got, err := test.input.ToTerraformValue(context.Background())
-			if err != nil {
-				t.Errorf("Unexpected error: %s", err)
-				return
+			got, gotErr := test.input.ToTerraformValue(context.Background())
+			if gotErr != nil {
```
bflad, line 382: Nit: It'd be good to check against `gotErr == nil && test.expectedErr != nil` after this `if` conditional as well.

Another, potentially simpler way to test this is via `cmp.Diff()` as well.

types/list.go
```diff
@@ -162,6 +162,9 @@ func (l List) Type(ctx context.Context) attr.Type {
 // ToTerraformValue returns the data contained in the AttributeValue as
 // a tftypes.Value.
 func (l List) ToTerraformValue(ctx context.Context) (tftypes.Value, error) {
+	if l.ElemType == nil {
+		return tftypes.Value{}, fmt.Errorf("cannot convert List to Terraform Value if ElemType field is null")
```
bflad, line 166: Nit: I'd suggest being a little bit more explicit about the typing and using "missing" or "nil" over "null" (which is a Terraform specific concept in the type system)

```suggestion
		return tftypes.Value{}, fmt.Errorf("cannot convert List to tftypes.Value if ElemType field is not set")
```

bflad (approved):
Overall 👍 Just some small things for consideration 🚀

types/list_test.go
detro: I have updated all the tests I added to check the result this way:

```golang
			if test.expectedErr == "" && gotErr != nil {
				t.Errorf("Unexpected error: %s", gotErr)
				return
			}

			if test.expectedErr != "" {
				if gotErr == nil {
					t.Errorf("Expected error to be %q, got none", test.expectedErr)
					return
				}

				if test.expectedErr != gotErr.Error() {
					t.Errorf("Expected error to be %q, got %q", test.expectedErr, gotErr.Error())
					return
				}
			}
```

This should cover all combinations we want to spot.

PR #357, opened 2022-05-30
Adding additional tests for ValueAs

bendbennett:
Expanding test coverage for `ValueAs` to include more complex types.

### DIFF
--- a/tfsdk/value_as_test.go
+++ b/tfsdk/value_as_test.go
@@ -8,12 +8,13 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+
 	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/internal/reflect"
 	testtypes "github.com/hashicorp/terraform-plugin-framework/internal/testing/types"
 	"github.com/hashicorp/terraform-plugin-framework/types"
-	"github.com/hashicorp/terraform-plugin-go/tftypes"
 )
 
 func newStringPointer(in string) *string {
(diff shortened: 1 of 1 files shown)


detro (approved):

tfsdk/value_as_test.go
```diff
@@ -40,11 +89,225 @@ func TestValueAs(t *testing.T) {
...
+			val:      types.Int64{Value: 12},
+			target:   newInt64Pointer(0),
+			expected: newInt64Pointer(12),
+		},
+		"primitive int64 pointer pointer": {
+			val:      types.Int64{Value: 12},
+			target:   newInt64PointerPointer(0),
+			expected: newInt64PointerPointer(12),
+		},
+		// Following test fails as target.Type() is big.Float not *bigFloat.
```
bendbennett, line 122: @bflad is this a bug? Just wondering if `into.go` needs to be amended to handle this use-case?

bflad: My limited understanding with the [`math/big` package](https://pkg.go.dev/math/big) in general is to conventionally always use pointer types to save on various memory allocations and prevent issues with zero values on the struct. So doing something like:

```go
var myBigOleFloat big.Float
tfsdk.ValueAs(ctx, example, &myBigOleFloat)
```

Would be considered something of an anti-pattern. Could support it? I guess if it is not hard to implement it. _Should_ support it? Probably more debatable, although if the error behavior/messaging is unclear then maybe that adds some weight to supporting it.

bflad (approved):
Looks good to me 🚀

tfsdk/value_as_test.go
```diff
@@ -90,8 +349,20 @@ func TestValueAs(t *testing.T) {
 				return
 			}
 
-			if diff := cmp.Diff(tc.expected, tc.target); diff != "" {
-				t.Fatalf("Unexpected diff in results (-wanted, +got): %s", diff)
+			// Cannot use cmp.Diff for comparing big.Float, requires usage of *big.Float.Cmp()
```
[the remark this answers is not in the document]
bendbennett: Good shout. Have added a `cmp.Comparer` and removed the `switch`.

tfsdk/value_as_test.go
bendbennett: Ok. I'll leave this for now and re-visit if it becomes obvious that behaviour/messaging is unclear.

PR #359, opened 2022-06-02
internal: Implement fwserver ReadDataSource testing and update proto6server testing

bflad:

This also changes a potential panic into an error where the incoming Config is missing, which was discovered via the new unit testing. Terraform always fills in this data currently, so it does not warrant any CHANGELOG entry, but is good from a defensive coding standpoint.


bendbennett (approved):
LGTM

PR #367, opened 2022-06-08
internal: Implement fwserver PlanResourceChange testing and update proto6server testing

### LINKED ISSUE #215: Refactor how we're serving providers

bflad:
Closes #215

Also removes the placeholder `internal/testing/emptyprovider` implementation in preference of the fully built out declarative one in `internal/testing/testprovider`.

At this point we should hopefully feel pretty confident in the framework and protocol version 6 server unit testing focusing on their logic and data handoffs to other functions/methods. This also means we should be in a good spot to create a protocol version 5 implementation, very similar to the protocol version 6 implementation.


detro (approved):
👍

Issue #158, opened 2021-09-14
Provide an upgrade path for SDKv2 providers


paddycarver:
As we build out the framework and it becomes a viable replacement for SDKv2, we need to define the migration path for providers built on SDKv2 to be built on framework, instead. We don't want to make provider devs rewrite their entire provider in one fell swoop. Ideally, people would be able to migrate over one resource at a time. This is going to take some doing, so we can keep practitioner-facing breaking changes in the control of provider developers.

* [x] hashicorp/terraform-plugin-mux#19
  * needed so we can have two protocol 6 providers muxed together
* [x] hashicorp/terraform-plugin-go-contrib#7
  * needed to make SDKv2 providers into protocol 6 providers
* [x] #85 
  * needed so provider blocks that use blocks can match (which mux requires)
  * needed so resources that use blocks can be upgraded without requiring a breaking change
* [x] #214
  * should be verified to ensure migration without the special SDKv2 logic works as expected
* [x] #222
  * needed so blocks have parity with attributes
* [x] ~#159~
  * ~even when providers can control it, it may be too much to ask of practitioners to insert `=` in so many places. If we ever want to get rid of blocks support, we need to solve this problem somehow.~ (out of scope)
* [x] Create terraform.io SDK -> framework migration page, which outlines the optional process of using terraform-plugin-mux functionality to individually migrate resources

bflad:
The last item, creating a terraform-plugin-sdk migration guide, is started here: https://www.terraform.io/plugin/framework/migrating

Closing!

PR #368, opened 2022-06-09
providerserver: Implement native protocol version 5 support

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/158

This enables provider developers to natively create and serve framework providers on protocol version 5, which is compatible with Terraform CLI 0.12 and later. For provider developers that are using terraform-plugin-mux to migrate from terraform-plugin-sdk to this framework, they can now omit the usage of the `tf6to5server` package by setting the `ServeOpts` type `ProtocolVersion` field to `5`.

Any protocol layer restrictions with the older protocol version 5 are raised as errors from the framework. For example, if `Attributes` is defined in a `tfsdk.Attribute` ("nested attributes"), the `GetProviderSchema` RPC response will include an error diagnostic.

This implementation is essentially a duplication of the `internal/fromproto6`, `internal/proto6server`, and `internal/toproto6` packages with renaming of 6 to 5 and any minor RPC naming differences to match the protocol (e.g. `ValidateDataResourceConfig` -> `ValidateDataSourceConfig`, `ValidateProviderConfig` -> `PrepareProviderConfig`, `ValidateResourceConfig` -> `ValidateResourceTypeConfig`).


bendbennett (approved):
I'm going to be completely honest, I only skimmed this owing to the size of the PR but it LGTM.

internal/fromproto5/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package fromproto5 contains functions to convert from protocol version 6
```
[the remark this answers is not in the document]
bflad: Ah yes, copypasta here and probably in the two other doc files, thanks for catching.

internal/proto5server/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package proto5server contains the provider server implementation compatible
+// with protocol version 6 (tfprotov5.ProviderServer).
```
bflad, line 2:
```suggestion
// with protocol version 5 (tfprotov5.ProviderServer).
```

internal/toproto5/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package toproto5 contains functions to convert from framework types to
+// protocol version 6 (tfprotov5) types.
```
bflad, line 2:
```suggestion
// protocol version 5 (tfprotov5) types.
```

PR #369, opened 2022-06-09
diag: Move tfprotov6.DiagnosticSeverity conversion to internal/toproto6

bflad:
Technical debt cleanup now that the `internal/toproto6` package exists. No CHANGELOG entry necessary as this was not intended for provider developer usage.

### DIFF
--- a/internal/toproto6/diagnostics_test.go
+++ b/internal/toproto6/diagnostics_test.go
@@ -10,6 +10,42 @@ import (
 	"github.com/hashicorp/terraform-plugin-go/tftypes"
 )
 
+func TestDiagnosticSeverity(t *testing.T) {
+	t.Parallel()
+
+	testCases := map[string]struct {
+		severity diag.Severity
+		expected tfprotov6.DiagnosticSeverity
+	}{
+		"error": {
+			severity: diag.SeverityError,
+			expected: tfprotov6.DiagnosticSeverityError,
+		},
+		"invalid": {
+			severity: diag.SeverityInvalid,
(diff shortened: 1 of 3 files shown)


internal/toproto6/diagnostics_test.go
```diff
@@ -10,6 +10,42 @@ import (
...
+			expected: tfprotov6.DiagnosticSeverityInvalid,
+		},
+		"warning": {
+			severity: diag.SeverityWarning,
+			expected: tfprotov6.DiagnosticSeverityWarning,
+		},
+	}
+
+	for name, testCase := range testCases {
+		name, testCase := name, testCase
```
bendbennett, line 35: Just wondering why the `name` and `testCase` vars that are declared in the `for` statement are not used directly?

bendbennett (approved):
LGTM

internal/toproto6/diagnostics_test.go
detro: I was going to ask the same question.
Then I tried: https://go.dev/play/p/JFh-5r_DgnS

I ... don't like this thing of Go.
The closure should capture the value, not forcing this so that the value gets copied in a separate variable each time.

Well, I guess I'd expect Go to do it by itself.

detro (approved):
LGTM (except, I don't like how Go does closuring).

Because it's essentially silent about why you will likely have a bug if you don't do the extra variable copying.

Java would simply refuse to work and force you to finalise the variable before you can use it in the lambda.

Anyway, 👍

PR #376, opened 2022-06-14
Add `.String()` method to `attr.Value` interface

### LINKED ISSUE #117: Better string representation of `terraform-plugin-framework/types` values
For logging and debugging purposes a better representation of `terraform-plugin-framework/types` values would help.
For example, currently `fmt.Printf("%s")` of a `types.Number` with value `6.5` displays:

```
{%!s(bool=false) %!s(bool=false) %!s(*big.Float=6.5)}
```

detro:
## Background

At the moment, it can be tricky to log or report errors where the aim is to represent a `attr.Value` concrete type. For a `types.Number` for example, it can look like this:

```
{%!s(bool=false) %!s(bool=false) %!s(*big.Float=6.5)}
```

This is not very practitioner friendly, when the aim is to put such representation in a warning or error diagnostic.

## Proposal

This PR provides a new `.String()` method on the `attr.Value` interface, that aims to address this issue. The representation for each type would follow these formatting "rules":

| Type      | Formatting                                | Unknown     | Null     | Example                                       |
| ---------:|:-----------------------------------------:|:-----------:|:--------:|:----------------------------------------------|
| `Bool`    | `true`/`false`                            | `<unknown>` | `<null>` | `true`                                        |
| `Int64`   | `%d`                                      | `<unknown>` | `<null>` | `123456`                                      |
| `Float64` | `%f`                                      | `<unknown>` | `<null>` | `123.456`                                     |
| `Number`  | `big.Float.String()`                      | `<unknown>` | `<null>` | `9.238793817e+13`                             |
| `String`  | `"__VALUE__"`                             | `<unknown>` | `<null>` | `"hello world"`                               |
| `List`    | `[v1.String(),v2.String(),...]`           | `<unknown>` | `<null>` | `["hello","world"]`                           |
| `Set`     | `[v2.String(),v1.String(),...]`           | `<unknown>` | `<null>` | `["hello","world"]`                           |
| `Map`     | `{"k1":v1.String(),"k2":v2.String(),...}` | `<unknown>` | `<null>` | `{"alpha":1234,"beta":56789}`                 |
| `Object`  | `{"k1":v1.String(),"k2":v2.String(),...}` | `<unknown>` | `<null>` | `{"alpha":1234,"beta":56789,"gamma":"ammag"}` |

For collection types `List` and `Set`, we use the delimiters `[ / ]`;
for collection types `Map` and `Object`, we use the delimiters `{ / }`.
This creates a notation _similar_ to how HCL represents those entities.

## Related

Closes #117

### DIFF
--- a/attr/value.go
+++ b/attr/value.go
@@ -7,11 +7,13 @@ import (
 )
 
 const (
-	// UnknownString should be returned by Value.String() implementations, when Value.IsUnknown() returns true.
-	UnknownString = "<unknown>"
+	// UnknownValueString should be returned by Value.String() implementations,
+	// when Value.IsUnknown() returns true.
+	UnknownValueString = "<unknown>"
 
-	// NullString should be returned by Value.String() implementations, when Value.IsNull() returns true.
-	NullString = "<null>"
+	// NullValueString should be returned by Value.String() implementations
+	// when Value.IsNull() returns true.
+	NullValueString = "<null>"
 )
 
 // Value defines an interface for describing data associated with an attribute.
@@ -36,10 +38,11 @@ type Value interface {
 	IsUnknown() bool
 
 	// String returns a summary representation of either the underlying Value,
-	// or UnknownString (`<unknown>`) when IsUnknown() returns true,
-	// or NullString (`<null>`) when IsNull() return true.
+	// or UnknownValueString (`<unknown>`) when IsUnknown() returns true,
+	// or NullValueString (`<null>`) when IsNull() return true.
 	//
-	// This is an intentionally lossy representation, that it's best used for
-	// logging and error reporting.
+	// This is an intentionally lossy representation, that are best suited for
+	// logging and error reporting, as they are not protected by
+	// compatibility guarantees within the framework.
 	String() string
 }
--- a/attr/value.go
+++ b/attr/value.go
@@ -6,6 +6,14 @@ import (
 	"github.com/hashicorp/terraform-plugin-go/tftypes"
 )
 
+const (
+	// UnknownString should be returned by Value.String() implementations, when Value.IsUnknown() returns true.
+	UnknownString = "<unknown>"
+
+	// NullString should be returned by Value.String() implementations, when Value.IsNull() returns true.
+	NullString = "<null>"
+)
+
 // Value defines an interface for describing data associated with an attribute.
 // Values allow provider developers to specify data in a convenient format, and
 // have it transparently be converted to formats Terraform understands.
@@ -26,4 +34,12 @@ type Value interface {
 
 	// IsUnknown returns true if the value is not yet known.
 	IsUnknown() bool
+
+	// String returns a summary representation of either the underlying Value,
+	// or UnknownString (`<unknown>`) when IsUnknown() returns true,
+	// or NullString (`<null>`) when IsNull() return true.
+	//
+	// This is an intentionally lossy representation, that it's best used for
+	// logging and error reporting.
+	String() string
 }
--- a/internal/testing/types/string.go
+++ b/internal/testing/types/string.go
@@ -41,14 +41,14 @@ func (t StringType) TerraformType(_ context.Context) tftypes.Type {
 func (t StringType) ValueFromTerraform(ctx context.Context, in tftypes.Value) (attr.Value, error) {
 	if !in.IsKnown() {
 		return String{
-			String:    types.String{Unknown: true},
-			CreatedBy: t,
+			InternalString: types.String{Unknown: true},
+			CreatedBy:      t,
 		}, nil
 	}
 	if in.IsNull() {
 		return String{
-			String:    types.String{Null: true},
-			CreatedBy: t,
+			InternalString: types.String{Null: true},
+			CreatedBy:      t,
 		}, nil
 	}
 	var s string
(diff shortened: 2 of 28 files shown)


types/set_test.go
```diff
@@ -962,3 +962,72 @@ func TestSetEqual(t *testing.T) {
...
+			input: Set{
+				ElemType: StringType,
+				Elems: []attr.Value{
+					String{Value: "hello"},
+					String{Value: "world"},
+				},
+			},
+			expectation: "(hello,world)",
+		},
+		"list-of-lists": {
```
bendbennett, line 984:
```suggestion
		"set-of-sets": {
```

bendbennett (approved):
Like it 👍

internal/testing/types/string.go
```diff
@@ -57,17 +57,21 @@ func (t StringType) ValueFromTerraform(ctx context.Context, in tftypes.Value) (a
...
-		String:    types.String{Value: s},
-		CreatedBy: t,
+		InternalString: types.String{Value: s},
+		CreatedBy:      t,
 	}, nil
 }
 
 type String struct {
-	types.String
+	InternalString types.String
```
detro, line 66: I had to change the name we use for the embedded struct here, as now it was clashing with the method name.

types/map.go
```diff
@@ -228,3 +230,33 @@ func (m Map) IsNull() bool {
...
+	// We want the output to be consistent, so we sort the output by key
+	keys := make([]string, 0, len(m.Elems))
+	for k := range m.Elems {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+
+	var res strings.Builder
+
+	res.WriteString("[")
```
bflad, line 252: For a map, I'd suggest curly braces (e.g. `{`) to differentiate between collection types 👍

types/map.go
```diff
@@ -228,3 +230,33 @@ func (m Map) IsNull() bool {
...
+	for i, k := range keys {
+		if i != 0 {
+			res.WriteString(",")
+		}
+		res.WriteString(fmt.Sprintf("%s:%s", k, m.Elems[k].String()))
```
bflad, line 257: I think we should quote (e.g. `%q`) the keys for clarity, as they are all strings and can contain spaces, braces, commas, colons, etc.

types/object.go
```diff
@@ -285,3 +285,33 @@ func (o Object) IsNull() bool {
...
+	// We want the output to be consistent, so we sort the output by key
+	keys := make([]string, 0, len(o.Attrs))
+	for k := range o.Attrs {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+
+	var res strings.Builder
+
+	res.WriteString("(")
```
bflad, line 307: Terraform tends to use curly braces (e.g. `{`) for objects, so maybe best to stick with that to prevent confusion. 👍

types/set.go
```diff
@@ -294,3 +295,26 @@ func (s Set) IsNull() bool {
...
+		return attr.UnknownString
+	}
+
+	if s.Null {
+		return attr.NullString
+	}
+
+	var res strings.Builder
+
+	res.WriteString("(")
```
bflad, line 310: Terraform also tends to bucket sets with the tuple/list "syntax" of square braces (e.g. `[`)

types/string.go
```diff
@@ -82,3 +82,15 @@ func (s String) IsNull() bool {
...
+func (s String) String() string {
+	if s.Unknown {
+		return attr.UnknownString
+	}
+
+	if s.Null {
+		return attr.NullString
+	}
+
+	return s.Value
```
bflad, line 95: It might be good to quote the value, e.g. `fmt.Sprintf("%q", s.Value)` as it may contain spaces, braces, commas, colons, etc that may interfere visually with other string outputs. Another odd case is an empty, but not null string causing the output to be like `[]` for a list with one empty string element, which I think is confusing.

bflad (commented):
Overall this is really good! Just some minor quibbles/discussions on the output "syntax" so it aligns a little closer with Terraform's configuration language syntax and removes some ambiguities for readers.

attr/value.go
```diff
@@ -26,4 +34,12 @@ type Value interface {
 
 	// IsUnknown returns true if the value is not yet known.
 	IsUnknown() bool
+
+	// String returns a summary representation of either the underlying Value,
+	// or UnknownString (`<unknown>`) when IsUnknown() returns true,
+	// or NullString (`<null>`) when IsNull() return true.
+	//
+	// This is an intentionally lossy representation, that it's best used for
```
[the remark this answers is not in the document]
detro: I can definitely add something here, but... what exactly do you mean?
That we reserve the right to change what value is returned here?

types/map.go
detro: I like it!

types/map.go
detro: 👍

types/string.go
detro: OK.

detro:
All right @bflad - I should have addressed everything you raised. Let me know, especially about the godoc for the `.String()` method itself.

types/bool_test.go
```diff
@@ -269,3 +269,46 @@ func TestBoolEqual(t *testing.T) {
...
+			expectation: "<null>",
+		},
+		"default-false": {
+			input:       Bool{},
+			expectation: "false",
+		},
+	}
+
+	for name, test := range tests {
+		name, test := name, test
```
SBGoods, line 304: Nit: Do we need this declaration? It may be redundant

SBGoods: Same comment for the rest of the tests

SBGoods (approved):

types/bool_test.go
detro: it is one thing of Go that I really don't like: it's about how closures keep a reference to the variable and not the value.

This means that each closure would refer to the last value `name` and `test` got during the loop, not the one that was in `name` and `test` when the specific closure got created.

So, the only way to circumvent it is to just create 2 variables local to the loop.

To show the effects, https://go.dev/play/p/JFh-5r_DgnS

attr/value.go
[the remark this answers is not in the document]
bflad: I think we want to be careful that folks don't start "depending" on those strings to do anything like equality checking or building downstream tooling that is dependent on any specific syntax, etc. decisions. Being explicit about it in our Go documentation is a good signal that there are no compatibility promises now or in the future.

PR #383, opened 2022-06-15
tfsdk: Clarify NestingMode requirement in Block Go documentation

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/326


detro (approved):
nice

PR #384, opened 2022-06-16
Detro/381 godoc for all types

### LINKED ISSUE #381: Add godoc to all the methods of `types.*`
### Module version
```
0.9.0
```

### Use-cases

* Documentation
* IDE-friendlyness

### Background

Currently, all the `types.*`, implementors of `attr.Value`, don't expose any godoc. This is circumvented by going up the implementation hierarchy and looking at the godoc of `attr.Value`.

This is sub-optimal, as it forces developers to either dig through our code OR to open [go.dev](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/attr#Value).

### Proposal

Add the godoc documentation for each method implemented by `types.*`, on the implementors themselves.

### References

* Relates to #376 
* Relates to #335

detro:
Closes #381 

_Sorry, I couldn't resist_ 😜


types/bool.go
```diff
@@ -79,14 +79,19 @@ func (b Bool) Equal(other attr.Value) bool {
 	return b.Value == o.Value
 }
 
+// IsNull returns true if the Bool represents null value.
```
bflad, line 82: Nit: Missing article here and other IsNull

```suggestion
// IsNull returns true if the Bool represents a null value.
```

types/bool.go
```diff
@@ -79,14 +79,19 @@ func (b Bool) Equal(other attr.Value) bool {
...
 func (b Bool) IsNull() bool {
 	return b.Null
 }
 
+// IsUnknown returns true if the Bool represents currently unknown value.
```
bflad, line 87: Nit: Missing article here and other IsUnknown

```suggestion
// IsUnknown returns true if the Bool represents a currently unknown value.
```

types/bool.go
```diff
@@ -79,14 +79,19 @@ func (b Bool) Equal(other attr.Value) bool {
...
 func (b Bool) IsUnknown() bool {
 	return b.Unknown
 }
 
+// String returns a summary and lossy representation of the Bool.
+// The string returned here is not protected by any compatibility guarantees,
+// and it's better suited for logging and error reporting.
```
bflad, line 94: Maybe these could be clarified slightly?

```suggestion
// String returns a human readable representation of the Bool value.
// The string returned here is not protected by any compatibility guarantees,
// and is intended for logging and error reporting.
```

types/number.go
```diff
@@ -86,24 +85,25 @@ func (n Number) Equal(other attr.Value) bool {
 	return n.Value.Cmp(o.Value) == 0
 }
 
+// IsNull returns true if the Number represents null value.
 func (n Number) IsNull() bool {
-	return n.Null
+	return n.Null || (!n.Unknown && n.Value == nil)
```
bflad, line 90: I'm 👍 for this change, however it should include a changelog entry, maybe something along the lines of an enhancement of types: The Number type IsNull method will now return true for the zero-value of a Number.

bflad (commented):
Overall excellent changes, just a few things for consideration 😄 Note: this may fail CI on updates until https://github.com/hashicorp/terraform-plugin-framework/pull/385 is handled in some fashion and rebased here.

bflad (approved):
Looks good to me 🚀  Also thank you for fixing some of those `AttributeValue` -> `attr.Value` comments.

Issue #365, opened 2022-06-08
Consider Refactoring tfsdk Schema/Attribute/Block Logic into internal Package(s)


bflad:
### Module version

```
v0.8.0
```

### Use-cases

While working through some of the framework server refactoring as part of #215, it was necessary to extract some of the `tfsdk.Attribute`, `tfsdk.Block`, and `tfsdk.Schema` unexported methods into temporary functions within the `internal/fwserver` package so they did not become part of the public `tfsdk` APIs. For example, `SchemaValidate()` and `SchemaModifyPlan()`. This highlighted that a larger portion of the schema handling would benefit as being in internal packages and colocated again, rather than splitting the logic across packages, while still exporting provider developer functionality only as necessary in `tfsdk` or some other "public" package.

### Proposal

To be written up, but a current thought is that there should be an `internal/schema` or similar package which contains the full framework schema/attribute/block handling, while `tfsdk` or another package merely exposes anything that should be in the public API.

### References

- #215

bflad:
This type of change is essentially required for #132, which provider developers are asking to be implemented.

bflad:
This issue should be closable early in the v0.12.0 development cycle once we've released deprecations of some existing functionality that shouldn't affect most provider developers.

bflad:
This `internal` package refactoring was completed previously. Other issues can be used to track additional efforts.

Issue #81, opened 2021-07-27
Support AttributePaths natively


paddycarver:
### Module version

```
v0.2.0
```

### Use-cases

We're still using `*tftypes.AttributePath` all over the place, and that's really verbose to build and can be bothersome to work with, given how close it sticks to the protocol. We should come up with a framework-specific abstraction for attribute paths.

Ideally, this would be much less verbose without sacrificing any of the clarity. Perhaps just shortening the names would be enough?

This would also ideally solve problems we've seen in validation, etc. of wanting to be able to specify wildcard/relative attribute paths. E.g., parent, self, sibling, etc. This would solve the problem with `ConflictsWith`, etc. behaviors where you may not know the full path, but you mean "this other attribute of the same object". We'd just need to be careful with this, as we want to make sure people don't try and send _these_ attribute paths back in diagnostics.

### Attempted Solutions

#52

PR #390, opened 2022-06-22
path: Introduce package with initial tftypes.AttributePath abstraction

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/81
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/161
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/172

This introduces a native abstraction over terraform-plugin-go's `tftypes.AttributePath`, allowing the framework to own the implementation details and extend the functionality further. Provider developers will be closer to removing a direct dependency on terraform-plugin-go. This is a major breaking change, however it is necessary before 1.0 to prevent compatibility issues in the future and long has been a goal of the framework.

This functionality is only intended to replace `tftypes.AttributePath` usage and not mess with the `tfsdk.Config`, `tfsdk.Plan`, and `tfsdk.State` type `tftypes.Value` data storage fields or their underlying data manipulation logic. This does leave the framework in an awkward half-state until those are further refactored (likely towards native `attr.Value`), but is done to try and reduce the review complexity of this initial migration. Additional followup changes will introduce the concept of path expressions, which will allow provider developers to match against multiple paths or reference parent paths in schema-based plan modifier and validation functionality.

To prevent import cycles between `attr` and `diag` packages due changing the `attr.TypeWithValidate` type `Validate` method signature from `Validate(context.Context, tftypes.Value, *tftypes.AttributePath) diag.Diagnostics` to `Validate(context.Context, tftypes.Value, path.Path) diag.Diagnostics`, the `TypeWithValidation` interface is moved a new `xattr` package underneath `attr` with Go and website documentation to direct provider developers to the other package. This will also solve a prior issue with trying to implement `TypeWithModifyPlan` support.

Naming and location of anything in this initial implementation can be adjusted as necessary.

Provider developers can migrate to the new path handling by replacing:

```go
tftypes.NewAttributePath().WithAttributeName("example")
```

With the new `github.com/hashicorp/terraform-plugin-framework/path` import equivalent:

```go
path.Root("example")
```

Then, if necessary, using the `(Path).At*` methods to extend the path definition instead of `(*tftypes.AttributePath).With*` methods:

| Current                  | New             |
| ------------------------ | --------------- |
| `WithAttributeName()`    | `AtName()`      |
| `WithElementKeyInt()`    | `AtListIndex()` |
| `WithElementKeyString()` | `AtMapKey()`    |
| `WithElementKeyValue()`  | `AtSetValue()`  |


bendbennett (approved):
I've only looked through the `path` pkg, but that LGTM.

path/path.go
```diff
@@ -0,0 +1,129 @@
...
+
+// RootPath creates a new attribute path starting with a root
+// PathStepAttributeName.
+func RootPath(rootAttributeName string) Path {
+	return Path{
+		steps: PathSteps{
+			PathStepAttributeName(rootAttributeName),
+		},
+	}
+}
```
bflad, line 129: Given it worked out that this could be in its own package, I wonder if its worth shortening these these to just `path.Root(string)` and `path.Empty()`?

bflad: I'll update and push up, one second

detro (approved):
I like this a lot.

Especially the hierarchy and combination of `Path` and `PathStep`, I find it quite elegant.

The builder-patterns to compose `Path`s is also quite concise and expressive, leading to smaller lines of code.

The CI is failing, probably a place where the replacement is still needed, but feels easy to fix:
```
Run go test -v ./internal/framework5provider
# github.com/hashicorp/terraform-provider-corner/internal/framework5provider [github.com/hashicorp/terraform-provider-corner/internal/framework5provider.test]
Error: internal/framework5provider/resource_user.go:[1](https://github.com/hashicorp/terraform-plugin-framework/runs/6995717727?check_suite_focus=true#step:8:1)99:90: cannot use tftypes.NewAttributePath().WithAttributeName("name") (type *tftypes.AttributePath) as type path.Path in argument to tfsdk.ResourceImportStatePassthroughID
FAIL	github.com/hashicorp/terraform-provider-corner/internal/framework[5](https://github.com/hashicorp/terraform-plugin-framework/runs/6995717727?check_suite_focus=true#step:8:6)provider [build failed]
FAIL
Error: Process completed with exit code 2.
```

bflad:
Updated the "starter" functions to just `Empty` and `Root` 👍 terraform-provider-corner will require a separate pull request to migrate everything for the breaking changes and verify things -- I'll work on that now and drop a link here.

bflad:
Downstream terraform-provider-corner updates: https://github.com/hashicorp/terraform-provider-corner/pull/73

bflad:
If anyone has any additional feedback on these changes, please reach out to me and we can go through it together before the next release. I'll also ensure terraform-provider-corner is pointing at main after merge so CI keeps working.

Issue #391, opened 2022-06-27
Add convenience functions for diagnostics `ErrorsCount`, `WarningsCount`, `Errors` and `Warnings`


bendbennett:
### Module version
```
v0.9.0
```

### Use-cases
Within terraform-plugin-framework-validators there is a [PR](https://github.com/hashicorp/terraform-plugin-framework-validators/pull/43) in which it would be useful to be able to retrieve `Warnings` from `Diagnostics`. This seems like a potentially generic use-case.

### Proposal
Add the following functions to `Diagnostics`:
- `ErrorsCount()`
- `WarningsCount()`
- `Errors()`
- `Warnings()`

PR #392, opened 2022-06-27
Add convenience functions for `ErrorsCount`, `WarningsCount`, `Errors` and`Warnings`

bendbennett:
Closes: #391 

Adds the following convenience functions to Diagnostics:

- `ErrorsCount()`
- `WarningsCount()`
- `Errors()`
- `Warnings()`

### DIFF
--- a/diag/diagnostics.go
+++ b/diag/diagnostics.go
@@ -70,3 +70,55 @@ func (diags Diagnostics) HasError() bool {
 
 	return false
 }
+
+// ErrorsCount returns the number of Diagnostic in Diagnostics that are SeverityError.
+func (diags Diagnostics) ErrorsCount() int {
+	count := 0
+
+	for _, d := range diags {
+		if SeverityError == d.Severity() {
+			count++
+		}
+	}
+
+	return count
+}
+
+// WarningsCount returns the number of Diagnostic in Diagnostics that are SeverityWarning.
+func (diags Diagnostics) WarningsCount() int {
+	count := 0
+
+	for _, d := range diags {
+		if SeverityWarning == d.Severity() {
+			count++
+		}
+	}
+
+	return count
+}
+
+// Errors returns all the Diagnostic in Diagnostics that are SeverityError.
+func (diags Diagnostics) Errors() Diagnostics {
+	var dd Diagnostics
+
+	for _, d := range diags {
+		if SeverityError == d.Severity() {
+			dd = append(dd, d)
+		}
+	}
+
+	return dd
+}
+
+// Warnings returns all the Diagnostic in Diagnostics that are SeverityWarning.
+func (diags Diagnostics) Warnings() Diagnostics {
+	var dd Diagnostics
+
+	for _, d := range diags {
+		if SeverityWarning == d.Severity() {
+			dd = append(dd, d)
+		}
+	}
+
+	return dd
+}
--- a/diag/diagnostics.go
+++ b/diag/diagnostics.go
@@ -83,7 +83,7 @@ func (diags Diagnostics) WarningsCount() int {
 
 // Errors returns all the Diagnostic in Diagnostics that are SeverityError.
 func (diags Diagnostics) Errors() Diagnostics {
-	var dd Diagnostics
+	dd := Diagnostics{}
 
 	for _, d := range diags {
 		if SeverityError == d.Severity() {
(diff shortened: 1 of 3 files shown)


diag/diagnostics.go
```diff
@@ -70,3 +70,55 @@ func (diags Diagnostics) HasError() bool {
...
+		if SeverityError == d.Severity() {
+			count++
+		}
+	}
+
+	return count
+}
+
+// WarningsCount returns the number of Diagnostic in Diagnostics that are SeverityWarning.
+func (diags Diagnostics) WarningsCount() int {
```
detro, line 88: Nit: both those `*Count` methods could be implemented as a `len(Call_to_non_Count_methods)`.

detro (approved):
As said verbally, the only thing missing is a CHANGELOG entry and I left a nitpick/suggestion regarding the count methods.

🚀

diag/diagnostics.go
bendbennett: Good point. Fixed.

detro (approved):
🚀

diag/diagnostics_test.go
```diff
@@ -518,3 +519,123 @@ func TestDiagnosticsHasError(t *testing.T) {
...
+
+func TestDiagnosticsErrors(t *testing.T) {
+	t.Parallel()
+
+	type testCase struct {
+		diags    diag.Diagnostics
+		expected diag.Diagnostics
+	}
+	tests := map[string]testCase{
+		"errors": {
```
bflad, line 589: Nit: Might be good to check what happens when:

- There is a `nil` `diag.Diagnostics`, it should return nil or empty `diag.Diagnostics`
- There is an empty `diag.Diagnostics`, it should return empty `diag.Diagnostics`
- There are only warning diagnostics, it should return empty `diag.Diagnostics`

diag/diagnostics_test.go
```diff
@@ -518,3 +519,123 @@ func TestDiagnosticsHasError(t *testing.T) {
...
+
+func TestDiagnosticsWarnings(t *testing.T) {
+	t.Parallel()
+
+	type testCase struct {
+		diags    diag.Diagnostics
+		expected diag.Diagnostics
+	}
+	tests := map[string]testCase{
+		"errors": {
```
bflad, line 620: Nit: Might be good to check what happens when:

- There is a `nil` `diag.Diagnostics`, it should return nil or empty `diag.Diagnostics`
- There is an empty `diag.Diagnostics`, it should return empty `diag.Diagnostics`
- There are only error diagnostics, it should return empty `diag.Diagnostics`

diag/diagnostics_test.go
```diff
@@ -518,3 +519,123 @@ func TestDiagnosticsHasError(t *testing.T) {
...
+
+func TestDiagnosticsErrorsCount(t *testing.T) {
+	t.Parallel()
+
+	type testCase struct {
+		diags    diag.Diagnostics
+		expected int
+	}
+	tests := map[string]testCase{
+		"errors": {
```
bflad, line 531: Nit: Similar to the `Errors` testing, I think we should cover some additional cases just to ensure it returns `0` instead of panics, etc.

diag/diagnostics_test.go
```diff
@@ -518,3 +519,123 @@ func TestDiagnosticsHasError(t *testing.T) {
...
+
+func TestDiagnosticsWarningsCount(t *testing.T) {
+	t.Parallel()
+
+	type testCase struct {
+		diags    diag.Diagnostics
+		expected int
+	}
+	tests := map[string]testCase{
+		"errors": {
```
bflad, line 560: Nit: Similar to the `Warning` testing, I think we should cover some additional cases just to ensure it returns `0` instead of panics, etc.

bflad (approved):
Code-wise I think this looks good, although I'd love to see/hear about expected use cases before expanding the public API here.

bendbennett:
> Code-wise I think this looks good, although I'd love to see/hear about expected use cases before expanding the public API here.

One use-case is in the instance where we're running `Any()` with `All()` validation and we want to be able to return all warnings. For instance https://github.com/hashicorp/terraform-plugin-framework-validators/pull/43/files#r907476258

detro:
To add to the "use cases": it's useful when writing tests to count the expected errors, instead of just checking `HasErrors()`.

diag/diagnostics_test.go
bendbennett: Added additional test coverage.

diag/diagnostics_test.go
bendbennett: Added additional test coverage.

diag/diagnostics_test.go
bendbennett: Added additional test coverage.

diag/diagnostics_test.go
bendbennett: Added additional test coverage.

bflad (approved):
Looks good to me 🚀

PR #395, opened 2022-06-28
Update plan-modification.mdx

roidelapluie:
Sounds like IsUnknown is needed, not IsNull.

### DIFF
--- a/website/docs/plugin/framework/resources/plan-modification.mdx
+++ b/website/docs/plugin/framework/resources/plan-modification.mdx
@@ -93,7 +93,7 @@ func (m stringDefaultModifier) Modify(ctx context.Context, req tfsdk.ModifyAttri
         return
     }
 
-    if !str.Null {
+    if !str.IsUnknown() {
         return
     }
 


detro:
Hey @roidelapluie, thanks for this.

No, I think the documentation is correct: we need to check if the value is `Null`, not `Unknown`.

If the value _is null_, it means it's not set in the HCL TF configuration.
If the value _is unknown_, it means it's not known at the time of processing of the plan modifier, but it is indeed set to something (otherwise it would be _null_).

The difference is subtle but crucial.

I'm going to close this now.

roidelapluie:
This is my code: https://github.com/o11ydev/terraform-provider-zammad/blob/42debd0cc7a93ea96d4082084a6e3e5abdd713c1/zammad/resource_ticket_priority.go#L351

detro:
I think your issue has to do that you are using the Plan for a boolean there.

Can I suggest to look at this implementation of default value via plan modifier? https://github.com/hashicorp/terraform-provider-tls/blob/main/internal/provider/attribute_plan_modification/attribute_plan_modifiers.go

It's not released yet, but it's a generic implementation that should work with any type you need.

roidelapluie:
It works, thanks. Hopefully we get that upstream in the framework soon :-)

detro:
We are working on a few options, that implementation could be one of them.

Glad it worked out for your scenario :)

PR #396, opened 2022-06-28
path: Initial Path Expression Support

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/81
Reference: https://github.com/hashicorp/terraform-plugin-framework-validators/issues/14
Reference: https://github.com/hashicorp/terraform-plugin-framework-validators/issues/15
Reference: https://github.com/hashicorp/terraform-plugin-framework-validators/issues/16
Reference: https://github.com/hashicorp/terraform-plugin-framework-validators/issues/17
Reference: https://github.com/hashicorp/terraform-plugin-framework-validators/issues/20

This introduces the concept of root and relative attribute path expressions, abstractions on top of an attribute path, which enables provider developers to declare logic which might match zero, one, or more paths.

Paths are directly convertible into path expressions as exact expression steps. The builder-like syntax for exact expression steps matches the syntax for regular path steps, such as `AtName()` in both cases always represents an exact transversal into the attribute name of an object. Additional expression steps enable matching any list, map, or set element, such as `AtAnyListIndex()`. It also supports relative attribute path expressions, by supporting a parent expression step `AtParent()` and starting an expression with `MatchRelative()` so it can be combined with a prior path expression.

The framework will automatically expose path expressions to attribute plan modifiers and validators, so they can more intuitively support relative paths as inputs to their logic. For example, the `terraform-plugin-framework-validators` Go module will implement support for `terraform-plugin-sdk` multiple attribute schema behaviors such as `ConflictsWith`. It is expected that the downstream implementation can allow provider developers to declare the validator with expressions such as:

```go
tfsdk.Attribute{
	// ... other fields ...

	Validators: []AttributeValidators{
		schemavalidator.ConflictsWith(
			// Example absolute path from root
			path.MatchRoot("root_attribute"),

			// Example relative path from current attribute
			// e.g. another attribute at the same list index of ListNestedAttributes
			path.MatchRelative().AtParent().AtName("another_same_level_attribute"),
		),
	},
}
```

Then the logic within the validator can take the `ValidateAttributeRequest.AttributePathExpression` and use the `(path.Expression).Merge()` method to combine the current attribute expression with any incoming expressions.

To find matching attribute paths based on a path expression within `tfsdk.Config`, `tfsdk.Plan`, and `tfsdk.State`, a `PathMatches(path.Expression)` method has been added to each type. The resulting paths can then be used to fetch data via existing functionality, such as the `GetAttribute()` method of each type.


path/expression.go
```diff
@@ -0,0 +1,163 @@
+package path
+
+import (
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+)
+
+// Expression represents an attribute path with expression steps, which can
+// represent zero, one, or more actual Paths.
+type Expression struct {
```
detro, line 9: Given the description, I was expecting an `.Append()` method here.

I see there is one on the `ExpressionSteps`, but not on the expression itself.

The reason for me asking, is that I'm trying to imagine how we would take the input for a `schemavalidator`, append it to the `path.Expression` of the attribute, and get to the referred-to attributes.

detro (approved):
Code wise is 👨‍🍳 💋 

But I do have a "how do I use this?" question.

path/expression.go
bflad: Oh goodness, you are absolutely right that I omitted that method. It's important to be able to merge expressions, especially in plan modifiers and validators. I'll get that pushed up shortly.

bflad:
@detro I wound up adding the `(path.Expression).Merge(path.Expression) path.Expression` method and the `(tfsdk.Config).PathMatches(path.Expression) path.Paths`, `(tfsdk.Plan).PathMatches(path.Expression) path.Paths`, and `(tfsdk.State).PathMatches(path.Expression) path.Paths` methods, so this pull request should be more fully functional downstream for testing. Can you please take a fresh look? Thanks so much.

PR #401, opened 2022-07-07
path: Add Append methods for Paths and Expressions

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework-validators/pull/32

No additional CHANGELOG necessary as this package is still unreleased.


detro (approved):
thank you - LGTM

PR #402, opened 2022-07-07
diag: Added `Diagnostics` type `Equal()` method

### LINKED ISSUE #397: diag: Consider Equal() method for Diagnostics
### Module version

```
v0.9.0
```

### Use-cases

Downstream code may desire to do equality checking of `diag.Diagnostics`, such as unit testing responses from various framework functionality. In particular, 

Most Go modules that implement checking of this sort, such as https://pkg.go.dev/github.com/google/go-cmp, will prefer an `Equal()` method on a Go type if it exists. Otherwise they will fallback to other methods, such as type and field inspection, which are more difficult to manage (e.g. differences between `diag.ErrorDiagnostic`/`diag.WarningDiagnostic` and `diag.withPath`) in some cases are not possible to test outside the framework (`internal/reflect.DiagIntoIncompatibleType`).

### Attempted Solutions

Using go-cmp with `diag.Diagnostics` types and skipping unit testing where problematic.

### Proposal

The `diag.Diagnostic` interface already requires an `Equal()` method that handles severity, summary, and detail equality, so all that is necessary would be implementing an `Equal()` method on the `diag.Diagnostics` type alias as well.

### References

Not sure if this would affect #217.

bflad:
Closes #397

This should help simplify downstream code wishing use testing libraries that perform equality checks based on types that implement `Equal()` methods, such as [`go-cmp`](https://pkg.go.dev/github.com/google/go-cmp).

### DIFF
--- a/.changelog/pending.txt
+++ b/.changelog/pending.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+diag: Added `Diagnostics` type `Equal()` method
+```
--- a/diag/diagnostics.go
+++ b/diag/diagnostics.go
@@ -60,6 +60,22 @@ func (diags Diagnostics) Contains(in Diagnostic) bool {
 	return false
 }
 
+// Equal returns true if all given diagnostics are equivalent in order and
+// content, based on the underlying (Diagnostic).Equal() method of each.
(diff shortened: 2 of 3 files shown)


detro (approved):
LGTM

PR #410, opened 2022-07-08
tfsdk: Ensure Config, Plan, and State PathMatches will return null and unknown parent paths

bflad:
This will also now return an error diagnostic if a given path expression is invalid for the schema data.

The `(path.Expression).MergeExpressions(...path.Expression) Expressions` is a helper method for the common use case of combining an attribute path expression with an incoming collection of expressions to perform generic validation across attributes.

This requires no CHANGELOG entries as it is unreleased functionality.

### DIFF
--- a/tfsdk/config.go
+++ b/tfsdk/config.go
@@ -60,6 +60,10 @@ func (c Config) GetAttribute(ctx context.Context, path path.Path, target interfa
 }
 
 // PathMatches returns all matching path.Paths from the given path.Expression.
+//
+// If a parent path is null or unknown, which would prevent a full expression
+// from matching, the parent path is returned rather than no match to prevent
+// false positives.
 func (c Config) PathMatches(ctx context.Context, pathExpr path.Expression) (path.Paths, diag.Diagnostics) {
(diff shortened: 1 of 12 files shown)


detro (approved):

tfsdk/config.go
```diff
@@ -60,6 +60,10 @@ func (c Config) GetAttribute(ctx context.Context, path path.Path, target interfa
 }
 
 // PathMatches returns all matching path.Paths from the given path.Expression.
+//
+// If a parent path is null or unknown, which would prevent a full expression
```
detro, line 64: I think we need to clarify what `parent path` is. 

I'm reading this  method doc and pretending I don't know anything about the path libraries yet: the fact that I don't know what a `parent path` is, makes it difficult to understand the follow up `, which would prevent a full expression from matching, ...`

detro: I think understanding that, would also explain why returning the parent path is better than returning empty.

bflad: From out of band discussion, the `Path` and `Expression` Go documentation will be expanded in a followup pull request to start really talking about how they relate to schemas and the underlying data.

PR #413, opened 2022-07-14
path+website: Additional Go documentation and initial website documentation for paths and path expressions

### LINKED ISSUE #81: Support AttributePaths natively

bflad:
Closes #81

The website documentation is less than ideal as it introduces these new path and path expression pages as top-level in the already lengthy navigation, however refactoring the website documentation to bundle together and fix wording/flow for all the concepts/pages around schemas, schema-based data such as Config/Plan/State, and paths in a new dedicated section felt out of scope to this feature specific documentation addition.

### DIFF
--- a/path/expression.go
+++ b/path/expression.go
@@ -5,7 +5,43 @@ import (
 )
 
 // Expression represents an attribute path with expression steps, which can
-// represent zero, one, or more actual Paths.
+// represent zero, one, or more actual paths in schema data. This logic is
+// either based on an absolute path starting at the root of the schema data,
+// similar to Path, or a relative path which is intended to be merged with an
+// existing absolute path.
+//
+// Use the MatchRoot() function to create an Expression for an absolute path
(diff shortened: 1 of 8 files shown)


detro (approved):
🚀

PR #420, opened 2022-07-20
path: Ensure Expression root flag is appropriately set and checked

### LINKED ISSUE #419: Any call to `.AtName()`, `.AtListIndex()`, etc. removes the `root` flag on root expressions
### Module version

```
0.10.0
```

### Description

For a terraform configuration like:

```hcl
provider "tls" {
  proxy {
    // url: conflicting with `from_env`
    from_env = true
  }
}
```

If we need to express that `url` conflicts with `from_env`, we need to write in the schema of the provider something like:

```golang
  // using an "absolute path"
  schemavalidator.ConflictsWith(path.MatchRoot("proxy").AtName("from_env")),
  // using a "relative path"
  schemavalidator.ConflictsWith(path.MatchRelative().AtParent().AtName("from_env")),
}						},
```

Both approaches should work, but unfortunately the calls to `AtName()` (and any other `.At*` method on a `path.Expression`) make use `.Copy()`, and that call inadvertently modifies the expression so that, if internally it contained the flag `root: true`, by virtue of Golang default values, it gets turned to `root: false`:

```golang
func (e Expression) Copy() Expression {
	return Expression{
		steps: e.Steps().Copy(),
	}
}
```

Because of this, the provider validation logic triggers an error when using the "absolute path" approach:

```
Error: Invalid Path Expression for Schema Data
        
          with provider["registry.terraform.io/hashicorp/tls"],
          on terraform_plugin_test.tf line 2, in provider "tls":
           2: 					provider "tls" {
        
        The Terraform Provider unexpectedly matched no paths with the given path
        expression and current schema data. This can happen if the path expression
        does not correctly follow the schema in structure or types. Please report
        this to the provider developers.
        
        Path Expression: proxy.url.proxy.from_env
```

### Expected Behavior

Validation should be express-able via "absolute" as well as "relative" paths.
The choice should be left to the provider developer.

### Actual Behavior

Any "absolute" path expression (i.e. created via `.MatchRoot()`, i.e. that contain internally the `root: true` flag), when deeper than 1 level, assemble an incorrect expression, where the "root-ness" gets lost.

bflad:
Closes #419

Since the `(Expression).Equal()` method was not checking this additional flag, it was missing the fact it wasn't handled with `Copy()`. The `(Path).Expression()` method should always set that flag to true, since paths are always root path expressions.

### DIFF
--- a/.changelog/420.txt
+++ b/.changelog/420.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+path: Ensured `Expression` type `Copy()` method appropriately copied root expressions and `Equal()` checked for root versus relative expressions
+```
--- a/path/expression.go
+++ b/path/expression.go
@@ -139,12 +139,17 @@ func (e Expression) AtSetValue(value attr.Value) Expression {
 // affecting the original.
 func (e Expression) Copy() Expression {
 	return Expression{
+		root:  e.root,
 		steps: e.Steps().Copy(),
 	}
 }
 
 // Equal returns true if the given expression is exactly equivalent.
 func (e Expression) Equal(o Expression) bool {
+	if e.root != o.root {
+		return false
+	}
+
 	if e.steps == nil && o.steps == nil {
 		return true
 	}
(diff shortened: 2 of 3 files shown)


detro (approved):
LGTM

Thanks for the super fast turnaround

bflad:
Thanks for the excellent bug report and analysis!

PR #422, opened 2022-07-21
internal/fwserver: Implement Block MaxItems/MinItems validation for Terraform 0.12 through 0.15.1

### LINKED ISSUE #421: tfsdk: Validate Block MaxItems and MinItems for Terraform 0.12-0.15.1 Compatibility
### Module version

```
v0.10.0
```

### Use-cases

When implementing `tfsdk.Block` schema definitions with `MaxItems` and/or `MinItems` fields set, e.g.:

```go
tfsdk.Block{
  // other fields omitted for brevity
  MaxItems: 2,
  MinItems: 2,
}
```

The prior terraform-plugin-sdk implemented configuration validation logic for list and set blocks for these fields until Terraform version 0.15.2 introduced the proper validation through HCL's configuration decoding validation support. Terraform versions between 0.12.0 and 0.15.1 implemented a conservative validation approach, which generally relied on terraform-plugin-sdk catching this class of configuration issue after decoding the configuration and during the `Validate*Config` RPCs.

### Attempted Solutions

Manually using an attribute validator, such as [`listvalidator.SizeBetween()`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework-validators@v0.4.0/listvalidator#SizeBetween), in the block definition.

### Proposal

When performing "attribute" validation on blocks, re-implement similar fallback validation logic to check the list/set block element counts against the defined `MaxItems` and `MinItems` fields if they are set above 0. This logic must double check that each element is known (not null or unknown).

### References

- https://github.com/hashicorp/terraform-provider-tls/pull/242
- https://github.com/hashicorp/terraform/pull/28424
- https://github.com/hashicorp/terraform-plugin-sdk/blob/b61403bc822f962eee39212bb131de93cfe903ae/helper/schema/schema.go#L1919-L1936

bflad:
Closes #421

This implementation is similar to validation logic present in terraform-plugin-sdk, allowing providers to migrate using the same `MaxItems` and `MinItems` fields.

The framework could also remove these two fields in preference of providers using the `Validators` field for consistency with `Attribute` (e.g. providers could use the `github.com/hashicorp/terraform-plugin-framework-validators/listvalidator.Size*()` functions), however it is unclear at the time of writing whether the additional migration burden is worth the conceptual consistency between attribute and block definitions. That type of change can happen as a breaking change at any time before the framework reaches 1.0.


internal/fwserver/block_validation.go
```diff
@@ -153,3 +202,53 @@ func BlockValidate(ctx context.Context, b tfsdk.Block, req tfsdk.ValidateAttribu
...
+func blockMaxItemsDiagnostic(attrPath path.Path, maxItems int64, elements int) diag.Diagnostic {
+	var details strings.Builder
+
+	details.WriteString("The configuration should declare a maximum of ")
+
+	if maxItems == 1 {
+		details.WriteString("1 block")
+	} else {
+		details.WriteString(fmt.Sprintf("%d blocks", maxItems))
+	}
```
detro, line 215: Nit: Nothing against perfectionism, but maybe a `%d block(s)` would have been enough ;)

detro (approved):
I'm OK to add this backward-compatible solution.

But I would also see no harm nor issue if instead we went for enforcing Validators here:

```
  listvalidator.SizeBetween(0, 1),
}
```

vs

```
MinItems: 0,
MaxItems: 1,
```

Doesn't look too much of an effort to me to use the first one.

What needs to happen to rely on Validators instead?

detro:
OK, I have actually tries to use the list validator, and while it definitely works, it's way less "ergonomical" than having validation on the Block level.

For a schema where you can only have 0 or 1 of a block, I get this if the config contains 2:

```
Error: Invalid Attribute Value
        
          on terraform_plugin_test.tf line 5, in resource "tls_cert_request" "test":
           5: 					resource "tls_cert_request" "test" {
        
        List must contain at least 0 elements and at most 1 elements, got: 2
```

That is technically correct, but because it doesn't point the path to the Block field, it makes it hard to understand what's wrong and where.

So, I see this PR as an additive thing, and I'm even more 👍 now.

bflad:
I have a discussion open with the core team to decide the fate of `Block` `MaxItems`/`MinItems` a little more clearly, which should give us some guidance on whether we should retain them or paradigm shift to using solely configuration validators with this new framework. In the meantime, going to pull this in so at least _something_ is validating these block sizes on older Terraform versions and we can always remove it later if its extraneous.

PR #424, opened 2022-07-26
Update docs to reflect breaking changes in v0.10.0 for path.Path

larstobi:
* Use the new path abstraction #390


bflad (approved):
Thank you for fixing these, @larstobi, looks good to me 🚀

PR #425, opened 2022-07-26
Readme update - TF CLI compatibility

vravind1:
Plugin-framework supports both 5 and 6 versions of the Terraform Plugin Protocol.


bflad (approved):
Looks good to me 🚀

bflad:
(If necessary, we can get into the specifics of when protocol version 5 is supported, since it is not by default)

Issue #415, opened 2022-07-14
Unable to Read Previously Saved State for UpgradeResourceState TF 0.12


bendbennett:
### Module version

```
v0.9.0
```

### Relevant provider source code

```go
func TestAccResourceString_UpgradeFromVersion3_3_2(t *testing.T) {
	resource.Test(t, resource.TestCase{
		Steps: []resource.TestStep{
			{
				ExternalProviders: providerVersion332(),
				Config: `resource "random_string" "min" {
							length = 12
							override_special = "!#@"
							min_lower = 2
							min_upper = 3
							min_special = 1
							min_numeric = 4
						}`,
				Check: resource.ComposeTestCheckFunc(
					resource.TestCheckResourceAttrWith("random_string.min", "result", testCheckLen(12)),
					resource.TestMatchResourceAttr("random_string.min", "result", regexp.MustCompile(`([a-z].*){2,}`)),
					resource.TestMatchResourceAttr("random_string.min", "result", regexp.MustCompile(`([A-Z].*){3,}`)),
					resource.TestMatchResourceAttr("random_string.min", "result", regexp.MustCompile(`([0-9].*){4,}`)),
					resource.TestMatchResourceAttr("random_string.min", "result", regexp.MustCompile(`([!#@])`)),
					resource.TestCheckResourceAttr("random_string.min", "special", "true"),
					resource.TestCheckResourceAttr("random_string.min", "upper", "true"),
					resource.TestCheckResourceAttr("random_string.min", "lower", "true"),
					resource.TestCheckResourceAttr("random_string.min", "numeric", "true"),
					resource.TestCheckResourceAttr("random_string.min", "min_special", "1"),
					resource.TestCheckResourceAttr("random_string.min", "min_upper", "3"),
					resource.TestCheckResourceAttr("random_string.min", "min_lower", "2"),
					resource.TestCheckResourceAttr("random_string.min", "min_numeric", "4"),
				),
			},
			{
				ProtoV5ProviderFactories: protoV5ProviderFactories(),
				Config: `resource "random_string" "min" {
							length = 12
							override_special = "!#@"
							min_lower = 2
							min_upper = 3
							min_special = 1
							min_numeric = 4
						}`,
				PlanOnly: true,
			},
			{
				ProtoV5ProviderFactories: protoV5ProviderFactories(),
				Config: `resource "random_string" "min" {
							length = 12
							override_special = "!#@"
							min_lower = 2
							min_upper = 3
[...]

PR #426, opened 2022-07-27
Using UnmarshalWithOpts in order to ignore undefined attributes

bendbennett:
Closes: #415 

During unmarshalling of JSON, if there are fields which do not have a corresponding attribute in the schema an error is returned.

A [PR](https://github.com/hashicorp/terraform-plugin-go/pull/213) on terraform-plugin-go addresses this issue.

The `UnmarshalWithOpts` function used in this PR will not be available to `terraform-plugin-framework` until the PR on `terraform-plugin-go` has been merged, tagged and released. Consequently, the build will fail on this PR until that time.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -66,6 +66,8 @@ github.com/hashicorp/go-uuid v1.0.3/go.mod h1:6SBZvOh/SIDV7/2o3Jml5SYk/TvGqwFJ/b
 github.com/hashicorp/go-version v1.2.0/go.mod h1:fltr4n8CU8Ke44wwGCBoEymUuxUHl09ZGVZPK5anwXA=
 github.com/hashicorp/terraform-plugin-go v0.12.0 h1:6wW9mT1dSs0Xq4LR6HXj1heQ5ovr5GxXNJwkErZzpJw=
 github.com/hashicorp/terraform-plugin-go v0.12.0/go.mod h1:kwhmaWHNDvT1B3QiSJdAtrB/D4RaKSY/v3r2BuoWK4M=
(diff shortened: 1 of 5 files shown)


bflad (commented):
Changes here looks good -- could you add a CHANGELOG entry file? Maybe something like

``````markdown
```release-note:bug
internal/fwserver: Ensured `UpgradeResourceState` calls from Terraform 0.12 properly ignored attributes not defined in the schema
```
``````

**Comment by bflad** [comment]
I think we'll also need to update the `req.RawState.Unmarshal` call later down in the upgrade state handlers as well (line 150 currently) and add a covering unit test for that situation.

File: go.sum [hunk 1]
```diff
@@ -66,6 +66,8 @@ github.com/hashicorp/go-uuid v1.0.3/go.mod h1:6SBZvOh/SIDV7/2o3Jml5SYk/TvGqwFJ/b
 github.com/hashicorp/go-version v1.2.0/go.mod h1:fltr4n8CU8Ke44wwGCBoEymUuxUHl09ZGVZPK5anwXA=
 github.com/hashicorp/terraform-plugin-go v0.12.0 h1:6wW9mT1dSs0Xq4LR6HXj1heQ5ovr5GxXNJwkErZzpJw=
```
bflad, line 67: This line should've been removed -- can you double check you called `go mod tidy`?

bflad (approved):
Looks really good! A `go mod tidy` and this should be good to merge. 🚀

go.sum
bendbennett: My bad. Done.

detro (approved):

bflad:
Merging so I can update and merge in some other things. 😄

PR #429, opened 2022-07-27
tfsdk: Clarify Attribute and Block type DeprecationMessage field usage

### LINKED ISSUE #428: tfsdk: Clarify DeprecationMessage Field Usage
### Module version

```
0.10.0
```

### Use-cases

Currently, the [`tfsdk.Attribute` type `DeprecationMessage` field](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/tfsdk#Attribute.DeprecationMessage) is a message that gets translated into a warning diagnostic during Terraform's validation phase:

```
Warning: Attribute Deprecated

{DeprecationMessage field string}
```

This handling however, due to limitations in the protocol, only occurs for `Required` or `Optional` attributes when they have a configuration value (known or unknown in the case of the framework). There is currently no effect when the attribute is `Computed`-only (read-only). The framework already enables the `deprecated` boolean field available in the protocol, so if/when Terraform core does something with that protocol field in a future version, it'll have an effect on read-only attributes then.

### Proposal

Update the Go documentation and website documentation to ensure it denotes this limitation. We could go further and begin failing unit testing for attributes with `DeprecationMessage` and only `Computed` via #113, however that feature can be treated separately, if desired.

### References

- https://github.com/hashicorp/terraform/issues/7569

bflad:
Closes #428

### DIFF
--- a/tfsdk/attribute.go
+++ b/tfsdk/attribute.go
@@ -73,9 +73,36 @@ type Attribute struct {
 	// file is sensitive.
 	Sensitive bool
 
-	// DeprecationMessage defines a message to display to practitioners
-	// using this attribute, warning them that it is deprecated and
-	// instructing them on what upgrade steps to take.
+	// DeprecationMessage defines warning diagnostic details to display to
+	// practitioners configuring this Attribute. The warning diagnostic summary
+	// is automatically set to "Attribute Deprecated" along with configuration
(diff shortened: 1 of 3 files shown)


detro (approved):

bendbennett (approved):
LGTM

Issue #399, opened 2022-07-06
Resource Private State Management


bflad:
### Module version

```
v0.9.0
```

### Use-cases

Terraform supports the concept of private state storage for resources, which generally is an advanced use case for provider development. Similar to terraform-plugin-sdk, the terraform-plugin-framework Go module currently does not have support for accessing or managing a resource's private state.

Private state data is byte data stored in the Terraform state and is intended for provider usage only (in the "anything on the provider side of the protocol" sense; so any SDK and actual provider code). It is considered opaque to Terraform core, as it is never accessed or exposed in plans, however it must be passed to and from Terraform core for the persistent storage of the data. Providers have the ability to save this data during create, import, planning, read, and update operations and the ability to read this data during delete, planning, read, and update operations.

Providers wishing to work with private state data must take care to copy any relevant saved data that is sent in a RPC request to the RPC response, as appropriate for the use case.

This functionality is a pre-cursor to supporting terraform-plugin-sdk/v2's resource timeout functionality, which is described in a separate GitHub issue: #400 

### Attempted Solutions

Managing everything in normal resource state, which will show up in plans.

### Proposal

Enable provider developers to read/write private state in a framework managed `map[string][]byte` of data in the appropriate request and response types of the `Create`, `Read`, `Update`, `Delete`, `ImportState`, and `ModifyPlan` methods associated with a `tfsdk.Resource`. For the `Read`, `Update`, and `ModifyPlan` methods, the framework can automatically copy any private state data in the request to the response, to prevent potential accidental data loss. Providers can opt to explicitly `nil` out a key, if necessary to remove data.

Provider developers can choose which implementation will maintain a portion of the private state data in the map based on the string key. The framework will validate and restrict the provider definable map keys to ensure that the framework has its own "namespace" in the future if necessary. The framework will ultimately manage the final private state data with a JSON encoding when interfacing with terraform-plugin-go and the protocol, where the data is the merging of the provider defined map data and any potential framework defined map data.

The framework will be responsible for the following validations, returning error diagnostics as appropriate:

* Incoming terraform-plugin-go `[]byte` data is valid JSON, and can be unmarshalled as `map[string]any`
* Provider defined map keys are valid
* Provider defined `[]byte` data is valid JSON and UTF-8 safe
* Outgoing terraform-plugin-go `[]byte` data is valid JSON and UTF-8 safe

[...]

PR #432, opened 2022-07-29
tfsdk: Migrate data source, provider, and resource types into new Go packages

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/132
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/365
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/366

Since the framework's initial release, the `tfsdk` Go package has been a monolithic collection of various provider concepts, including but not limited to handling of data sources, providers, resources, schemas, schema data (such as `Config`, `Plan`, and `State`), and various other helpers. For framework maintainers, this monolithic package has made implementations difficult for certain enhancements, such as import cycles. For provider developers, this monolithic package has been difficult to navigate in the Go documentation. This change represents the first major provider coding paradigm shift to colocate related concepts into their own Go packages, starting with new top-level `datasource`, `provider`, and `resource` packages.

This particular change and method (no deprecation period) was not desired, but it was unavoidable due to the interconnectedness of the `tfsdk` package and due to the amount of effort it would require to attempt to support both the old and new types. This change is necessary to complete before there are additional code compatibility promises added to this Go module, such as a version 1.0.0 release or release candidate.

This type of change is not taken lightly, as it is quite disruptive for existing provider codebases. On the surface, this change may look fairly unbeneficial for provider developers other than the easier discoverability and reduced wordiness, however it is required to unblock other future refactoring and enhancement efforts in the framework. It is unclear at the time of writing whether splitting out the other concepts from the `tfsdk` package, such as schemas, will present the same issues. Regardless, any other major changes will be spread across releases.

Provider developers should be able to update their code using find and replace operations using the tables below.

To reduce framework maintainer review burden, all code migrations were lift and shift operations while most code and documentation updates were find and replace operations.

| Prior tfsdk Package Type | New provider Package Type |
| --- | --- |
| `tfsdk.ConfigureProviderRequest` | `provider.ConfigureRequest` |
| `tfsdk.ConfigureProviderResponse` | `provider.ConfigureResponse` |
| `tfsdk.Provider` | `provider.Provider` |
| `tfsdk.ProviderConfigValidator` | `provider.ConfigValidator` |
| `tfsdk.ProviderWithConfigValidators` | `provider.ProviderWithConfigValidators` |
| `tfsdk.ProviderWithProviderMeta` | `provider.ProviderWithMetaSchema` (note naming realignment) |
[...]

PR #433, opened 2022-08-01
Implement Resource Private State Management

bendbennett:
Closes: #399


internal/privatestate/data.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	return bytes, diags
+}
+
+// NewData creates a new Data based on the given slice of bytes.
+// It must be a JSON encoded slice of bytes, that is map[string][]byte.
+func NewData(ctx context.Context, data []byte) (Data, diag.Diagnostics) {
+	var (
+		u     map[string][]byte
```
bflad, line 46: Nit: On first read I wasn't sure what `u` was here -- should we call it something like "dataMap"? 👍

internal/privatestate/data.go
```diff
@@ -0,0 +1,163 @@
...
+
+		return diags
+	}
+
+	if !json.Valid(value) {
+		diags.AddError("JSON Invalid",
+			"Values stored in private state must be valid JSON")
+
+		return diags
+	}
```
bflad, line 136: Should these errors mention that the problem is with the provider and to contact the provider developer?

internal/privatestate/data.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+// ValidateProviderDataKey determines whether the key supplied is allowed on the basis of any
+// restrictions that are in place, such as key prefixes that are reserved for use with
+// framework private state data.
+func ValidateProviderDataKey(ctx context.Context, key string) diag.Diagnostics {
+	if isInvalidProviderDataKey(ctx, key) {
+		return diag.Diagnostics{
+			diag.NewErrorDiagnostic(
+				"Restricted Namespace",
```
bflad, line 150: Nit:

```suggestion
				"Restricted Resource Private State Namespace",
```

internal/privatestate/data.go
```diff
@@ -0,0 +1,163 @@
...
+				"Using a period ('.') as a prefix for a key used in private state is not allowed",
```
bflad, line 151: Similar question about the details mentioning this is likely a provider code issue 😄

internal/privatestate/data_test.go
```diff
@@ -0,0 +1,307 @@
...
+		Provider: map[string][]byte{
+			"providerKeyOne": []byte("provider value one"),
+			"providerKeyTwo": []byte("provider value two"),
+		},
+	})
+
+	testCases := map[string]struct {
+		data          privatestate.Data
+		expected      []byte
+		expectedDiags diag.Diagnostics
```
bflad, line 48: It'd be great to have a test case that returns an error diagnostic, maybe something like bytes that don't successfully convert to a valid JSON string?

internal/privatestate/data.go
```diff
@@ -0,0 +1,163 @@
+package privatestate
```
bflad, line 1: Let's add a `doc.go` file with a brief Go documentation comment for the package 👍

bflad (commented):
Great start! I'm not sure if you're interested in trying the next steps using this new package or potentially waiting until #432 is a little more sorted since it'll be moving the types where this data will wind up. Whatever works for me (if you'd rather do the bulk of the other work with an understanding you may need to rebase for the `tfsdk` package changes). 😄

internal/privatestate/data.go
```diff
@@ -0,0 +1,163 @@
...
+	// Provider contains private state data for provider usage.
+	Provider ProviderData
+}
+
+// Bytes returns a JSON encoded slice of bytes containing the merged
+// framework and provider private state data.
+func (d Data) Bytes(_ context.Context) ([]byte, diag.Diagnostics) {
+	var diags diag.Diagnostics
+
+	bytes, err := json.Marshal(d)
```
bflad, line 28: I think we may need to merge the framework and provider maps before marshalling directly here, so we don't wind up with top-level JSON keys of `Framework` and `Provider` (e.g. we should ensure we can roundtrip this data).

internal/privatestate/data_test.go
bendbennett: Believe we'll need to introduce a check in the `Bytes` func that determines whether each of the `[]byte` that is associated with each key is valid JSON (i.e., `json.Valid`) in order to be able to add that test. Is that what you had in mind?

bflad: Ah ha, you can technically convince `json.Marshal()` itself to error on invalid map key types, e.g., https://go.dev/play/p/CFA4MqvyoPw

```go
json.Marshal(
	map[interface{}]struct{}{
		big.NewFloat(123): {},
	},
)
```

However our usage of `map[string][]byte` for storage prevents that particular issue. I think we're good here. 😄

internal/privatestate/data.go
```diff
@@ -0,0 +1,178 @@
...
+	diags := ValidateProviderDataKey(ctx, key)
+
+	if diags.HasError() {
+		return diags
+	}
+
+	if !utf8.Valid(value) {
+		diags.AddError("UTF-8 Invalid",
+			"Values stored in private state must be valid UTF-8\n\n"+
+				"This is always a problem with Terraform or terraform-plugin-framework. Please report this to the provider developer.",
```
bflad, line 137: Since this data provided from the provider developer in their codebase, should this messaging be adjusted? It might also be good to include the key that is causing the error and maybe a log message with the value for troubleshooting.

internal/privatestate/data.go
```diff
@@ -0,0 +1,178 @@
...
+		)
+
+		return diags
+	}
+
+	if !json.Valid(value) {
+		diags.AddError("JSON Invalid",
+			"Values stored in private state must be valid JSON\n\n"+
+				"This is always a problem with Terraform or terraform-plugin-framework. Please report this to the provider developer.",
```
bflad, line 146: Similar to above.

internal/privatestate/data.go
```diff
@@ -0,0 +1,178 @@
...
+// ValidateProviderDataKey determines whether the key supplied is allowed on the basis of any
+// restrictions that are in place, such as key prefixes that are reserved for use with
+// framework private state data.
+func ValidateProviderDataKey(ctx context.Context, key string) diag.Diagnostics {
+	if isInvalidProviderDataKey(ctx, key) {
+		return diag.Diagnostics{
+			diag.NewErrorDiagnostic(
+				"Restricted Resource Private State Namespace",
+				"Using a period ('.') as a prefix for a key used in private state is not allowed\n\n"+
+					"This is always a problem with Terraform or terraform-plugin-framework. Please report this to the provider developer.",
```
bflad, line 166: Similar to above (minus the value logging part 😉 )

internal/privatestate/data.go
bflad: We could also consider adding the value to the error diagnostic, however I think we should be conservative with the potential to output sensitive data.

bendbennett: D'oh! Good point.

internal/privatestate/data.go
```diff
@@ -0,0 +1,200 @@
...
+//
+// If the key is reserved for framework usage, an error diagnostic
+// is returned. If the key is valid, but private state data is not found,
+// nil is returned.
+//
+// The naming of keys only matters in context of a single resource,
+// however care should be taken that any historical keys are not reused
+// without accounting for older resource instances that may still have
+// older data at the key.
+func (d *ProviderData) GetKey(ctx context.Context, key string) ([]byte, diag.Diagnostics) {
```
bendbennett, line 111: Does `GetKey` need to be declared on a pointer receiver given that `ProviderData` is a `map[string][]byte`?

internal/privatestate/data.go
```diff
@@ -0,0 +1,200 @@
...
+//
+// If the key is reserved for framework usage, an error diagnostic
+// is returned. The data must be valid JSON and UTF-8 safe or an error
+// diagnostic is returned.
+//
+// The naming of keys only matters in context of a single resource,
+// however care should be taken that any historical keys are not reused
+// without accounting for older resource instances that may still have
+// older data at the key.
+func (d *ProviderData) SetKey(ctx context.Context, key string, value []byte) diag.Diagnostics {
```
bendbennett, line 140: Does `SetKey` need to be declared on a pointer receiver given that `ProviderData` is a `map[string][]byte`?

internal/privatestate/data.go
bflad: It's probably fine either which way -- read-only methods shouldn't need pointer receiver handling unless handling a `nil` pointer of that type is important, such as panic prevention.

internal/privatestate/data.go
bflad: My best recommendation would be to write unit testing that verifies the expected map after a key has been set, e.g. setting a new key should verify that the new key was added or updating an existing key should verify that the value was updated. Both you and future code travelers can understand and verify the expected behaviors via that unit testing. 👍

internal/proto5server/server_readresource_test.go
```diff
@@ -172,6 +174,43 @@ func TestServerReadResource(t *testing.T) {
...
+					Provider: &testprovider.Provider{
+						GetResourcesMethod: func(_ context.Context) (map[string]provider.ResourceType, diag.Diagnostics) {
+							return map[string]provider.ResourceType{
+								"test_resource": &testprovider.ResourceType{
+									GetSchemaMethod: func(_ context.Context) (tfsdk.Schema, diag.Diagnostics) {
+										return tfsdk.Schema{}, nil
+									},
+									NewResourceMethod: func(_ context.Context, _ provider.Provider) (resource.Resource, diag.Diagnostics) {
+										return &testprovider.Resource{
+											ReadMethod: func(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {},
```
bflad, line 189: One thing that I find very valuable during these RPC-based request field tests is validating that the provider logic (what would be inside this `ReadMethod` here) gets the expected value from the request field or raise an error diagnostic that would show up eventually as a test failure (since the diagnostic would be unexpected).

For example here:

```suggestion
											ReadMethod: func(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
												expected := "provider value"
												got, diags := req.PrivateState.GetKey("providerKey")

												resp.Diagnostics.Append(diags...)

												if string(got) != expected {
													resp.Diagnostics.AddError(
														"Unexpected req.Private Value",
														fmt.Sprintf("expected %q, got %q", expected, got),
													)
												}
											},
```

internal/proto5server/server_readresource_test.go
```diff
@@ -172,6 +174,43 @@ func TestServerReadResource(t *testing.T) {
 				NewState: testEmptyDynamicValue,
 			},
 		},
+		"request-private": {
```
bflad, line 177: We should also create a "response-private" test, similar to the below, but it calls `resp.Private.SetKey()` 👍

internal/proto6server/server_readresource_test.go
```diff
@@ -172,6 +174,43 @@ func TestServerReadResource(t *testing.T) {
 				NewState: testEmptyDynamicValue,
 			},
 		},
+		"request-private": {
```
bflad, line 177: Similarly with "response-private" testing.

internal/proto6server/server_readresource_test.go
```diff
@@ -172,6 +174,43 @@ func TestServerReadResource(t *testing.T) {
...
+					Provider: &testprovider.Provider{
+						GetResourcesMethod: func(_ context.Context) (map[string]provider.ResourceType, diag.Diagnostics) {
+							return map[string]provider.ResourceType{
+								"test_resource": &testprovider.ResourceType{
+									GetSchemaMethod: func(_ context.Context) (tfsdk.Schema, diag.Diagnostics) {
+										return tfsdk.Schema{}, nil
+									},
+									NewResourceMethod: func(_ context.Context, _ provider.Provider) (resource.Resource, diag.Diagnostics) {
+										return &testprovider.Resource{
+											ReadMethod: func(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {},
```
bflad, line 189: Similarly with GetKey testing

internal/fwserver/server_readresource.go
```diff
@@ -71,10 +72,20 @@ func (s *Server) ReadResource(ctx context.Context, req *ReadResourceRequest, res
...
+
 	logging.FrameworkDebug(ctx, "Calling provider defined Resource Read")
 	resourceImpl.Read(ctx, readReq, &readResp)
 	logging.FrameworkDebug(ctx, "Called provider defined Resource Read")
 
 	resp.Diagnostics = readResp.Diagnostics
 	resp.NewState = &readResp.State
+
+	if readResp.Private != nil {
+		resp.Private.Provider = readResp.Private
```
bflad, line 89: We may need to instantiate `resp.Private` when there was no `req.Private` -- it'd be great to have unit testing for that scenario.

resource/read.go
```diff
@@ -14,6 +15,9 @@ type ReadRequest struct {
 	// operation.
 	State tfsdk.State
 
+	// Private is resource private state data.
```
bflad, line 18:
```suggestion
	// Private is provider-defined resource private state data which was previously
	// stored with the resource state. This data is opaque to Terraform and does
	// not affect plan output. Any existing data is copied to
	// ReadResourceResponse.Private to prevent accidental private state data loss.
	//
	// Use the GetKey method to read data. Use the SetKey method on
	// ReadResourceResponse.Private to update or remove a value.
```

internal/fwserver/server_readresource.go
bendbennett: Good shout. Blows up with a nil pointer otherwise :grimacing:

internal/fwserver/server_updateresource.go
```diff
@@ -91,6 +93,32 @@ func (s *Server) UpdateResource(ctx context.Context, req *UpdateResourceRequest,
 		updateReq.ProviderMeta = *req.ProviderMeta
 	}
 
+	if req.PlannedPrivate != nil {
+		updateReq.Private = req.PlannedPrivate.Provider
+		updateResp.Private = req.PlannedPrivate.Provider
+		resp.Private = req.PlannedPrivate
+	}
+
```
bflad, line 101: I think this logic is extraneous, given the below logic

```suggestion
```

resource/modify_plan.go
```diff
@@ -26,6 +27,9 @@ type ModifyPlanRequest struct {
 
 	// ProviderMeta is metadata from the provider_meta block of the module.
 	ProviderMeta tfsdk.Config
+
+	// Private is provider private state data.
```
bflad, line 31: Nit: Maybe we should say this is "resource" private state data, mention it will pre-populate the response data, and how to work with it? `ReadRequest`/`ReadResponse` has some expanded verbiage we can borrow.

resource/modify_plan.go
```diff
@@ -42,6 +46,10 @@ type ModifyPlanResponse struct {
 	// recreated.
 	RequiresReplace path.Paths
 
+	// Private is provider private state data following potential modifications
+	// by provider developer.
```
bflad, line 50: Nit: Similar here about mentioning the pre-populating

resource/update.go
```diff
@@ -37,6 +42,11 @@ type UpdateResponse struct {
 	// should be set during the resource's Update operation.
 	State tfsdk.State
 
+	// Private is the private state resource data following the Update operation.
+	// This field is not pre-populated as there is no pre-existing private state
+	// data at the time a resource is created.
```
bflad, line 47: Update should be pre-populated 😄

tfsdk/attribute_plan_modification.go
```diff
@@ -84,6 +85,9 @@ type ModifyAttributePlanRequest struct {
 
 	// ProviderMeta is metadata from the provider_meta block of the module.
 	ProviderMeta Config
+
+	// Private is provider private state data.
```
bflad, line 89: Nit: Similar note here about expanding the verbiage for provider developers. 👍

bflad (commented):
Overall this is looking really good, just a few minor things we can tidy up. If you throw in a changelog file, I can review that too. Thanks!

internal/fwserver/server_updateresource.go
bendbennett: Good catch! I meant to remove that.

internal/fwserver/attribute_plan_modification.go
```diff
@@ -122,6 +131,7 @@ func AttributeModifyPlan(ctx context.Context, a fwschema.Attribute, req tfsdk.Mo
 					Plan:          resp.Plan,
 					ProviderMeta:  req.ProviderMeta,
 					State:         req.State,
+					Private:       privateProviderData,
```
bflad, line 134: Open question: I wonder if these should be using the _updated_ private state data from the `AttributeWithPlanModifiers` handling above?

I think we may want that to make it slightly less confusing. It can use the `resp.Private` in these places here and below.

Either which way, I think we should document the behavior in the `ModifyAttributePlanRequest`/`ModifyAttributePlanResponse` fields to say that they either the private state is always the original data when the schema-based plan modification began or that its updated as the logic traverses deeper into underlying attributes.

internal/fwserver/block_plan_modification.go
```diff
@@ -101,6 +110,7 @@ func BlockModifyPlan(ctx context.Context, b fwschema.Block, req tfsdk.ModifyAttr
 					Plan:          resp.Plan,
 					ProviderMeta:  req.ProviderMeta,
 					State:         req.State,
+					Private:       privateProviderData,
```
bflad, line 113: Similarly here about using the updated private state data.

internal/fwserver/attribute_plan_modification.go
```diff
@@ -49,11 +50,18 @@ func AttributeModifyPlan(ctx context.Context, a fwschema.Attribute, req tfsdk.Mo
 
 	var requiresReplace bool
 
+	privateProviderData := privatestate.EmptyProviderData(ctx)
+
+	if req.Private != nil {
+		privateProviderData = req.Private
+	}
```
bflad, line 57: Does this need to be copied to `resp.Private` to prevent data loss? Unfortunately difficult to unit test currently since `tfsdk.Attribute` implements `fwxschema.AttributeWithPlanModifiers`, but we'll have other attribute implementations 🔜

internal/fwserver/server_planresourcechange.go
```diff
@@ -196,9 +217,10 @@ func (s *Server) PlanResourceChange(ctx context.Context, req *PlanResourceChange
 		logging.FrameworkTrace(ctx, "Resource implements ResourceWithModifyPlan")
 
 		modifyPlanReq := resource.ModifyPlanRequest{
-			Config: *req.Config,
-			Plan:   stateToPlan(*resp.PlannedState),
-			State:  *req.PriorState,
+			Config:  *req.Config,
+			Plan:    stateToPlan(*resp.PlannedState),
+			State:   *req.PriorState,
+			Private: privateProviderData,
```
bflad, line 223: Similarly here regarding using the updated private state data from the schema handling above -- this can be unit tested by implementing both schema and resource level modifications where the resource performs a conditional update based on the result of a schema change.

internal/fwserver/server_planresourcechange.go
```diff
@@ -209,6 +231,7 @@ func (s *Server) PlanResourceChange(ctx context.Context, req *PlanResourceChange
 			Diagnostics:     resp.Diagnostics,
 			Plan:            modifyPlanReq.Plan,
 			RequiresReplace: path.Paths{},
+			Private:         privateProviderData,
```
bflad, line 234: I've found that setting the request up correctly then copying the request field into the response makes the logic in these a little easier to follow

```suggestion
			Private:         modifyPlanReq.Private,
```

internal/privatestate/data.go
```diff
@@ -0,0 +1,363 @@
...
+
+	return &output, diags
+}
+
+// NewProviderData creates a new ProviderData based on the given slice of bytes.
+// It must be a JSON encoded slice of bytes, that is map[string][]byte.
+func NewProviderData(ctx context.Context, data []byte) (*ProviderData, diag.Diagnostics) {
+	providerData := &ProviderData{
+		data: make(map[string][]byte),
+	}
```
bflad, line 179: World's smallest nit in case the logic is ever more complicated 😎 

```suggestion
	providerData := EmptyData(ctx)
```

resource/update.go
```diff
@@ -37,6 +46,15 @@ type UpdateResponse struct {
...
+
```
bflad, line 52: Duplicate comment 😎 

```suggestion
```

bflad (commented):
Almost there! Those new Empty/Must functions make things so much easier to read, nice!

internal/fwserver/attribute_plan_modification.go
bflad: One way to unit test this would be to have an attribute implementing nested attributes and a plan modifier that updates the parent attribute while one of the nested attributes has a plan modifier that is conditional on that update.

bflad (approved):
Looks good to me 🚀

Issue #366, opened 2022-06-08
Consider Refactoring tfsdk Config/Plan/State Logic into internal Package(s)


bflad:
### Module version

```
v0.8.0
```

### Use-cases

While working through some of the framework server refactoring as part of #215, it was necessary to extract some of the `tfsdk.Config`, `tfsdk.Plan`, and `tfsdk.State` unexported methods into temporary functions within the `internal/fwserver` package so they did not become part of the public `tfsdk` APIs. For example, `ConfigGetAttributeValue()`. This highlighted that a larger portion of the "schema data" handling would benefit as being in internal packages and colocated again, rather than splitting the logic across packages, while still exporting provider developer functionality only as necessary in `tfsdk` or some other "public" package.

### Proposal

To be written up, but a current thought is that there should be an `internal/schemadata` or similar package which contains the full framework schema-based data handling, while `tfsdk` or another package merely exposes anything that should be in the public API. Ideally, there would be a single implementation for readable and writable data that is shared across the Config/Plan/State implementations.

### References

- #215

bflad:
This `internal` package refactoring was completed previously. Other issues can be used to track additional efforts.

Issue #132, opened 2021-09-07
Consider Splitting tfsdk.Attribute into datasource.Attribute, provider.Attribute, and resource.Attribute


bflad:
### Module version

Post plan modification support

### Use-cases

Plan modifications can only occur on resource attributes. Since the `Attribute` type is shared across data source, provider, and resource schemas, it is possible to define plan modifiers on data source and provider attributes which will have no effect. This can be confusing for provider developers.

### Attempted Solutions

Note behavior in Go documentation.

### Proposal

Consider creating separate Go types for data source, provider, and resource attributes which can cause compile-time errors for functionality not supported by each type. e.g.

```go
package datasource
type Attribute struct{
  // ... Type, etc. ....
}

package provider
type Attribute struct{
  // ... Type, etc. ....
}

package resource
type Attribute struct{
  // ... Type, etc. ....
  PlanModifiers []AttributePlanModifier
}
```

### References

- https://github.com/hashicorp/terraform-plugin-framework/pull/123/files#r702095995

paddycarver:
I am gently suspicious that the fix will make things confusing (nested attributes can accept provider attributes inside a data source attribute?) but I think it's an interesting line of inquiry.

PR #434, opened 2022-08-02
tfsdk: Migrate `RequiresReplace()`, `RequiresReplaceIf()`, and `UseStateForUnknown()` plan modifier functions to the `resource` package

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/pull/432

Plan modification in the framework is a concept that only applies to managed resources via the protocol `PlanResourceChange` RPC. Following the migration of other `tfsdk` package types into separate `datasource`, `provider`, and `resource` packages, this change migrates the `RequiresReplace()`, `RequiresReplaceIf()` and `UseStateForUnknown()` plan modifier functions into the `resource` package. This change should be bundled in the same release as the package refactoring since it is similar in nature.

This change has two immediate benefits:

- Aligning the Go package placement to these functions o hint to provider developers that these are only intended for managed resources.
- For future refactoring this removes additional schema-based code imports from the `tfsdk` package, which could enable refactoring the schema logic into internal packages with potentially less breaking changes for provider developers by removing a source of import cycles.

Provider developers should be able to update their code using find and replace operations using the table below:

| Prior tfsdk Package Function | New resource Package Function |
| --- | --- |
| `tfsdk.RequiresReplace` | `resource.RequiresReplace` |
| `tfsdk.RequiresReplaceIf` | `resource.RequiresReplaceIf` |
| `tfsdk.UseStateForUnknown` | `resource.UseStateForUnknown` |

To reduce framework maintainer review burden, this code migrations was primarily a lift and shift operation while most code and documentation updates were find and replace operations. The modifier types were unexported as exposing them should not be beneficial for provider developers and should help reduce the Go documentation surface area. The plan modifier unit testing was updated to use a `_test` package so it is verifying only exported functionality.


bflad:
terraform-provider-corner changes: https://github.com/hashicorp/terraform-provider-corner/pull/83

bflad:
Just to explicitly say it out loud since I didn't mention in the original description, we cannot move `tfsdk.AttributePlanModifier` to `resource.AttributePlanModifier` yet until #132 happens, otherwise we'd have an import cycle.

bendbennett (approved):
LGTM

PR #436, opened 2022-08-02
all: Go package documentation updates post datasource, provider, and resource package migration

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/pull/432


bendbennett (approved):
The expanded comments really provide an excellent improvement in clarity.
LGTM

Issue #389, opened 2022-06-21
Consider Whether Attribute (Or Subset) Should be Exposed in ModifyAttributePlanRequest


bflad:
### Module version

```
v0.9.0
```

### Use-cases

Certain generic attribute plan modifiers will desire performing actions based on the `Attribute` definition itself. A current use case is `RequiresReplace` and `RequiresReplaceIf`, which currently have some checks for the `Computed` field to preserve some terraform-plugin-sdk/v2 behavior that was present with `ForceNew`.

### Attempted Solutions

Using `req.State.Schema.AttributeAtPath(req.AttributePath)` to extract the `Attribute`. This will require logic updates post #81, as the underlying implementation will remain `*tftypes.AttributePath` until #365 is done.

### Proposal

First, we should decide _if_ the top level `Attribute` definition should be exposed. This would be a brittle implementation detail for framework compatibility and provider developers may make unintended decisions based on exposing everything. The benefit however, is that provider developers do not need to jump through hoops to get the information as the framework would handle all the nitty gritty.

Most (all?) plan modifiers shouldn't need that type of information though, except potentially `Computed` if they are logically treating that as special for legacy reasons. Another potential future `Attribute` field is `Default attr.Value`. If implemented, plan modifiers will most certainly warrant access to this value.

Therefore an initial proposal here would be to only expose targeted `Attribute` fields, such as `Computed` or optionally `Default` if that becomes implemented. It would likely be best to let some of the referenced issues shake out first though, before making any implementation decisions.

### References

- #31 (whether `Attribute` will continue to have `Computed` field)
- #81 (initial impetus)
- #285 (consideration if `Attribute` may introduce `Default`-like field)
- #365 (could change things yet again)

bflad:
The new resource schema plan modifier for RequiresReplace now does not perform the `Computed` check. A RequiresReplaceIfConfigured variant is available. As for other potential use cases of attribute-specific information, it is generally recommended to not base plan modifier implementations on those attribute-specific details (e.g. whether or not something is `Computed` or not, should not be used). If there are other valid use cases out there, please raise a new issue.

Issue #223, opened 2021-11-03
Consider Interface for Current (Schema).AttributeTypeAtPath() Handling


bflad:
### Module version

```
v0.4.2
```

### Use-cases

To potentially refactor type handling better:

```go
// AttributeTypeAtPath returns the attr.Type of the attribute at the given path.
func (s Schema) AttributeTypeAtPath(path *tftypes.AttributePath) (attr.Type, error) {
	rawType, remaining, err := tftypes.WalkAttributePath(s, path)
	if err != nil {
		return nil, fmt.Errorf("%v still remains in the path: %w", remaining, err)
	}

	switch typ := rawType.(type) {
	case attr.Type:
		return typ, nil
	case nestedAttributes:
		return typ.AttributeType(), nil
	case nestedBlock:
		return typ.Block.attributeType(), nil
	case Attribute:
		return typ.attributeType(), nil
	case Block:
		return typ.attributeType(), nil
	case Schema:
		return typ.AttributeType(), nil
	default:
		return nil, fmt.Errorf("got unexpected type %T", rawType)
	}
}
```

### References

- https://github.com/hashicorp/terraform-plugin-framework/pull/188/files#r741846034

bflad:
This is sort of being handled with #365 and the like.

PR #438, opened 2022-08-05
tfsdk: Migrate schema handling to internal interfaces

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/31
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/223
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/326
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/389

The main goals of this change are:

- Prepare the Go module to support multiple packages that implement concept-specific schema declarations, in particular the `datasource`, `provider`, and `resource` packages
- Continue supporting the existing `tfsdk` package schema implementation with minimal provider developer breaking changes, allowing a deprecation period when the concept-specific schemas are introduced
- Migrate unexported or unintentially exported `tfsdk` schema functionality to internal packages

These goals are accomplished by creating new `internal/fwschema` and `internal/fwxschema` packages, which contain interface types that the provider developer facing packages implement. Currently, the `tfsdk` package is the only implementation, until the design of those concept-specific schema declarations is fully determined. Those designs may include changes to schema implementation details, which cannot be accomplished in the existing `tfsdk` implementation without breaking it and provider developers migrating code to the new packages is a great time to make those types of changes.

The internal interface method design here is purposefully similar to the existing `tfsdk` implementations as we cannot name the methods the same as existing fields and to reduce review burden. After the `tfsdk` implementations are removed, we can consider tidying up the interface methods to drop the `Get` and `Is` method name prefixes.

There are some minor followup changes that will happen either between or during concept-specific schema implementation work, namely:

- Swapping calls `tfsdk` package type `TerraformType()` methods for `AttributeType().TerraformType()` to reduce implementation work for new schema implementations
- Updating unit testing that relies on `tfsdk.Schema` to set up sub-tests via `(*testing.T).Run()` to against multiple implementations

These were not included here to reduce the review burden as they are separate details which can be handled later.


bendbennett (approved):
Wow! Nice job.

bflad:
Thank you for the quick review, @bendbennett -- I think this will be the last v0.11.0 breaking change for refactoring, since there's already a few things queued up. I'll try to get to the two followup items I mention above sometime soon so our v0.12.0 can focus on the actual concept-specific schema design and the enhancements we can do with those.

PR #440, opened 2022-08-08
tfsdk: Additional preparations for multiple schema implementations

bflad:

This is a followup to the initial schema logic migration that tries to further prepare the Go module for multiple schema implementations.

- Removes unexported schema-based functions/methods from the `tfsdk` package (`pathMatches` will be handled as part of schema data internalization)
- Removes `tftypes.Type` handling from schemas as `(attr.Type).TerraformType()` already does this
- Tries to standardize on `Type()` for fetching framework type (`attr.Type`) information from schema types (`Attribute` will get fixed once `tfsdk.Attribute` is removed)
- Tries to standardize `Schema` method signatures for framework types/diagnostics and terraform-plugin-go types/errors
- Prepares `NestedAttributes` implementations for unit testing across multiple attribute implementations


bendbennett (approved):
You are a refactoring machine!
LGTM

PR #451, opened 2022-08-12
internal/fwschemadata: Initial package

bflad:

This change creates the initial `internal/fwschemadata` package, which will hold the shared implementation details of schema-based data, such as the existing `tfsdk.Config`, `tfsdk.Plan`, and `tfsdk.State` types and potentially other implementations in the future. This refactoring will be gradual over time to reduce review burden and should not introduce provider developer changes until potentially at the end of the refactoring.

The main interaction point will be the `Data` type, which currently contains the schema and terraform-plugin-go value information. This type may be expanded in the future to support diagnostic message customization (e.g. denoting the data is a "configuration") and eventually could support value migration to a framework native type (e.g. `types.Object`).

The initial functionality refactored is:

- Fetch and reflect entire value (`Get()` method)
- Path existence (`PathExists()` method)
- Path matching (`PathMatches()` method)
- terraform-plugin-go value extraction at a given terraform-plugin-go path (`TerraformValueAtTerraformPath()` method)

The unit testing is meant to be more exhaustive than what was previously performed in the `tfsdk` package.


bendbennett (approved):
LGTM

PR #452, opened 2022-08-12
internal/privatestate: Handle terraform-plugin-sdk private state in NewData

bflad:
Previously, errors such as these could be returned:

```
Error: Error Decoding Private State

          with random_password.min,
          on terraform_plugin_test.tf line 1, in resource "random_password" "min":
           1: resource "random_password" "min" {

        An error was encountered when decoding private state: illegal base64 data at
        input byte 0.

        This is always a problem with Terraform or terraform-plugin-framework. Please
        report this to the provider developer.
```

The new unit test case would fail with the same error prior to this fix.


bendbennett (approved):
LGTM 🚀

PR #453, opened 2022-08-14
internal/fwschemadata: Migrate tfsdk GetAttribute and getAttributeValue logic

bflad:

This change continues the migration of `tfsdk.Config`/`tfsdk.Plan`/`tfsdk.State` type logic into the `internal/fwschemadata` package `Data` type, so the implementation is shared and to potentially introduce other implementations.


bendbennett (approved):
LGTM

Issue #456, opened 2022-08-19
Computed attrs are not set to unknown after a AttributePlanModifier causes a non-empty plan


goncalo-rodrigues:
### Module version

v0.11.1

### Relevant provider source code

Resource has a computed field (e.g. timestamp) that might be changed upon any update. Terraform normally sets it to unknown on any plan (correctly) if there's anything to update.

```
tfsdk.Schema{
    "timestamp": {
        ....
	Computed:    true,
    },
    "name": {
        ....
	Required:    true,
    },
    "status": {
       	Computed:      true,
	PlanModifiers: []tfsdk.AttributePlanModifier{modifiers.ResourceStatusModifier{}},
    }
}
```

However, if the plan is changed from an empty plan to a required update due to an AttributePlanModifier, it is no longer set to unknown.

```
func (m ResourceStatusModifier) Modify(ctx context.Context, req tfsdk.ModifyAttributePlanRequest, resp *tfsdk.ModifyAttributePlanResponse) {
	resp.AttributePlan = types.String{Value: "Running"}
}
```

### Inconsistent planning

If update is triggered by a user change to a config, we get this plan:

```
resource test "test" {
    timestamp = "2022-08-19T10:29:57+00:00" -> (known after apply)
    name = "test1" -> "test2"
    status = "STOPPED" -> "RUNNING"
}

```

If instead the user changed nothing, but a plan is still created due to the AttributePlanModifier, we get:

```
resource test "test" {
    status = "STOPPED" -> "RUNNING"
}

```

which generates the error after the apply is successful:

╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to test.test, provider "provider[\"hashicorp.com/dev/multy\"]" produced an unexpected new value: timestamp: was cty.StringVal("2022-08-19T10:29:57+00:00"), but now cty.StringVal("2022-08-19T10:33:51+00:00").
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.

So, my question is: why does this differ in these 2 cases? Is this a known bug? Is there a workaround?

bflad:
Hi @goncalo-rodrigues 👋 Thank you for raising this and sorry this caused you some trouble.

The behavior you mention is an artifact of the existing internal implementation of resource planning in the framework, where the following steps are executed if the resource is not planned for destruction:

[...]

Issue #183, opened 2021-09-28
Plan Modifiers Not Executed Before Plan Computed-Only Unknown Marking


bflad:
### Module version

Followup to #181 fix.

### Relevant framework source code

```go
// In tfsdk.TestServerPlanResourceChange:
		"attr_plan_modifiers_trigger_computed_unknown": {
			resource:     "test_attribute_plan_modifiers",
			resourceType: testServeResourceTypeAttributePlanModifiersType,
			priorState: tftypes.NewValue(testServeResourceTypeAttributePlanModifiersType, map[string]tftypes.Value{
				"computed_string_no_modifiers": tftypes.NewValue(tftypes.String, "statevalue"),
				"name":                         tftypes.NewValue(tftypes.String, "TESTATTRONE"),
				"size":                         tftypes.NewValue(tftypes.Number, 3),
				"scratch_disk": tftypes.NewValue(tftypes.Object{AttributeTypes: map[string]tftypes.Type{
					"id":        tftypes.String,
					"interface": tftypes.String,
					"filesystem": tftypes.Object{AttributeTypes: map[string]tftypes.Type{
						"size":   tftypes.Number,
						"format": tftypes.String,
					}},
				}}, map[string]tftypes.Value{
					"id":        tftypes.NewValue(tftypes.String, "my-scr-disk"),
					"interface": tftypes.NewValue(tftypes.String, "scsi"),
					"filesystem": tftypes.NewValue(tftypes.Object{AttributeTypes: map[string]tftypes.Type{
						"size":   tftypes.Number,
						"format": tftypes.String,
					}}, nil),
				}),
				"region": tftypes.NewValue(tftypes.String, "DEFAULTVALUE"),
			}),
			config: tftypes.NewValue(testServeResourceTypeAttributePlanModifiersType, map[string]tftypes.Value{
				"computed_string_no_modifiers": tftypes.NewValue(tftypes.String, nil),
				"name":                         tftypes.NewValue(tftypes.String, "TESTATTRONE"),
				"size":                         tftypes.NewValue(tftypes.Number, 3),
				"scratch_disk": tftypes.NewValue(tftypes.Object{AttributeTypes: map[string]tftypes.Type{
					"id":        tftypes.String,
					"interface": tftypes.String,
					"filesystem": tftypes.Object{AttributeTypes: map[string]tftypes.Type{
						"size":   tftypes.Number,
						"format": tftypes.String,
					}},
[...]

PR #463, opened 2022-08-29
internal/fwserver: Add unit test for no configuration change plan which has modified and unmodified Computed attributes

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/183
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/456


bendbennett (approved):
LGTM

PR #464, opened 2022-08-29
internal/fwschemadata: Replace fwserver duplicate logic, add data description

bflad:

This change represents another evolutionary step to move away from `tfsdk.Config`, `tfsdk.Plan`, and `tfsdk.State` types directly, where the temporarily duplicated logic in `internal/fwserver` which worked directly with those types is replaced with the shared data implementation.


bendbennett (approved):
LGTM

PR #468, opened 2022-09-07
internal/fwserver: Use existing data instead of path-based lookups during plan modification

### LINKED ISSUE #293: Issues with Nested Set Attributes with PlanModifiers
### Module version

```
github.com/hashicorp/terraform-plugin-framework v0.6.1
```

```go
			"members": {
				Description: "The members of the group",
				Optional:    true,
				Computed:    true,
				Attributes: tfsdk.SetNestedAttributes(map[string]tfsdk.Attribute{
					"email": {
						Description: "...",
						Type:     types.StringType,
						Required: true,
					},
					"role": {
						Description: "...",
						Type:     types.StringType,
						Optional: true,
						Computed: true,
							DefaultModifier{
								DefaultVal: "MEMBER",
								ValType:    types.StringType,
							},
						},
							stringInSliceValidator{
								stringOptions: []string{"MANAGER", "MEMBER", "OWNER"},
							},
						},
					},
					"type": {
						Description: "...",
						Type:     types.StringType,
						Optional: true,
						Computed: true,
							DefaultModifier{
								DefaultVal: "USER",
								ValType:    types.StringType,
							},
						},
							stringInSliceValidator{
								stringOptions: []string{"CUSTOMER", "GROUP", "USER"},
							},
						},
					},
					"delivery_settings": {
						Description: "...",
						Type:     types.StringType,
						Optional: true,
						Computed: true,
							DefaultModifier{
								DefaultVal: deliverySettingsDefault,
								ValType:    types.StringType,
							},
						},
							stringInSliceValidator{
								stringOptions: []string{"ALL_MAIL", "DAILY", "DIGEST", "DISABLED", "NONE"},
							},
						},
					},
					"status": {
						Description: "Status of member.",
						Computed:    true,
					},
					"member_id": {
						Description: "...",
						Computed:    true,
					},
				}, tfsdk.SetNestedAttributesOptions{}),
			},
```

```hcl
resource "googleworkspace_group_members" "my-group-members" {
 group_id = googleworkspace_group.my-group.id

  members = [{
    email = googleworkspace_user.my-new-user.primary_email
  }]
}
```

### Expected Behavior

When the Default Modifier runs it checks if the attribute is `Unknown` and all the nested attributes should be `Unknown` in this case, since they aren't set in the config.

### Actual Behavior

All the attributes are coming into the plan modifier like so: `{Value: , Unknown: false, Null: true}` so the default doesn't get set.

### Steps to Reproduce

terraform apply

### References

Digging into it I see this is where it's happening:
[When planning each attribute elem](https://github.com/hashicorp/terraform-plugin-framework/blob/main/tfsdk/attribute.go#L615) >
[Get the attribute plan](https://github.com/hashicorp/terraform-plugin-framework/blob/fa1b0205433af2849a791a6cabbc44e5749fbed9/tfsdk/attribute.go#L507) >
[Get the terraform value at path](https://github.com/hashicorp/terraform-plugin-framework/blob/fa1b0205433af2849a791a6cabbc44e5749fbed9/tfsdk/plan.go#L78) >
[Walk the path](https://github.com/hashicorp/terraform-plugin-framework/blob/fa1b0205433af2849a791a6cabbc44e5749fbed9/tfsdk/plan.go#L320) > 
[Apply Terraform 5 Path Step](https://github.com/hashicorp/terraform-plugin-go/blob/main/tftypes/value.go#L197)

That's where I got stuck, I know the `len(diffs)` should equal `0` but it equals 1

bflad:
Closes #293

This change is a major rewrite of the plan modification logic to move away from globally mutating the schema response to instead localize handling of each attribute/block. The logic still transverses downward in the schema while updated plan modifier values are passed up the call stack to eventually overwrite the top level attribute/block data. The main benefit of this logic is that it cannot lose track of which element to update within a set and may have some nominal performance improvements.

The majority of the logic changes are convincing the plan modification process that nested attributes and blocks have specific types for their data, which allows it to safely traverse deeper based on the schema definition. There is just a lot of repeated diagnostics handling.

The new `TestAttributeModifyPlan/attribute-set-nested-usestateforunknown` and `TestBlockModifyPlan/block-set-nested-usestateforunknown` unit test cases specifically test a similar issue to the linked one where sets previously could not properly use `resource.UseStateForKnown()` for set nested attributes or set blocks.


bflad:
Taking a fresh look at this PR this morning, I think this is something we do want to do for now. I'm really sorry for the awful Git differences with the testing code, but essentially the changes were to replace `Config`/`Plan`/`State` fields with `AttributeConfig`/`AttributePlan`/`AttributeState` and remove any test cases dealing with "previous errors". The "previous errors" testing was there before because all response diagnostics were passed to all plan modifiers before. We don't have a known use case for that behavior and even if we did, the behavior would be completely inconsistent as attribute plan modifiers run in arbitrary ordering (not something we likely want to fix/adjust).

internal/fwserver/attr_value.go
```diff
@@ -0,0 +1,178 @@
...
+	return m, nil
+}
+
+func coerceObjectValue(schemaPath path.Path, value attr.Value) (types.Object, diag.Diagnostics) {
+	nullObject := types.Object{Null: true}
+
+	object, ok := value.(types.Object)
+
+	if !ok {
+		return nullObject, diag.Diagnostics{
```
bendbennett, line 44:
```suggestion
	object, ok := value.(types.Object)

	if !ok {
		return types.Object{Null: true}, diag.Diagnostics{
```

Tiny nit for consistency.

bendbennett (approved):
LGTM

PR #472, opened 2022-09-08
provider: Deprecate `DataSourceType`, `ResourceType`, `Provider` type `GetDataSources`, and `Provider` type `GetResources`

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/441
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/442

The goal of this change is twofold:

- Deprecate the separate `DataSourceType` and `ResourceType` types.
- Enable `DataSource` and `Resource` to be self-documenting with respects to their type name.

Removing `DataSourceType` and `ResourceType` simplifies provider implementations and gives developers the choice of whether or not to configure extra data in their `DataSource` and `Resource` implementations by adding the optional `Configure` method. This data is no longer required to be the `provider.Provider` itself, but rather can be any type such as a provider-defined structure or vendor-supplied client type.

Enabling `DataSource` and `Resource` type name handling within the implementation allows provider developers to further split out their data source and resource code across package boundaries without repeated map addition logic.

Given this framework version 0.11.1 code (which generally is split across Go files/packages, but shown altogether here for brevity):

```go
var _ provider.Provider = &ExampleProvider{}

type ExampleProvider struct{
  configured bool
  // ... additional remote client fields, etc.
}

func (p ExampleProvider) GetDataSources(ctx context.Context) (map[string]provider.DataSourceType, diag.Diagnostics) {
	return map[string]provider.DataSourceType{
		"example_thing": ThingDataSourceType{},
	}, nil
}

func (p ExampleProvider) GetResources(ctx context.Context) (map[string]provider.ResourceType, diag.Diagnostics) {
	return map[string]provider.ResourceType{
		"example_thing": ThingResourceType{},
	}, nil
}

func (p ExampleProvider) GetSchema(ctx context.Context) (tfsdk.Schema, diag.Diagnostics) { /* ... */ }

func (p *ExampleProvider) Configure(ctx context.Context, req provider.ConfigureRequest, resp *provider.ConfigureResponse) {
  // ... remote client setup, etc.
  p.configured = true
}

var _ provider.DataSourceType = ThingDataSourceType{}
var _ datasource.DataSource = ThingDataSource{}

type ThingDataSourceType struct{}

type ThingDataSource struct{
  provider ExampleProvider
}

func (t ThingDataSourceType) GetSchema(ctx context.Context) (tfsdk.Schema, diag.Diagnostics) { /* ... */ }

[...]

PR #473, opened 2022-09-09
website: Migration guide updates for provider.DataSourceType and provider.ResourceType deprecation

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/pull/472

Note: The source code reference links to the http, random, and tls providers will be out of date until they are migrated after this functionality is released in the framework. I'm guessing we should hold off merging this until after they are migrated?


bendbennett:
> Reference: #472
> 
> Note: The source code reference links to the http, random, and tls providers will be out of date until they are migrated after this functionality is released in the framework. I'm guessing we should hold off merging this until after they are migrated?

We could update the migration docs to remove source code links and replace with generic code examples?

website/docs/plugin/framework/migrating/data-sources/index.mdx
```diff
@@ -45,48 +45,49 @@ schema.Resource {
...
 
-The `GetSchema` function returns a `tfsdk.Schema` struct which defines your data source's attributes. This is the same
+The `GetSchema` method returns a `tfsdk.Schema` struct which defines your data source's attributes. This is the same
 struct you use to define provider and resource attributes.
 
-The `NewResource` function returns a type that you define. The type implements the `resource.Resource` interface,
-including the CRUD functions for your resource.
+The `TypeName` method returns a type name that you define.
+
+For `Read` method implements the logic for writing into the Terraform state.
```
bendbennett, line 71:
```suggestion
The `Read` method implements the logic for writing into the Terraform state.
```

bendbennett (approved):
LGTM

bflad:
I think I caught all the links pointing to code that was updated.

> We could update the migration docs to remove source code links and replace with generic code examples?

I'm generally a proponent of this approach as we can show code with only the necessary context and without needing to update links while the framework is under development, but that can be an exercise done separately. 👍

PR #474, opened 2022-09-09
datasource+resource: Rename TypeName method to Metadata

bflad:
For consistency with the provider `Metadata` method naming and to future-proof against unknown data source or resource use cases, such as specifying generic "groupings" for resources. One con is that it may be less obvious where to set the type name until the method is required.

If you do not agree with this change, please reach out. My preference would be to include this with the first release that deprecates `provider.DataSourceType`/`provider.ResourceType` so provider developers do not need to make two changes in this area.


bendbennett (approved):
LGTM

PR #475, opened 2022-09-09
internal/fwserver: Prevent two sources of errors/panics

bflad:
Reference: #468 (unreleased bugs)

Previously, it was possible to get a `Attribute Plan Modification Walk Error` diagnostic when a resource was being created and the `req.AttributeState` was set to `nil` instead of a null value for the given `attr.Value`. This changes the logic to always return a proper `attr.Value`.

Previously, it was possible that while iterating through the plan elements for a list or set collection, that configuration or state elements at the same index may not exist. The logic was off-by-one due to Go's 0-based slice indexing to return early with a null value in those cases.


bendbennett (approved):
LGTM

PR #477, opened 2022-09-12
tfsdk: Add support for single nested mode blocks

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/400

This support is required for provider developers to migrate from terraform-plugin-sdk to terraform-plugin-framework without affecting practitioner configurations or requiring state upgrades.

For example:

```terraform
resource "examplecloud_thing" "example" {
  timeouts {
    create = "5m"
  }
}
```

Provider developers are encouraged to use single nested attributes where possible though, as Terraform's single nested mode block support has only been thoroughly tested in production with the terraform-plugin-sdk timeouts block and its associated handling.


bendbennett (approved):
LGTM

Issue #467, opened 2022-09-04
xattr.TypeWithValidate support for list elements (and possibly contents of other collections)


Doridian:
### Module version
```
github.com/hashicorp/terraform-plugin-framework v0.11.1
```

### Use-cases
I want to make a list that validates its contents, such as a list of IP addresses

### Attempted Solutions
I created an entirely separate type called `ValidatedList`. This does work, but it prevents anything that is used to dealing with `types.List` (such as pre-made validators) from interacting with those lists.

See my currently working solutin (albeit without the build-in validators I'd love to be able to use): https://github.com/Doridian/terraform-provider-hexonet/blob/main/hexonet/utils/validated_list.go

Another solution would be to use `listvalidator.ValuesAre`, but if the value creator (in the case of IPs for example) fails by creating a `nil` value (such as un-parseable IP), then this cannot work.

### Proposal
Implement `xattr.TypeWithValidate`'s `Validate` function in lists (and possibly other collections)
Add type coercion to collection types and if their element type is a `xattr.TypeWithValidate`, run its `Validate` function and actr accordingly.

### References
I could not find any prior issue referring to `xattr.TypeWithValidate` in collections.

PR #481, opened 2022-09-14
Add Validate() to all collections

Doridian:
This also fixes a bug in `internal/reflect/map.go` where it used the map's own type to Validate elements, which does not seem correct at all.

Fixes https://github.com/hashicorp/terraform-plugin-framework/issues/467

### DIFF
--- a/types/map.go
+++ b/types/map.go
@@ -133,6 +133,7 @@ func (m MapType) Validate(ctx context.Context, in tftypes.Value, path path.Path)
 			"Map Type Validation Error",
 			"An unexpected error was encountered trying to validate an attribute value. This is always an error in the provider. Please report the following to the provider developer:\n\n"+err.Error(),
 		)
+		return diags
 	}
 
 	if !in.IsKnown() || in.IsNull() {
--- a/types/map.go
+++ b/types/map.go
@@ -121,6 +121,9 @@ func (m MapType) String() string {
 // xattr.TypeWithValidate.
 func (m MapType) Validate(ctx context.Context, in tftypes.Value, path path.Path) diag.Diagnostics {
 	var diags diag.Diagnostics
+	if true {
+		return diags
+	}
 
 	if in.Type() == nil {
 		return diags
--- a/types/map.go
+++ b/types/map.go
@@ -150,8 +150,8 @@ func (m MapType) Validate(ctx context.Context, in tftypes.Value, path path.Path)
 		return diags
 	}
 
-	validatableType, ok := m.ElemType.(xattr.TypeWithValidate)
-	if !ok {
+	validatableType, isValidatable := m.ElemType.(xattr.TypeWithValidate)
+	if !isValidatable {
 		return diags
 	}
 
--- a/types/map.go
+++ b/types/map.go
@@ -7,8 +7,10 @@ import (
 	"strings"
 
 	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/attr/xattr"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/internal/reflect"
+	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-go/tftypes"
 )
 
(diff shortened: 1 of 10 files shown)


types/set.go
```diff
@@ -148,15 +148,23 @@ func (st SetType) Validate(ctx context.Context, in tftypes.Value, path path.Path
 		return diags
 	}
 
+	validatableType, isValidatable := st.ElemType.(xattr.TypeWithValidate)
```
bflad, line 151: We should ensure there is a covering unit test for this change. 👍

types/map.go
```diff
@@ -115,6 +117,54 @@ func (m MapType) String() string {
...
+// Validate validates all elements of the map that are of type
+// xattr.TypeWithValidate.
+func (m MapType) Validate(ctx context.Context, in tftypes.Value, path path.Path) diag.Diagnostics {
+	var diags diag.Diagnostics
+
+	if in.Type() == nil {
+		return diags
+	}
+
+	if !in.Type().Is(tftypes.Map{}) {
```
bflad, line 129: We should ensure there is a covering unit test for this new conditional. 👍

bflad (changes requested):
Hi @Doridian 👋 Thank you for submitting this. Overall this is looking pretty good, it could just use some additional unit testing and small adjustments. If you have any questions about adding that testing, please reach out.

Also, if you could, it would be great to include a changelog entry which can be done by adding a `.changelog/481.txt` file with contents such as:

``````markdown
```release-note:bug
types: Ensured `List`, `Map`, and `Set` types with `xattr.TypeWithValidate` elements run validation on those elements
```
``````

**Comment by Doridian** [comment]
It seems that `AtSetValue` expects an `attr.Value` which is not present in `Validate`. Pushed a fix for that.

**Comment by bflad** [comment]
@Doridian since we are adding a method onto the type, we can unit test the method itself, e.g.

```go
func TestListTypeValidate(t *testing.T) {
	t.Parallel()

	testCases := map[string]struct {
		listType      ListType
		tfValue       tftypes.Value
		path          path.Path
		expectedDiags diag.Diagnostics
	}{
		"wrong-value-type": {
			listType: ListType{
				ElemType: StringType,
			},
			tfValue: tftypes.NewValue(tftypes.Set{
				ElementType: tftypes.String,
			}, []tftypes.Value{
				tftypes.NewValue(tftypes.String, "testvalue"),
			}),
			path: path.Root("test"),
			expectedDiags: diag.Diagnostics{
				diag.NewAttributeErrorDiagnostic(
					path.Root("test"),
					"List Type Validation Error",
					"An unexpected error was encountered trying to validate an attribute value. This is always an error in the provider. Please report the following to the provider developer:\n\n"+
						"expected List value, received tftypes.Value with value: tftypes.Set[tftypes.String]<tftypes.String<\"testvalue\">>",
				),
			},
		},
		"no-validation": {
			listType: ListType{
				ElemType: StringType,
			},
			tfValue: tftypes.NewValue(tftypes.List{
				ElementType: tftypes.String,
			}, []tftypes.Value{
				tftypes.NewValue(tftypes.String, "testvalue"),
			}),
			path: path.Root("test"),
		},
	}

	for name, testCase := range testCases {
		name, testCase := name, testCase

			t.Parallel()

			diags := testCase.listType.Validate(context.Background(), testCase.tfValue, testCase.path)

			if diff := cmp.Diff(diags, testCase.expectedDiags); diff != "" {
				t.Errorf("unexpected diagnostics difference: %s", diff)
			}
		})
	}
}
```

Rinse and repeat for `MapType` and `SetType`.

This is okay for now as it would _unfortunately_ take a lot more effort to import the existing `internal/testing/types.StringTypeWithValidateError` to verify the element validation diagnostics are passed back, as importing that right now would cause an import cycle and updating the testing file to `package types_test` and putting `types.` in front of everything in the `types` package is a lot to ask. 😅

bflad:
(We could also create an unexported type with validation errors in the test files too, but it is totally fine.)

Doridian:
I have added tests for list and map exactly as you described.
For map, as it already was validatable, I have just appended the `wrong-value-type` test case.

Doridian:
> This is okay for now as it would _unfortunately_ take a lot more effort to import the existing `internal/testing/types.StringTypeWithValidateError` to verify the element validation diagnostics are passed back, as importing that right now would cause an import cycle and updating the testing file to `package types_test` and putting `types.` in front of everything in the `types` package is a lot to ask. 😅

Maybe in another PR if I find the time 😉

bflad (approved):
This looks great, @Doridian, thank you so much for your efforts here 🚀

PR #497, opened 2022-09-23
attr: Add ValueType method to Type interface

### LINKED ISSUE #496: attr: Consider Adding ValueType Method to Type Interface
### Module version

```
v0.13.0
```

### Use-cases

When the internal reflection package is attempting to build values or convert them into given types, it needs to raise errors when the given target type/value is incompatible. Today, these errors can tell provider developers what went wrong, but not necessarily offer help for how to resolve it.

An example diagnostic that we'd like to raise:

```text
Received unknown value, however the target type cannot handle unknown values. Use the corresponding `types` package type or a custom type that handles unknown values.

Path: example
Target Type: *bool
Suggested Type: types.Bool <-- this (but also working for any attr.Value)
```

### Attempted Solutions

It's technically possible to try to get the associated `attr.Value` of an `attr.Type` by doing something like:

```go
attrValue, err := attrType.ValueFromTerraform(ctx, tftypes.NewValue(attrType.TerraformType(ctx), nil))
```

However, supporting null values in an `attr.Type` is not a requirement, as edge case-y as that may be, since in all implementations may be `Required` in the schema.

### Proposal

Consider introducing a new method in the `attr.Type` interface:

```go
type Type interface {
  // ... existing methods ...

  // ValueType should return the attr.Value type returned by ValueFromTerraform.
  // The returned attr.Value can be any null, unknown, or known value for the type,
  // as this is only intended for type detection.
  ValueType(context.Context) attr.Value
}
```

Or, make it optional, where the reflection package can pick it up if it is detected:

```go
type TypeWithValueType interface {
  // ValueType should return the attr.Value type returned by ValueFromTerraform.
  // The returned attr.Value can be any null, unknown, or known value for the type,
  // as this is only intended for type detection.
  ValueType(context.Context) attr.Value
}
```

### References

- #191
- #495

bflad:
~Pending #495~
Closes #496

This will allow the internal reflection package to return better error diagnostics.


bendbennett (approved):
LGTM
Quickest turnaround on an issue ever?!

bflad:
Latest commit shows the updated error details 😄

PR #502, opened 2022-09-29
types: Deprecate Attrs, AttrTypes, Elems, ElemTypes, Null, Unknown, and Value fields

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/447

When the framework type system was originally being developed, the value types were introduced with exported fields which also served as the internal details of whether a value was null, unknown, or a known value of a friendlier Go type. It was known that there was the potential for issues, but the simplified developer experience seemed to outweigh the potential for developer issues. Fast forward a few months, this decision appears to have two consequences that the framework maintainers hear about across various forums.

One issue is that the value types directly expose their internal implementation details and support the three states of a Terraform type value: being null, unknown, or a known value. Only one state should ever be set, but provider developers can make a value that is any combination of those states. This makes the framework behavior potentially indeterminate from the provider developer perspective whether, for example, a null AND unknown value becomes null OR unknown as it works its way through the framework.

```go
type ThingResourceModel struct{
  Computed types.String `tfsdk:"computed"`
}

func (r ThingResource) Create(ctx context.Context, req resource.CreateResource, resp *resource.CreateResponse) {
  var data ThingResourceModel

  tflog.Trace(ctx, "Data Values", map[string]any{
    // Unknown value: types.String{Null: false, Unknown: true, Value: ""}
    "computed": plan.Computed,
  })

  // Maybe some external API responses here, but showing hardcoded updates for
  // brevity. This will make the value invalid by enabling Null without
  // disabling Unknown.
  data.Computed.Null = true

  tflog.Trace(ctx, "Data Values", map[string]any{
    // Invalid value: types.String{Null: true, Unknown: true, Value: ""}
    "computed": data.Computed,
  })

  // The invalid value will be either null or unknown, depending on the
  // type implementation. If unknown, Terraform will error, since unknown
  // values are never allowed in state.
}
```

Another issue is that it is possible to create collection value types that do not match their type definition. This issue is especially more likely with `types.Object` where it is possible to accidentially omit attributes. While the framework handling would eventually catch this issue when dealing with the invalid value, it can be caught sooner.

```go
// Invalid value (missing attribute and differing attribute name)
types.Object{
    "one": types.StringType,
    "two": types.BoolType,
  },
  Attrs: map[string]attr.Value{
    "not_one": types.String{Value: "wrong name"},
  },
}
```

Another issue is that the default (zero-value) state for an "unset" value type turns into a known value, which is confusing since these values explicitly support being null. This causes Terraform errors which would surface to practitioners (especially when untested) that provider developers then have to troubleshoot the error message containing Terraform's type system details, potentially discover the reason why it is happening by looking at the framework type source code, then figure out a workable solution. It's not intuitive.

```go
type ThingResourceModel struct{
  // let's assume this is left unconfigured (null in config and plan)
  Optional types.String `tfsdk:"optional"`
}

func (r ThingResource) Create(ctx context.Context, req resource.CreateResource, resp *resource.CreateResponse) {
  // Providers can opt to use a single variable that is updated based on an
  // external response, however that logic can be more difficult sometimes,
  // so it can be easier to split them. Showing the split way to exemplify
  // the "unset" problem.

  tflog.Trace(ctx, "Plan Values", map[string]any{
    // Null value: types.String{Null: true, Unknown: false, Value: ""}
    "optional": plan.Optional,
  })

  // Maybe some external API responses here, but intentionally not
  // doing any state.Optional setting, which might happen if the
  // external response for that data was null for example.

  tflog.Trace(ctx, "State Values", map[string]any{
    // Zero-value: types.String{Null: false, Unknown: false, Value: ""}
    "optional": state.Optional,
  })

  // The state zero-value will later cause Terraform to error, such as:
  // Error: Provider produced inconsistent result after apply
  // ... expected cty.NullVal(cty.String), got cty.StringVal("")
  // Since the plan value said it would be null.
  resp.Diagnostics.Append(resp.State.Set(ctx, &state)...)
}
```

This deprecation of the fields in preference of functions and methods aims to unexport the internal details and treat the value types as immutable once they are created. When provider developers switch over to the new model, any errant changes to the deprecated exported fields will have no effect. A future version will remove the exported fields entirely and switch the zero-value implementation of these values to being null, instead of a known zero-value of the underlying type.

While not recommended for production usage without extensive testing, provider developers can opt for `panic()` inducing versions of collection value creations, rather than dealing with `diag.Diagnostics` everywhere.

Accessing value information after the migration can be accomplished with the following:

| Prior Value Access | New Value Access |
|----------------------|--------------------|
| `(types.Bool).Value` | `(types.Bool).ValueBool()` |
| `(types.Bool).Null` | `(types.Bool).IsNull()` |
| `(types.Bool).Unknown` | `(types.Bool).IsUnknown()` |
| `(types.Float64).Value` | `(types.Float64).ValueFloat64()` |
| `(types.Float64).Null` | `(types.Float64).IsNull()` |
| `(types.Float64).Unknown` | `(types.Float64).IsUnknown()` |
| `(types.Int64).Value` | `(types.Int64).ValueInt64()` |
| `(types.Int64).Null` | `(types.Int64).IsNull()` |
| `(types.Int64).Unknown` | `(types.Int64).IsUnknown()` |
| `(types.List).Elems` | `(types.List).Elements()` or `(types.List).ElementsAs()` |
| `(types.List).ElemType` | `(types.List).ElementType()` |
| `(types.List).Null` | `(types.List).IsNull()` |
| `(types.List).Unknown` | `(types.List).IsUnknown()` |
| `(types.Map).Elems` | `(types.Map).Elements()` or `(types.Map).ElementsAs()` |
| `(types.Map).ElemType` | `(types.Map).ElementType()` |
| `(types.Map).Null` | `(types.Map).IsNull()` |
| `(types.Map).Unknown` | `(types.Map).IsUnknown()` |
| `(types.Number).Value` | `(types.Number).ValueBigFloat()` |
| `(types.Number).Null` | `(types.Number).IsNull()` |
| `(types.Number).Unknown` | `(types.Number).IsUnknown()` |
| `(types.Object).Attrs` | `(types.Object).Attributes()` or `(types.Object).As()` |
| `(types.Object).AttrTypes` | `(types.Object).AttributeTypes()` |
| `(types.Object).Null` | `(types.Object).IsNull()` |
| `(types.Object).Unknown` | `(types.Object).IsUnknown()` |
| `(types.Set).Elems` | `(types.Set).Elements()` or `(types.Set).ElementsAs()` |
| `(types.Set).ElemType` | `(types.Set).ElementType()` |
| `(types.Set).Null` | `(types.Set).IsNull()` |
| `(types.Set).Unknown` | `(types.Set).IsUnknown()` |
| `(types.String).Value` | `(types.String).ValueString()` |
| `(types.String).Null` | `(types.String).IsNull()` |
| `(types.String).Unknown` | `(types.String).IsUnknown()` |

Go does not allow methods with the same name as a struct field, so a `ValueXXX()` method where XXX represents the returned type was chosen. After the `Value` struct fields are removed, there can be consideration for dropping the XXX in the method naming.

Creating values after the migration can be accomplished with the following:

| Prior Value Creation | New Value Creation |
|----------------------|--------------------|
| `types.Bool{Value: /* value */}` | `types.BoolValue(/* value */)` |
| `types.Bool{Null: true}` | `types.BoolNull()` |
| `types.Bool{Unknown: true}` | `types.BoolUnknown()` |
| `types.Float64{Value: /* value */}` | `types.Float64Value(/* value */)` |
| `types.Float64{Null: true}` | `types.Float64Null()` |
| `types.Float64{Unknown: true}` | `types.Float64Unknown()` |
| `types.Int64{Value: /* value */}` | `types.Int64Value(/* value */)` |
| `types.Int64{Null: true}` | `types.Int64Null()` |
| `types.Int64{Unknown: true}` | `types.Int64Unknown()` |
| `types.List{ElemType: /* element type */, Elems: /* value */}` | `diags := types.ListValue(/* element type */, /* value */)` or `list := types.ListValueMust(/* element type */, /* value */)` |
| `types.List{ElemType: /* element type */, Null: true}` | `types.ListNull(/* element type */)` |
| `types.List{ElemType: /* element type */, Unknown: true}` | `types.ListUnknown(/* element type */)` |
| `types.Map{ElemType: /* element type */, Elems: /* value */}` | `m, diags := types.MapValue(/* element type */, /* value */)` or `m := types.MapValueMust(/* element type */, /* value */)` |
| `types.Map{ElemType: /* element type */, Null: true}` | `types.MapNull(/* element type */)` |
| `types.Map{ElemType: /* element type */, Unknown: true}` | `types.MapUnknown(/* element type */)` |
| `types.Number{Value: /* value */}` | `types.NumberValue(/* value */)` |
| `types.Number{Null: true}` | `types.NumberNull()` |
| `types.Number{Unknown: true}` | `types.NumberUnknown()` |
| `types.Object{AttrTypes: /* attribute types */, Attrs: /* attribute values */}` | `object, diags := types.ObjectValue(/* attribute types */, /* attribute values */)` or `object := types.ObjectValueMust(/* attribute types */, /* attribute values */)` |
| `types.Object{AttrTypes: /* attribute types */, Null: true}` | `types.ObjectNull(/* attribute types */)` |
| `types.Object{AttrTypes: /* attribute types */, Unknown: true}` | `types.ObjectUnknown(/* attribute types */)` |
| `types.Set{ElemType: /* element type */, Elems: /* value */}` | `set, diags := types.SetValue(/* element type */, /* value */)` or `set := types.SetValueMust(/* element type */, /* value */)` |
| `types.Set{ElemType: /* element type */, Null: true}` | `types.SetNull(/* element type */)` |
| `types.Set{ElemType: /* element type */, Unknown: true}` | `types.SetUnknown(/* element type */)` |
| `types.String{Value: /* value */}` | `types.StringValue(/* value */)` |
| `types.String{Null: true}` | `types.StringNull()` |
| `types.String{Unknown: true}` | `types.StringUnknown()` |

### DIFF
--- a/types/bool.go
+++ b/types/bool.go
@@ -103,7 +103,11 @@ type Bool struct {
 	// use the ValueBool method to retrieve the Bool value instead.
 	Value bool
 
-	// state represents whether the Bool is null, unknown, or known.
+	// state represents whether the Bool is null, unknown, or known. During the
+	// exported field deprecation period, this state can also be "deprecated",
+	// which remains the zero-value for compatibility to ensure exported field
+	// updates take effect. The zero-value will be changed to null in a future
+	// version.
 	state valueState
 
 	// value contains the known value, if not null or unknown.
--- a/types/bool.go
+++ b/types/bool.go
@@ -12,15 +12,51 @@ var (
 	_ attr.Value = Bool{}
 )
 
+// BoolNull creates a Bool with a null value. Determine whether the value is
+// null via the Bool type IsNull method.
+//
+// Setting the deprecated Bool type Null, Unknown, or Value fields after
+// creating a Bool with this function has no effect.
+func BoolNull() Bool {
+	return Bool{
+		state: valueStateNull,
+	}
+}
+
+// BoolUnknown creates a Bool with an unknown value. Determine whether the
+// value is unknown via the Bool type IsUnknown method.
+//
+// Setting the deprecated Bool type Null, Unknown, or Value fields after
+// creating a Bool with this function has no effect.
+func BoolUnknown() Bool {
+	return Bool{
+		state: valueStateUnknown,
+	}
+}
+
+// BoolValue creates a Bool with a known value. Access the value via the Bool
+// type ValueBool method.
+//
+// Setting the deprecated Bool type Null, Unknown, or Value fields after
+// creating a Bool with this function has no effect.
+func BoolValue(value bool) Bool {
+	return Bool{
+		state: valueStateKnown,
+		value: value,
+	}
+}
+
 func boolValueFromTerraform(ctx context.Context, in tftypes.Value) (attr.Value, error) {
 	if in.IsNull() {
 		return Bool{
-			Null: true,
+			Null:  true,
+			state: valueStateDeprecated,
 		}, nil
 	}
 	if !in.IsKnown() {
 		return Bool{
 			Unknown: true,
+			state:   valueStateDeprecated,
 		}, nil
 	}
 	var b bool
@@ -28,21 +64,50 @@ func boolValueFromTerraform(ctx context.Context, in tftypes.Value) (attr.Value, 
 	if err != nil {
 		return nil, err
 	}
-	return Bool{Value: b}, nil
+	return Bool{
+		Value: b,
+		state: valueStateDeprecated,
+	}, nil
 }
 
 // Bool represents a boolean value.
 type Bool struct {
 	// Unknown will be true if the value is not yet known.
+	//
+	// If the Bool was created with the BoolValue, BoolNull, or BoolUnknown
+	// functions, changing this field has no effect.
+	//
+	// Deprecated: Use the BoolUnknown function to create an unknown Bool
+	// value or use the IsUnknown method to determine whether the Bool value
+	// is unknown instead.
 	Unknown bool
 
 	// Null will be true if the value was not set, or was explicitly set to
 	// null.
+	//
+	// If the Bool was created with the BoolValue, BoolNull, or BoolUnknown
+	// functions, changing this field has no effect.
+	//
+	// Deprecated: Use the BoolNull function to create a null Bool value or
+	// use the IsNull method to determine whether the Bool value is null
+	// instead.
 	Null bool
 
 	// Value contains the set value, as long as Unknown and Null are both
 	// false.
+	//
+	// If the Bool was created with the BoolValue, BoolNull, or BoolUnknown
+	// functions, changing this field has no effect.
+	//
+	// Deprecated: Use the BoolValue function to create a known Bool value or
+	// use the ValueBool method to retrieve the Bool value instead.
 	Value bool
+
+	// state represents whether the Bool is null, unknown, or known.
+	state valueState
+
+	// value contains the known value, if not null or unknown.
+	value bool
 }
 
 // Type returns a BoolType.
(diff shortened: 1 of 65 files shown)


types/bool.go
```diff
@@ -12,37 +12,113 @@ var (
 	_ attr.Value = Bool{}
 )
 
+// BoolValue creates a Bool with a known value. Access the value via the Bool
+// type BoolValue method.
+//
+// Setting the deprecated Bool type Null, Unknown, or Value fields after
+// creating a Bool with this function has no effect.
+func BoolValue(value bool) Bool {
```
bendbennett, line 20: Wonder whether we might want to consider using `ValueBool` for consistency with `NullBool` and `UnknownBool`? This would also differentiate the function for instantiating a `Bool` with a value from the function that retrieves the value (`b.BoolValue`).

bflad: I'm okay with that. 👍

bflad: Pushed up that change and rebased. 😄

bflad:
@bendbennett this should be ready to review now. 👍

bendbennett (approved):
LGTM

For the refactoring of functions such as `ValueFromTerraform` I'm guessing that dropping usage of the deprecated fields will happen at the time these fields are removed?

types/object.go
```diff
@@ -152,6 +152,126 @@ func (t ObjectType) ValueType(_ context.Context) attr.Value {
...
+//
+// Setting the deprecated Object type AttrTypes, Attrs, Null, or Unknown fields
+// after creating a Object with this function has no effect.
+func ObjectValue(attributeTypes map[string]attr.Type, attributes map[string]attr.Value) (Object, diag.Diagnostics) {
+	var diags diag.Diagnostics
+
+	// Ideally, Type() would not require a context.Context as it has no benefit
+	// here or elsewhere. There is also no benefit to adding it to the function
+	// parameters at the moment.
+	ctx := context.Background()
```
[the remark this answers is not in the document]
bflad: I'll raise an issue, sorry for the code comment banter here. We can decide whether its a good idea or not (I personally think so because the type information should be very static and not need anything like logging, but maybe I'm off base here).

Issue #498, opened 2022-09-23
Consider Always Returning Errors with Get/GetAttribute for Target Types Without Unknown Support


bflad:
### Module version

```
v0.13.0
```

### Use-cases

When accessing schema-based data (such as configuration, plan, and state), there are a plethora of various rules associated with the possible values being null or unknown. Some of this is based on the following:

- Which RPC is involved
- Whether it is the RPC request or response data
- How the attribute is defined (Computed, Optional, or Required)

Some of these rules are documented in the current [website documentation](https://www.terraform.io/plugin/framework/accessing-values#when-can-a-value-be-unknown-or-null):

- Required attributes will never be null or unknown in Create, Read, Update, or Delete methods.
- Optional attributes that are not computed will never be unknown in Create, Read, Update, or Delete methods.
- Computed attributes, whether optional or not, will never be null in the plan for Create, Read, Update, or Delete methods.
- Computed attributes that are read-only (Optional is not true) will always be unknown in the plan for Create, Read, Update, or Delete methods. They will always be null in the configuration for Create, Read, Update, and Delete methods.
Required attributes will never be null in a provider's Configure method. They may be unknown.
- The state never contains unknown values.
- The configuration for Create, Read, Update, and Delete methods never contains unknown values.

These all tend to confuse provider developers as evidenced by the various amount of inbound GitHub issues, HashiCorp Discuss posts, and out-of-public feedback the framework maintainers receive. Provider developers are free to implement Go standard library types or provider-defined types to accept this schema-based data, however the current framework design takes a runtime error approach when it comes to feedback the provider developer. This feedback loop likely involves practitioners, since the provider developer may not be able to tease out this potential issue during development or testing.

### Attempted Solutions

The next version of the framework will improve the runtime error messaging for null and unknown values, however there is potential for unknown values based on the practitioner configuration. This is outside the provider developers control and leaving a potential practitioner-based feedback to provider developers is a poor experience. The potential for a value to be unknown can also tends to be hard to acceptance test, as acceptance testing tends to use either hardcoded values or a preference to not include unnecessary data sources or resources.

### Proposal

[...]

Issue #418, opened 2022-07-18
Refactor Schema and Schema Data Related Documentation


bflad:
### Module version

```
v0.10.0
```

### Use-cases

The website documentation is less than ideal as it has quite a few schema-related pages as top-level in the lengthy side navigation. In particular:

* https://terraform.io/plugin/framework/accessing-values
* https://terraform.io/plugin/framework/paths
* https://terraform.io/plugin/framework/path-expressions
* https://terraform.io/plugin/framework/schemas
* https://terraform.io/plugin/framework/writing-state

It would be clearer if these were bundled together as they are conceptually related.

### Proposal

Create a new top-level section named "Schemas and Schema Data" or "Schemas and Config/Plan/State" with the following:

* Overview page, brief introduction to the concepts and linking to others
* Redirected accessing values page
* Redirected paths page
* Redirected path expressions page
* Redirected schemas page (maybe "Schema Definitions"?)
* Redirected writing state page (probably should not say state as it also related to setting the plan in plan modifiers)

Doing this will likely warrant fixing wording/flow for all the concepts/pages to ensure they seem cohesive together.

### References

- https://github.com/hashicorp/terraform-plugin-framework/pull/413

bflad:
This updated documentation should also contain a reference page for schema block support.

bflad:
The updated documentation should also contain _complex_ data modeling and provider logic (e.g. nested lists, nested attributes)

PR #504, opened 2022-09-30
website: Further discourage usage of non-types package types with accessing schema data

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/498
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/418
Reference: https://discuss.hashicorp.com/t/terraform-plan-does-not-allow-sending-only-delta-for-list-type-parameters-in-the-resource-block/45027
Reference: https://discuss.hashicorp.com/t/set-state-for-optional-nested-attributes-for-singlenestedattributes/44838
Reference: https://discuss.hashicorp.com/t/why-is-req-plan-get-in-my-create-method-throwing-an-error/44441
Reference: https://discuss.hashicorp.com/t/terraform-plugin-framework-value-conversion-error-unhandled-unknown-value/44382
... plenty more ...

The current recommendation is now to always use `types` package types, unless there is no chance for null/unknown values. The conversion rules sections will likely be further handled in #418 by having a schema conversion page that outlines how to map each framework-defined type into value types.

### DIFF
--- a/website/docs/plugin/framework/accessing-values.mdx
+++ b/website/docs/plugin/framework/accessing-values.mdx
@@ -17,7 +17,7 @@ its source.
 The data is usually stored in a request object:
 
 ```go
-func (m myResource) Create(ctx context.Context,
+func (r ThingResource) Create(ctx context.Context,
 	req resource.CreateRequest, resp *resource.CreateResponse)
 ```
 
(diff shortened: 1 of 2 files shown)


bendbennett (approved):
LGTM

PR #505, opened 2022-10-01
website: Fix old code examples to tfsdk.ModifyResourcePlanRe(quest|sponse)

bflad:
Reference: https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/resource#ModifyPlanRequest
Reference: https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/resource#ModifyPlanResponse

This was missed when the `resource` package was initially created.


bendbennett (approved):
LGTM

PR #522, opened 2022-10-24
types: Introduce ListValueFrom, MapValueFrom, ObjectValueFrom, and SetValueFrom functions

### LINKED ISSUE #520: types: Consider Adding Collection Type ValueFrom Functions
### Module version

```
v0.15.0 (not yet released)
```

### Use-cases

Provider developers can find it difficult to convert populated API SDK types or standard Go types (e.g. `map[string]string]`) to the framework collection types (e.g. `types.Map`). While there are various ways to get this working, this requires some additional discovery beyond using just the `types` package, which is not a great experience.

### Attempted Solutions

Using [`tfsdk.ValueFrom()`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework@v0.14.0/tfsdk#ValueFrom), which I do not believe is documented in the website documentation:

```go
var fwMap types.Map
diags := tfsdk.ValueFrom(
  ctx,
  map[string]string{"test-key": "test-value"},
  types.MapType{ElemType: types.StringType},
  &fwMap,
)
```

### Proposal

As part of #502, there will be new `types` value type creation functions such as:

- `func ListValue(elementType attr.Type, elements []attr.Value) types.List`
- `func MapValue(elementType attr.Type, elements map[string]attr.Value) types.Map`
- `func SetValue(elementType attr.Type, elements []attr.Value) types.Set`

The framework could also introduce these related creation functions, which use the same underlying logic as `tfsdk.ValueFrom()` to perform reflection from non-framework types.

- `func ListValueFrom(ctx context.Context, elementType attr.Type, elements []any) (types.List, diag.Diagnostics)`
- `func MapValueFrom(ctx context.Context, elementType attr.Type, elements map[string]any) (types.Map, diag.Diagnostics)`
- `func SetValueFrom(ctx context.Context, elementType attr.Type, elements []any) (types.Set, diag.Diagnostics)`

Compared to `tfsdk.ValueFrom()`, they would automatically handle the collection type (only requiring the element type) and offer a slight amount of additional compiler checking over just `any`.

### References

- https://github.com/hashicorp/terraform-plugin-framework/issues/323
- https://github.com/hashicorp/terraform-plugin-framework/pull/350
- https://discuss.hashicorp.com/t/setting-state-for-types-map/45882

bflad:
Closes #520

These will enable provider developers to use the framework type system's built-in conversion rules to create collection types, rather than using more generic `tfsdk.ValueFrom()` or other methodologies.

In this example, a map using standard Go types is used to create a `types.Map` framework type with known values:

```go
apiMap := map[string]string{
  "key1": "value1",
  "key2": "value2",
}
fwMap, diags := types.MapValueFrom(ctx, types.StringType, apiMap)
```

There may be additional use cases or needs that get teased out with this introduction, such as the ability to create a `types.Object` from a `map[string]any`, however those can be handled in potential future feature requests.


bendbennett (approved):
LGTM 👍

PR #525, opened 2022-10-31
fix broken links

laurapacilio:
Broken links are a bad user experience - this PR fixes some that we found in the latest review of devdot! :)


bflad (approved):
Looks good to me 🚀

Issue #535, opened 2022-11-11
attr: Add `ValueFromFramework` and `ToFrameworkValue` Interface Functions to attr.Value


bendbennett:
### Module version
```
github.com/hashicorp/terraform-plugin-framework v0.15.0
```

### Use-cases
A first step required for making the use of custom types for nested attributes feasible requires a mechanism for converting to/from the framework type (e.g., `types.List`). 

### Attempted Solutions
Using an interface implemented by custom types that is specific to the framework type (e.g., `types.List`) that they embed and then using type assertion to determine if the type being used is either a framework type (e.g., `types.List`) or a type that implements an interface that gives access to the framework type (e.g., `types.List`) embedded in the custom type. 

### Proposal
Introduce separate interfaces for each of the framework types that can be implemented by custom types to allow access to the embedded type and to obtain an instance of the custom type by passing the framework type.

```go
package attr

type ListType interface {
  attr.Type // common bits

  ValueFromFramework(context.Context, types.List) (Value, diag.Diagnostics)
}

type ListValue interface {
  attr.Value

  ToFrameworkValue(context.Context) (types.List, diag.Diagnostics)
}
```

Once these interfaces are in place then custom types that embed framework types (e.g., `types.List`) can be created and used in place of framework types.

### References
- #532

PR #536, opened 2022-11-11
types: Add Typable and Valuable Interfaces

bendbennett:
Closes: #535

### DIFF
--- a/internal/fwserver/attribute_validation.go
+++ b/internal/fwserver/attribute_validation.go
@@ -140,7 +140,7 @@ func AttributeValidateNestedAttributes(ctx context.Context, a fwschema.Attribute
 	nm := a.GetAttributes().GetNestingMode()
 	switch nm {
 	case fwschema.NestingModeList:
-		listVal, ok := req.AttributeConfig.(types.ListVal)
+		listVal, ok := req.AttributeConfig.(types.ListValuable)
 
 		if !ok {
 			err := fmt.Errorf("unknown attribute value type (%T) for nesting mode (%T) at path: %s", req.AttributeConfig, nm, req.AttributePath)
@@ -153,7 +153,7 @@ func AttributeValidateNestedAttributes(ctx context.Context, a fwschema.Attribute
 			return
 		}
 
-		l, diags := listVal.ToFrameworkValue(ctx)
+		l, diags := listVal.ToListValue(ctx)
 
 		resp.Diagnostics.Append(diags...)
 		if resp.Diagnostics.HasError() {
@@ -177,7 +177,7 @@ func AttributeValidateNestedAttributes(ctx context.Context, a fwschema.Attribute
 			}
 		}
 	case fwschema.NestingModeSet:
-		setVal, ok := req.AttributeConfig.(types.SetVal)
+		setVal, ok := req.AttributeConfig.(types.SetValuable)
 
 		if !ok {
 			err := fmt.Errorf("unknown attribute value type (%T) for nesting mode (%T) at path: %s", req.AttributeConfig, nm, req.AttributePath)
@@ -190,7 +190,7 @@ func AttributeValidateNestedAttributes(ctx context.Context, a fwschema.Attribute
 			return
 		}
 
-		s, diags := setVal.ToFrameworkValue(ctx)
+		s, diags := setVal.ToSetValue(ctx)
 
 		resp.Diagnostics.Append(diags...)
 		if resp.Diagnostics.HasError() {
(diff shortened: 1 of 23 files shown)


bflad (commented):
Overall this looks like a really good direction. We'll also want to ensure there's Go documentation for the new interfaces and methods.

types/bool.go
```diff
@@ -5,13 +5,21 @@ import (
...
+
 	"github.com/hashicorp/terraform-plugin-go/tftypes"
 )
 
 var (
-	_ attr.Value = Bool{}
+	_ BoolValuable = Bool{}
 )
 
+type BoolValuable interface {
```
bflad, line 17: Let's ensure all these new `types` interfaces have Go documentation. 👍 

```suggestion
// BoolValuable extends attr.Value for boolean value types.
// Implement this interface to create a custom Bool value type.
type BoolValuable interface {
```

types/bool.go
```diff
@@ -5,13 +5,21 @@ import (
...
+	attr.Value
+
+	ToBoolValue(ctx context.Context) (Bool, diag.Diagnostics)
```
bflad, line 20:
```suggestion
	// ToBoolValue should convert the value type to a Bool.
	ToBoolValue(ctx context.Context) (Bool, diag.Diagnostics)
```

types/list.go
```diff
@@ -15,10 +15,22 @@ import (
...
-	_ attr.Type  = ListType{}
-	_ attr.Value = &List{}
+	_ ListTypable  = ListType{}
+	_ ListValuable = &List{}
 )
 
+type ListTypable interface {
+	attr.Type
+
+	ValueFromList(context.Context, List) (attr.Value, diag.Diagnostics)
```
bflad, line 25: Another thing we should likely consider for these is that the returned attr.Value can be constrained further to only allowing the value interface specific to the type.

```
// ValueFromList should return the equivalent value from a given List.
ValueFromList(context.Context, List) (ListValuable, diag.Diagnostics)
```

types/list.go
```diff
@@ -15,10 +15,28 @@ import (
...
 
 var (
-	_ attr.Type  = ListType{}
-	_ attr.Value = &List{}
+	_ ListTypable  = ListType{}
+	_ ListValuable = &List{}
 )
 
+// ListTypable extends attr.Type for list type types.
+// Implement this interface to create a custom ListType type type.
```
bflad, line 23: Nit: I think we can just refer to the "type type" as just "type" and leave the distinction of "value type" to be separate 😎 

```suggestion
// ListTypable extends attr.Type for list types.
// Implement this interface to create a custom ListType type.
```

types/list.go
```diff
@@ -15,10 +15,28 @@ import (
...
+type ListTypable interface {
```
bflad, line 24: Can you add changelog entries for these? e.g.

```
types: Added `ListTypable` and `ListValuable` interface types, which enable embedding existing list types for custom types
```

Or something along those lines?

internal/fwserver/attribute_validation.go
```diff
@@ -140,19 +140,26 @@ func AttributeValidateNestedAttributes(ctx context.Context, a fwschema.Attribute
...
+		listVal, ok := req.AttributeConfig.(types.ListValuable)
 
 		if !ok {
 			err := fmt.Errorf("unknown attribute value type (%T) for nesting mode (%T) at path: %s", req.AttributeConfig, nm, req.AttributePath)
 			resp.Diagnostics.AddAttributeError(
 				req.AttributePath,
-				"Attribute Validation Error",
-				"Attribute validation cannot walk schema. Report this to the provider developer:\n\n"+err.Error(),
+				"Attribute Validation Error Invalid Value Type",
+				"A type from which a types.List can be obtained is expected here. Report this to the provider developer:\n\n"+err.Error(),
```
bflad, line 150: Nit: Should we give the direct hint about the `types.ListValuable` interface for these?

internal/fwserver/attribute_validation.go
```diff
@@ -140,19 +140,26 @@ func AttributeValidateNestedAttributes(ctx context.Context, a fwschema.Attribute
...
 			)
 
 			return
 		}
 
+		l, diags := listVal.ToListValue(ctx)
```
bflad, line 156: I think we're ready for unit testing these! Maybe some additional types in `internal/testing/types`?

bflad (commented):
I think we're getting close here!

bflad (approved):
Looks good to me 🚀 Nice work.

Issue #258, opened 2022-02-03
Consider Documentation and Tutorials for Common Use Cases


bflad:
### Use-cases

Provider developers across the ecosystem will have some common use cases they will want to research documentation on how to accomplish their goals. For example, how to implement provider configuration validation for attributes which should also source standard vendor SDK environment variables.

Currently provider developers are left to their own devices beyond piecing together the Go documentation and terraform.io documentation that typically describes in detail a single piece of the framework rather than implementations that require crossing those boundaries. This means that it can potentially take much longer to be successful as the next options may be researching other Terraform Provider implementations, asking on [HashiCorp Discuss](https://discuss.hashicorp.com), or reaching out on other community forums.

### Proposal

For a timeboxed period, use this issue to collect some "common" provider development use cases with a brief summary and potential solution, linking to other documentation or providing code examples as necessary. When the collection period is over, holistically evaluate which (if any or all) should potentially be surfaced in the terraform.io documentation website or learn.hashicorp.com tutorials and how best to accomplish that since they are very likely to be disparate in nature.

bflad:
## Provider Attribute Validation With Environment Variables

Providers may need to support attributes that are configurable via Terraform configuration as well as standard vendor environment variables (e.g. `AWS_DEFAULT_REGION`, `AZURE_DEFAULTS_LOCATION`, `GITHUB_TOKEN`, `CLOUDSDK_CORE_PROJECT`, etc.). To accomplish this, the attribute should be optional, and the validation implemention should occur in the provider `Configure` method. This ensures that `terraform validate` can remain offline (or put another way, statically validated anywhere), while also giving the provider the opportunity to set the expected precedence and potential unknown value handling.

Refer to [terraform-provider-hashicups-pf](https://github.com/hashicorp/terraform-provider-hashicups-pf/blob/main/hashicups/provider.go#L64-L88) for an example of how this can be accomplished today.

bflad:
## Attribute Default Values

Resources and data sources often want to specify a default value for an attribute, so practitioners do not need to declare everything explicitly and to perform drift detection against the corresponding default value from the provider. While the framework documentation touches on this handling in the [plan modifiers](https://www.terraform.io/plugin/framework/resources/plan-modification#creating-attribute-plan-modifiers) page, this probably warrants its own page to explain the necessary `Computed: true` schema as well.

We can separately consider whether the framework should be providing some standard plan modifiers for these cases.

tmccombs:
[...]

PR #537, opened 2022-11-11
website: Split resource lifecycle management into separate pages

### LINKED ISSUE #527: Document Resource Update HasChange Equivalent
### Module version

```
v0.15.0
```

### Use-cases

Provider developers migrating from terraform-plugin-sdk may be looking for something similar to its [`(helper/schema.ResourceData).HasChanges(...string)` method](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema#ResourceData.HasChanges), which checking for individual attribute changes between plan and prior state. Some APIs require only sending differences during in-place updates of resources.

### Attempted Solutions

Figuring out how to do it without documentation.

### Proposal

Document current solutions in resource documentation and migration guide.

If using a schema data model type:

```go
type ThingResourceModel struct {
  Attribute1 types.String `tfsdk:"attribute1"`
  Attribute2 types.String `tfsdk:"attribute2"`
  // ...
}

// Update logic

resp.Diagnostics.Append(req.State.Get(ctx, &state)...)

  return
}

if plan.Attribute1.Equal(state.Attribute1) {
  // ...
}

if plan.Attribute2.Equal(state.Attribute2) {
  // ...
}
```

If getting individual attributes:

```go
// Update logic
// Reusable variables for brevity
var attributePlan, attributeState types.String

resp.Diagnostics.Append(req.Plan.GetAttribute(ctx, path.Root("attribute1"), &attributePlan)...)
resp.Diagnostics.Append(req.State.GetAttribute(ctx, path.Root("attribute1"), &attributeState)...)

  return
}

if attributePlan.Equal(attributeState) {
  // ...
}

resp.Diagnostics.Append(req.Plan.GetAttribute(ctx, path.Root("attribute2"), &attributePlan)...)
resp.Diagnostics.Append(req.State.GetAttribute(ctx, path.Root("attribute2"), &attributeState)...)

  return
}

if attributePlan.Equal(attributeState) {
  // ...
}
```

### References

- https://github.com/hashicorp/terraform-plugin-framework/issues/526

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/258
Closes #527

Each lifecycle operation has its own set of caveats, recommendations, and specific use cases. These individualized pages will clarify and enable the additional documentation requirements. This change also adds update documentation to show how to determine whether an individual attribute value was changed during update, to show the new layout in action.

### DIFF
--- a/website/docs/plugin/framework/resources/delete.mdx
+++ b/website/docs/plugin/framework/resources/delete.mdx
@@ -150,3 +150,4 @@ Note these caveats when implementing the `Delete` method:
 Note these recommendations when implementing the `Delete` method:
 
 * Ignore errors that signify the resource is no longer existent.
+* Skip calling the response state `RemoveResource()` method. The framework automatically handles this logic with the response state if there are no error diagnostics.
--- a/website/docs/plugin/framework/resources/delete.mdx
+++ b/website/docs/plugin/framework/resources/delete.mdx
@@ -1,0 +1,152 @@
+---
+page_title: 'Plugin Development - Framework: Delete Resources'
+description: >-
+  How to implement resource deletion in the provider development framework.
+---
+
+# Delete Resources
+
+-> **Note:** The Plugin Framework is in beta.
+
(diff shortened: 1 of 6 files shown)


bendbennett (approved):
Nice!

website/docs/plugin/framework/resources/delete.mdx
```diff
@@ -0,0 +1,152 @@
...
+Note these caveats when implementing the `Delete` method:
+
+* An error is returned if the response state is set to anything other than null.
+* Any response errors will cause Terraform to keep the resource under management.
+
+## Recommendations
+
+Note these recommendations when implementing the `Delete` method:
+
+* Ignore errors that signify the resource is no longer existent.
```
bflad, line 152:
```suggestion
* Ignore errors that signify the resource is no longer existent.
* Skip calling the response state `RemoveResource()` method. The framework automatically handles this logic with the response state if there are no error diagnostics.
```

Issue #540, opened 2022-11-15
Update migration guide to indicate Set field and SchemaSetFunc equivalents are not required when using the Framework 


bendbennett:
### Module version
```
terraform-plugin-framework v0.15.0
```

### Use-cases
When migrating providers from `terraform-plugin-sdk/v2` to `terraform-plugin-framework` it is not clear whether the `Set` field which holds a [SchemaSetFunc](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2@v2.24.1/helper/schema#SchemaSetFunc) needs to be migrated. For instance, the [archive provider](https://github.com/hashicorp/terraform-provider-archive/blob/v2.2.0/internal/provider/data_source_archive_file.go#L48) makes use of the `Set` field on `schema.Schema` to define a custom `SchemaSetFunc`. 

### Attempted Solutions
Look at the documentation for `terraform-plugin-sdk/v2` to determine whether migration of the `Set` field and associated `SchemaSetFunc` is required when implementing a provider in `terraform-plugin-framework`.

### Proposal
Update the [migration guide](https://developer.hashicorp.com/terraform/plugin/framework/migrating) to make it explicit that there is no need to migrate `Set` and associated `SchemaSetFunc` when migrating a provider to the Framework.

PR #541, opened 2022-11-15
Update migration guide to indicate Set field and SchemaSetFunc equivalents are not required when using the Framework

bendbennett:
Closes: #540


website/docs/plugin/framework/migrating/schema/index.mdx
```diff
@@ -96,6 +96,8 @@ type Schema struct {
 }
 ```
 
+
+
```
bflad, line 100:
```suggestion
```

bflad (approved):
Otherwise, looks good to me 🚀

PR #542, opened 2022-11-16
schema/validator: New type-specifc validators package

### LINKED ISSUE #83: Create a Describer interface
We keep using `Description` and `MarkdownDescription` everywhere. We should just capture them as a `Describer` interface we can include in other interfaces, so if we add or change things, it's consistent.

bflad:
Closes #83 (we should explicitly create individual `Describer` interfaces per concept, if desired)

As part of upcoming effort to split schema functionality into the `datasource`, `provider`, and `resource` packages, there are some improvements that will land in the new implementations rather than breaking the existing `tfsdk` package schema functionality.

One area which has caused developer burden is that "attribute" validators, currently implementations of the `tfsdk.AttributeValidator` interface,  receive a generic `attr.Value` as the configuration value to perform validation logic. This means that implementors must currently handle validating and converting the value into the concrete type they expect.

The upcoming split schemas handling will introduce separate attribute/block types that will enable to framework to strongly type validators and other future schema enhancements. This change prepares the exported interfaces and internal validation logic for those enhancements.

### DIFF
--- a/.changelog/pending.txt
+++ b/.changelog/pending.txt
@@ -1,0 +1,3 @@
+```release-note:feature
+schema/validator: New package which contains type-specific schema validator interfaces
+```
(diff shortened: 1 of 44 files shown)


bendbennett (approved):
Awesome. LGTM

PR #544, opened 2022-11-17
tfsdk: Switch Config, Plan, and State type Schema fields to interface

bflad:

As part of upcoming effort to split schema functionality into the datasource, provider, and resource packages, there are some internal implementation details which need to be sorted beforehand. Throughout the 0.x versions, schema functionality has been migrated from the `tfsdk` package into the `internal/fwschema` package, however the exported `Config`, `Plan`, and `State` type `Schema` fields remained hardcoded to the `tfsdk.Schema` type. This change migrates the field type to the `fwschema.Schema` interface, so other framework schema implementations can be introduced.


bendbennett (approved):
LGTM 👍

Issue #532, opened 2022-11-10
Consider Allowing Usage of Custom Types for Nested Attributes


bendbennett:
### Module version
```
github.com/hashicorp/terraform-plugin-framework v0.15.0
```

### Use-cases
In order to provider greater flexibility when using nested attributes, the ability to use a custom type for a nested attribute would be beneficial. This would provide the same advantages afforded when using custom types for simple types, for example using an `RFC3339` type rather than a string type for the handling of strings that hold RFC3339 timestamps. See [usage of a prototype for an RFC3339 type](https://github.com/hashicorp/terraform-provider-time/pull/126/commits/61bf0d1ba4af0d20c21f54bd469c36d8dbacd950) for an example.

### Attempted Solutions
Helper functions can be created which will handle nested attributes in a specific way, but this is not as streamlined as having custom types for nested attributes where functions can be directly associated with the type. 

### Proposal
Refactoring to expose the `NestedAttributes` interface and having all `attr.Value` and any custom nested attribute implement a new `FrameworkValue` interface allows custom types to be used for nested attributes.

For example, the following terraform configuration, provider resource schema, model (for unmarshalling onto) and create function should allow the usage of a function associated with the custom type. Both a "standard" list nested attribute and a list nested attribute using a custom type are illustrated for comparison

#### Terraform Configuration

```terraform
resource "example_resource" "example" {
  list_nested_attribute = [
    {
      int64_attribute = 9223372036854775807
      list_attribute  = ["list-element", "list-element"]
    },
    {
      int64_attribute = 9223372036854775807
      list_attribute  = ["list-element", "list-element"]
    }
  ]

  list_nested_attribute_custom = [
    {
      int64_attribute = 9223372036854775807
      list_attribute  = ["list-element", "list-element"]
    },
    {
      int64_attribute = 9223372036854775807
      list_attribute  = ["list-element", "list-element"]
    }
  ]
```

#### Provider Resource Schema

```go
func (e *exampleResource) GetSchema(ctx context.Context) (tfsdk.Schema, diag.Diagnostics) {
	return tfsdk.Schema{
		Attributes: map[string]tfsdk.Attribute{
			"list_nested_attribute": {
				Optional: true,
				Attributes: tfsdk.ListNestedAttributes(map[string]tfsdk.Attribute{
					"int64_attribute": {
						Optional: true,
						Type:     types.Int64Type,
					},
					"list_attribute": {
						Optional: true,
[...]

Issue #508, opened 2022-10-05
Consider Allowing Usage of Custom Types for Structural Elements


bendbennett:
### Module version

```
github.com/hashicorp/terraform-plugin-framework v0.14.0
```

### Use-cases

During the development of the [v0.1.0 of terraform-plugin-framework-timeouts](https://github.com/hashicorp/terraform-plugin-framework-timeouts/releases/tag/v0.1.0) it became clear that it would be useful to be able to use custom types for structural elements such as blocks. The use-case in this instance was to be able to use a custom type within the schema and model for timeouts so that behaviour (e.g., functions) associated with the custom type could be used by provider developers when accessing timeouts.

For instance, given the following configuration:

```terraform
resource "timeouts_example" "example" {
  timeouts {
    create = "60m"
    read = "30m"
    update = "12m"
  }
``` 

It would simplify provider developer experience if schema and model could be defined as follows:

```go
func (t *exampleResource) GetSchema(ctx context.Context) (tfsdk.Schema, diag.Diagnostics) {
	return tfsdk.Schema{
		Attributes: map[string]tfsdk.Attribute{
			/* ... */
		},

		Blocks: map[string]tfsdk.Block{
			"timeouts": timeouts.Block(ctx, timeouts.Opts{
				Create: true,
				Read:   true,
				Update: true,
			}),
		},
	}, nil
}

type exampleResourceData struct {
	/* ... */
	Timeouts    timeouts.Timeouts `tfsdk:"timeouts"`
}
```

Access and usage of the timeouts within CRUD functions might then look something like the following:

```go
func (r exampleResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
	var data exampleResourceData

	diags := req.Plan.Get(ctx, &data)
	resp.Diagnostics.Append(diags...)

	if resp.Diagnostics.HasError() {
		return
	}

	createTimeout := data.Timeouts.Create(ctx, time.Minute*20)
	
	ctx, cancel := context.WithTimeout(ctx, createTimeout)
	defer cancel()

	/* ... */
}
```

Currently, it's not possible to use a custom type (e.g., `timeouts.Timeouts`) for structural elements, in this case `tfsdk.Block` as `types.Object` is used within the Framework and using a custom type results in errors indicating a type mismatch between the custom type and `types.Object`.

### Attempted Solutions

[...]

Issue #489, opened 2022-09-15
Plan modifiers inside nested attributes doesn't seem to do anything in SingleNestedAttribute


pksunkara:
### Module version

```
github.com/hashicorp/terraform-plugin-framework v0.11.1
```

### Relevant provider source code

```go
func (t teamResourceType) GetSchema(ctx context.Context) (tfsdk.Schema, diag.Diagnostics) {
	return tfsdk.Schema{
		MarkdownDescription: "Test team.",
		Attributes: map[string]tfsdk.Attribute{
			"bool_known_default": {
				MarkdownDescription: "Boolean with known default. **Default** `true`.",
				Type:                types.BoolType,
				Optional:            true,
				Computed:            true,
				PlanModifiers: tfsdk.AttributePlanModifiers{
					modifiers.DefaultBool(true),
				},
			},
			"nullable_bool": {
				MarkdownDescription: "Nullable boolean. **Default** `null`.",
				Type:                types.BoolType,
				Optional:            true,
				Computed:            true,
				PlanModifiers: tfsdk.AttributePlanModifiers{
					modifiers.NullableBool(),
				},
			},
			"nested": {
				MarkdownDescription: "Nested block.",
				Optional:            true,
				Computed:            true,
				// PlanModifiers: tfsdk.AttributePlanModifiers{
				//	defaultObjectModifier{},
				// },
				Attributes: tfsdk.SingleNestedAttributes(map[string]tfsdk.Attribute{
					"bool_empty_default": {
						MarkdownDescription: "Bool with empty default. **Default** `false`.",
						Type:                types.BoolType,
						Optional:            true,
						Computed:            true,
						PlanModifiers: tfsdk.AttributePlanModifiers{
							modifiers.DefaultBool(false),
						},
					},
				}),
			},
		},
	}, nil
}

type teamResourceData struct {
	BoolKnownDefault            types.Bool    `tfsdk:"bool_known_default"`
	NullableBool                types.Bool    `tfsdk:"nullable_bool"`
	Nested                      types.Object  `tfsdk:"nested"`
}

func (r teamResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
	var data teamResourceData

	diags := req.Plan.Get(ctx, &data)
	resp.Diagnostics.Append(diags...)

	if resp.Diagnostics.HasError() {
		return
	}

	input := TeamsInsertInput{
[...]

PR #545, opened 2022-11-18
internal/fwschema: Add NestedAttributeObject and NestedBlockObject interfaces

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/508
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/532

Since the initial version of the framework, it has treated nested attributes and blocks as a singular entity with some internal gymnastics to handle the one (single; object) or two (list, map, set to object) actual types that make up the nested attribute or block. Having these schema abstractions as single entities caused developer confusion when attempting to extract data or create paths.

Other desirable features for the framework are the ability to customize the types, define validators, and define plan modifiers of nested attributes and blocks. This type customization, validation, or plan modification may need to be on the nested object to simplify provider implementations. Exposing these options with a flat abstraction could be confusing.

This change introduces the concept of a nested attribute object and a nested block object into the `internal/fwschema` package and makes the necessary implementation changes to `tfsdk.Attribute` and `tfsdk.Block`. While the `tfsdk` package schema handling does not expose the new potential enhancements, the upcoming `datasource`, `provider`, and `resource` schema handling will. The existing unit testing shows no regressions and the new unit tests with validation show it should be possible to implement nested object validation (and when implemented in the near future, plan modification).

This changeset additionally migrates more `tfsdk.Schema` logic into the `internal/fwschema` package in further preparation for creating additional schema implementations that will benefit from the existing schema logic.


bendbennett (approved):

pksunkara:
This is some great work that's much needed (which I already mentioned [here](https://github.com/hashicorp/terraform-plugin-framework/issues/489#issuecomment-1256947639)). But I have no idea on how to start using this, the docs are not good enough. 

Can you guys please add some examples in the repository with each use case? That would be really helpful.

Issue #136, opened 2021-09-08
Framework allows top-level schema attributes that conflict with Terraform meta-arguments


ewbankkit:
Terraform Core allows a number of per-resource/data source _meta-arguments_ that override or enhance Terraform behaviors.
The framework currently allows top-level schema attributes that conflict with these, causing who-knows-what (potentially silent) failures or errors.
The framework should validate that a resource or data source schema does not define any of these top-level attributes:
* `count`
* `depends_on`
* `for_each`
* `lifecycle`
* `provider`

PR #548, opened 2022-11-21
Fix Framework allows top-level schema attributes that conflict with Terraform meta-arguments

bendbennett:
Closes: #136

Once all references to `tfsdk.Schema` have been removed we can ammend `fwschema/schema.go`:

```go
type Schema interface {
    /* ... */

    Validate() diag.Diagnostics
}
```

### DIFF
--- a/datasource/schema/schema.go
+++ b/datasource/schema/schema.go
@@ -2,12 +2,15 @@ package schema
 
 import (
 	"context"
+	"fmt"
+	"regexp"
+
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
 
 	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/internal/fwschema"
 	"github.com/hashicorp/terraform-plugin-framework/path"
-	"github.com/hashicorp/terraform-plugin-go/tftypes"
 )
 
 // Schema must satify the fwschema.Schema interface.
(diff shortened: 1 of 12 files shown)


datasource/schema/schema.go
```diff
@@ -122,6 +125,130 @@ func (s Schema) TypeAtTerraformPath(ctx context.Context, p *tftypes.AttributePat
...
+		nestedObject := na.GetNestedObject()
+
+		if nestedObject == nil {
+			return diags
+		}
+
+		attributes := nestedObject.GetAttributes()
+
+		for k, v := range attributes {
+			d := validateAttributeFieldName(path.AtName(k), k, v)
```
bflad, line 205: Since the nested attributes might be done using list, map, or set mode, we'll also need to account for adding those intermediate steps (AtListIndex, AtMapKey, AtSetValue) to the path. We're not looking at a real configuration though, so it's a little awkward to make it "correct", although Terraform likely won't show the source configuration matching the path information when it is returned by the GetProviderSchema RPC.

Let's create a followup issue, but we should likely try to fix up this path information and consider including it in the diagnostic details in case it is not in Terraform's output. Maybe we can use something like the zero-value (e.g. `AtListIndex(0)`, `AtMapKey("")`, `AtSetValue(/* ? */)`) for describing those intermediate path steps for static validation.

datasource/schema/schema.go
```diff
@@ -122,6 +125,130 @@ func (s Schema) TypeAtTerraformPath(ctx context.Context, p *tftypes.AttributePat
...
+	nestedObject := b.GetNestedObject()
+
+	if nestedObject == nil {
+		return diags
+	}
+
+	blocks := nestedObject.GetBlocks()
+
+	for k, v := range blocks {
+		d := validateBlockFieldName(path.AtName(k), k, v)
```
bflad, line 236: Similarly for block pathing and block nesting modes.

provider/metaschema/schema.go
```diff
@@ -96,6 +99,72 @@ func (s Schema) TypeAtTerraformPath(ctx context.Context, p *tftypes.AttributePat
...
+// Validate verifies that the schema is not using a reserved field name for a top-level attribute.
+func (s Schema) Validate() diag.Diagnostics {
+	var diags diag.Diagnostics
+
+	// Raise error diagnostics when data source configuration uses reserved
+	// field names for root-level attributes.
+	reservedFieldNames := map[string]struct{}{
+		"alias":   {},
+		"version": {},
+	}
```
bflad, line 111: Provider meta schemas relate to the [`terraform` configuration block `provider_meta`](https://developer.hashicorp.com/terraform/internals/provider-meta) so we can skip these `provider` configuration block specific name checks 👍 

```suggestion
```

provider/metaschema/schema.go
```diff
@@ -96,6 +99,72 @@ func (s Schema) TypeAtTerraformPath(ctx context.Context, p *tftypes.AttributePat
...
+
+	for k, v := range attributes {
+		if _, ok := reservedFieldNames[k]; ok {
+			diags.AddAttributeError(
+				path.Root(k),
+				"Schema Using Reserved Field Name",
+				fmt.Sprintf("%q is a reserved field name", k),
+			)
+		}
+
```
bflad, line 123:
```suggestion
```

provider/metaschema/schema_test.go
```diff
@@ -806,3 +807,166 @@ func TestSchemaTypeAtTerraformPath(t *testing.T) {
...
+					"list_nested_attribute": metaschema.ListNestedAttribute{
+						NestedObject: metaschema.NestedAttributeObject{
+							Attributes: map[string]metaschema.Attribute{
+								"alias": metaschema.Int64Attribute{},
+							},
+						},
+					},
+				},
+			},
+		},
```
bflad, line 858:
```suggestion
```

bflad (approved):
Looks good to me 🚀 Just little things to followup on.

datasource/schema/schema.go
bendbennett: https://github.com/hashicorp/terraform-plugin-framework/issues/574

datasource/schema/schema.go
bendbennett: https://github.com/hashicorp/terraform-plugin-framework/issues/574

bflad (approved):

Issue #551, opened 2022-11-28
Amend Plan Resource Change To Allow For Optional Blocks


bendbennett:
### Module version
```
v0.16.0
```

### Relevant provider source code

The issue is described in more detail in [timeouts.Block causes error in modify plan set](https://github.com/hashicorp/terraform-plugin-framework-timeouts/issues/14#top).

Briefly, defining a block within the schema, in this instance, for timeouts and then omitting timeouts from configuration generates an error. 

```go
func (e *exampleResource) GetSchema(ctx context.Context) (tfsdk.Schema, diag.Diagnostics) {
	return tfsdk.Schema{
		/*...*/        
		Blocks: map[string]tfsdk.Block{
			"timeouts": timeouts.Block(
				ctx,
				timeouts.Opts{
					Create: true,
					Update: true,
					Delete: true,
				},
			),
		},
	}, nil
}
```

### Terraform Configuration Files
```terraform
resource "example_resource" "example" {
}
```

### Debug Output

https://gist.github.com/bendbennett/8a9ecd46848abc0aaeebf47129fd2a14

### Expected Behavior

`terraform apply` should have run successfully

### Actual Behavior

The following error is generated:

```shell
│ Error: Error modifying plan
│ 
│   with playground_resource.example,
│   on resource.tf line 9, in resource "playground_resource" "example":
│    9: resource "playground_resource" "example" {
│ 
│ There was an unexpected error updating the plan. This is always a problem with the provider. Please report the following to the provider developer:
│ 
│ AttributeName("timeouts"): couldn't find attribute in resource schema: path leads to block, not an attribute
╵
```

### Steps to Reproduce

Using the schema and configuration listed above and running `terraform apply` should be sufficient to reproduce the issue.

### References

- [timeouts.Block causes error in modify plan set](https://github.com/hashicorp/terraform-plugin-framework-timeouts/issues/14#top)

PR #552, opened 2022-11-28
Amend Plan Resource Change To Allow For Optional Blocks 

bendbennett:
Closes: #551
Closes: hashicorp/terraform-plugin-framework-timeouts#14


internal/fwserver/server_planresourcechange_test.go
```diff
@@ -1935,6 +1976,76 @@ func TestServerPlanResourceChange(t *testing.T) {
...
+								"test_optional_one": tftypes.String,
+								"test_optional_two": tftypes.String,
+							},
+						}, nil),
+					}),
+					Schema: testSchemaBlock,
+				},
+				PriorState: &tfsdk.State{
+					Raw: tftypes.NewValue(testSchemaBlockType, map[string]tftypes.Value{
+						"test_required": tftypes.NewValue(tftypes.String, "test-new-value"),
```
bflad, line 2010: To trigger this new unit test to fail without the fix and therefore verify the fix, there needs to be a difference between the prior state and proposed new state to trigger the call to `MarkComputedNilsAsUnknown`. Adjusting this line should do the trick.

```suggestion
						"test_required": tftypes.NewValue(tftypes.String, "test-value"),
```

The reason this issue occurs upstream is mainly because during resource creation, the prior state is null, but the proposed new state contains blocks.

bflad (approved):
Other than the one test fix and missing changelog, looks good to me 🚀

bflad (approved):

Issue #483, opened 2022-09-14
Computed values within blocks that are nested within set blocks are unset by the framework


liamcervante:
### Module version

```
v0.12.0
```

### Relevant provider source code

[PR reproducing the bug](https://github.com/hashicorp/terraform-provider-corner/pull/89/files)

### Terraform Configuration Files

```hcl
resource "tf6muxprovider_nested" "example" {
  set {
    id = "one"

    list {

    }

    list {

    }
  }

  set {

  }
}
```

### Debug Output
N/A, the error is reproduced in a test case in the PR

### Expected Behavior

The test case in the PR should pass.

### Actual Behavior

```
=== RUN   TestAccResourceNested
    nested_resource_test.go:11: Step 1/1 error: Error running post-apply plan: exit status 1
        
        Error: Error modifying plan
        
          with tf6muxprovider_nested.example,
          on terraform_plugin_test.tf line 2, in resource "tf6muxprovider_nested" "example":
           2: resource "tf6muxprovider_nested" "example" {
        
        There was an unexpected error updating the plan. This is always a problem
        with the provider. Please report the following to the provider developer:
        
        AttributeName("set").ElementKeyValue(tftypes.Object["id":tftypes.String,
        "list":tftypes.List[tftypes.Object["id":tftypes.String]]]<"id":tftypes.String<"xhxKQFDa">,
        "list":tftypes.List[tftypes.Object["id":tftypes.String]]<>>).AttributeName("list"):
        couldn't find attribute in resource schema: path leads to block, not an
        attribute
--- FAIL: TestAccResourceNested (310.30s)
```

### Steps to Reproduce

1. Checkout the branch in terraform-provider-corner (or the forked repo)
2. TF_ACC=1 go test internal/tf6muxprovider/provider1/*.go

### References

- https://github.com/hashicorp/terraform-plugin-framework/issues/479

liamcervante:
Some other observations.

If you replace the resource with any of the following the test works fine:

```hcl
resource "tf6muxprovider_nested" "example" {
  set {
    id = "one"

    list {

    }

    list {

    }
  }

  set {
    id = "two"
  }
}
```

```hcl
resource "tf6muxprovider_nested" "example" {
  set {

    list {

    }

    list {

    }
  }

  set {

  }
}
```

```
resource "tf6muxprovider_nested" "example" {
  set {
    id = "one"

    list {
      id = "one"
    }

    list {
      id = "two"
    }
  }

  set {

[...]

PR #555, opened 2022-11-28
Skip blocks while marking computed nils unknown

jar-b:
Relates #483 
Relates https://github.com/hashicorp/terraform-provider-corner/pull/89
Relates https://github.com/hashicorp/terraform-provider-aws/pull/27857 (acceptance tests pass after this fix)

`MarkComputedNilsAsUnknown` now passes block values through unmodified, rather than throwing an error. 

I extended the single large test object to add block attributes, but I can also split into individual test cases if that'd be helpful. Original issue is reproduced with the updated test case and the fix commented out:

```console
$ go test -count=1 -run "^TestMarkComputedNilsAsUnknown$" ./internal/fwserver/...
--- FAIL: TestMarkComputedNilsAsUnknown (0.00s)
    server_planresourcechange_test.go:283: Unexpected error: AttributeName("block-nil-optional-computed"): couldn't find attribute in resource schema: path leads to block, not an attribute
FAIL
FAIL    github.com/hashicorp/terraform-plugin-framework/internal/fwserver       0.353s
FAIL
```

---
**Edit:** #552 included the change to ignore block type errors, so this PR has been updated to just extend test coverage.

### DIFF
--- a/internal/fwserver/server_planresourcechange_test.go
+++ b/internal/fwserver/server_planresourcechange_test.go
@@ -115,6 +115,41 @@ func TestMarkComputedNilsAsUnknown(t *testing.T) {
 				Computed: true,
 			},
 		},
+		Blocks: map[string]tfsdk.Block{
+			// nil blocks should remain nil
+			"block-nil-optional-computed": {
+				Attributes: map[string]tfsdk.Attribute{
+					"string-nil": {
+						Type:     types.StringType,
+						Optional: true,
+						Computed: true,
+					},
+					"string-set": {
+						Type:     types.StringType,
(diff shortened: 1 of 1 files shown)


bflad:
Hi @jar-b 👋 Thank you for submitting this change. This appears to overlap with #552 which was recently merged, however this change does include additional `TestMarkComputedNilsAsUnknown` testing. If you're willing or able, I'd suggest rebasing this against `main` and dropping the (now duplicate) changelog entry.

bflad (approved):
Looks good to me, thanks so much @jar-b 🚀

Issue #560, opened 2022-11-29
schema: Fix nesting mode for data source, resource and provider map, set and single nested attribute


bendbennett:
### Module version
```
main @55244fe874b7c426a0ae430e3a6c67d768a40120
```

### Relevant source code

The nesting mode returned is incorrect for the type (i.e., `NestingModeList` is returned for `SetNestedAttribute`).

```go
// GetNestingMode always returns NestingModeList.
func (a SetNestedAttribute) GetNestingMode() fwschema.NestingMode {
	return fwschema.NestingModeList
}
```

PR #561, opened 2022-11-29
Fix nesting mode for map, set and single nested attribute within data source, provider and resource schema

bendbennett:
Closes: #560


bflad (approved):
Thanks for this!

PR #565, opened 2022-11-30
resource/schema: New packages which contain type-specific schema plan modifier implementations

bflad:

When developers migrate from `tfsdk.Schema` to `resource/schema.Schema`, they will need to also migrate to the type-specific plan modifiers. New packages have been introduced under `resource/schema`, such as `resource/schema/stringplanmodifier`. The existing plan modifiers are deprecated, similar to the `tfsdk` schema handling.

Of note, the `Computed` handling is no longer present in any of the `RequiresReplace*` plan modifiers. The configuration value null check has also been moved into a separate `RequiresReplaceIfConfigured` plan modifier. Now the `RequiresReplace` plan modifier operates unconditionally, assuming the resource is being updated and there is a difference between plan and prior state values.


bendbennett (approved):
LGTM 👍

PR #573, opened 2022-12-09
fixes #572

### LINKED ISSUE #572: StateUpgrader only works with deprecated APIs
### Module version

```
github.com/hashicorp/terraform-plugin-framework v0.17.0
```

```go
return map[int64]resource.StateUpgrader{
  0: {
    PriorSchema: &schema.Schema{
	"id": schema.StringAttribute{
          Required: true,
	},
      },
    },
    StateUpgrader: func(...
```

```hcl
N/A
```

### Expected Behavior

The new separated APIs should work with StateUpgraders, i'm assuming this hasn't been done as it's planned for the removal of the deprecated APIs.

### Actual Behavior

Doesn't compile.

iwarapter:
Switches from the deprecated API to the new one.

### DIFF
--- a/internal/fwserver/server_upgraderesourcestate_test.go
+++ b/internal/fwserver/server_upgraderesourcestate_test.go
@@ -17,7 +17,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
-	"github.com/hashicorp/terraform-plugin-framework/types"
 )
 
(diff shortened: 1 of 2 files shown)


bflad (approved):
Thank for you for this contribution, @iwarapter! The code changes here look good for our next release, which will remove `tfsdk.Schema`.

If you have a second, could you also create a changelog entry by adding a `.changelog/573.txt` file with something like the below?

``````markdown
```release-note:breaking-change
resource: The `StateUpgrader` type `PriorSchema` field type has been migrated from `tfsdk.Schema` to `resource/schema.Schema`, similar to other resource schema handling
```
``````

Otherwise, we'll handle that post-merge. Thanks again!

bflad (approved):

PR #585, opened 2022-12-14
website: Align Attributes page with schema attribute types

bflad:
The content was still based on the prior `tfsdk.Schema` way. With the split attribute types, we can now refer provider developers to those types, rather than the framework type system types.


bendbennett (approved):
LGTM 👍

PR #586, opened 2022-12-14
website: Further clarify configuration validation and when API access is available

bflad:
Given that configuration validation occurs before any provider configuration, it is important to note that ordering in resource and data source documentation. Hopefully the information callout can prevent confusion and point developers in the right direction.

This also adjusts the resource plan modification page to mention returning diagnostics more generically, rather than the specific Terraform 1.3+ planned resource destruction feature.


bendbennett (approved):
LGTM 👍

PR #598, opened 2022-12-21
website: Point overview muxing link at migration guide content

bflad:
The overview page tries to hint that muxing can be used to iteratively migrate providers from SDK to framework, however the current link points to the generic mux documentation. The migration guide providers page has excellent content that is specific for this situation so it seems ideal to point there (which also links out to the generic mux documentation) rather than giving developers less directly useful information.


bendbennett (approved):
LGTM 👍

PR #602, opened 2022-12-22
tfsdk+types: Fix ObjectType missing attribute error and PathMatches schema validation

### LINKED ISSUE #600: path: Valid Expressions for the Schema with No Matches Should Not Return Error
### Module version

```
v1.0.1
```

```go
func (d ThingDataSource) Schema(ctx context.Context, req datasource.SchemaRequest, resp *datasource.SchemaResponse) {
	resp.Schema = schema.Schema{
		Blocks: map[string]schema.Block{
			"example": schema.SetNestedBlock{
				NestedObject: schema.NestedBlockObject{
						"nested_example": schema.StringAttribute{
							Required:    true,
						},
					},
				},
			},
			/* ... */
		},
		/* ... */
	},
}
```

Along with configuration validators that use path expressions, e.g.

```go
datasourcevalidator.Conflicting(
	// AtAnySetValue being tested because of Terraform core logic to never send blocks as null
	path.MatchRoot("example").AtAnySetValue(),
	/* ... */
)
```

And no configuration value (as shown below).

```hcl
data "examplecloud_thing" "example" {
  # no example block configuration
}
```

### Expected Behavior

The configuration validator should work insofar as it would not return a path expression error for this situation, and instead return no path expression matches and therefore no error from the validator.

### Actual Behavior

```
│ Error: Invalid Path Expression for Schema Data
│ 
│   with data.examplecloud_thing.example,
│   on config.tf line ##, in data "examplecloud_thing" "example":
│   ##: data "examplecloud_thing" "example" {
│ 
│ The Terraform Provider unexpectedly matched no paths with the given path expression and current schema data. This can happen if the path expression does not correctly follow the
│ schema in structure or types. Please report this to the provider developers.
```

Which is raised here: https://github.com/hashicorp/terraform-plugin-framework/blob/27aff715d3d53b6d052358f3e4f1be6573f8a605/internal/fwschemadata/data_path_matches.go#L64-L74

The framework should perform this type of path expression to schema validation outside whether or not any matches are returned.  Empty collection values with deeper transversals are likely the most problematic case in this situation.

### Steps to Reproduce

1. `terraform apply`

### References

- https://github.com/hashicorp/terraform-plugin-framework-validators/issues/91

bflad:
Closes #600

This change will move the `PathMatches` expression to schema validation to be performed upfront and allow zero matches to be returned. Previously, valid expressions could be rejected errantly in cases of empty collection type values where the expression steps traversed deeper. Consumers of `PathMatches` should not be returned errors in those cases and instead will receive zero path matches, which they can choose whether that result is significant or not. Invalid expressions for the schema will still receive errors.

While implementing the new expression to schema validation logic, the `basetypes.ObjectType` implementation was missing an error return for missing attributes which caused false positives. This was actually noticed previously, however a covering issue was not created to fix the behavior, but it is now being fixed as part of correctly implementing the expression to schema validation. Returning an error in this situation is considered the valid implementation detail as object types are statically defined based on the underlying map of attribute names to attribute types. It is never valid to step into a non-existent attribute name.


bendbennett (approved):
LGTM 👍

PR #604, opened 2023-01-03
all: Convert empty list and set block values to null

### LINKED ISSUE #601: Consider Converting List/Set Block Data With No Objects to Null List/Set
### Module version

```
v1.0.1
```

### Use-cases

Blocks in Terraform are considered _structural_ meaning that they technically do not fit the definition of having a null value. When provider developers create a schema such as:

```go
func (d ThingDataSource) Schema(ctx context.Context, req datasource.SchemaRequest, resp *datasource.SchemaResponse) {
	resp.Schema = schema.Schema{
		Blocks: map[string]schema.Block{
			"example": schema.SetNestedBlock{
				NestedObject: schema.NestedBlockObject{
						"nested_example": schema.StringAttribute{
							Required:    true,
						},
					},
				},
			},
			/* ... */
		},
		/* ... */
	},
}
```

The set value will never itself be null, even with a configuration omitting the block wholly. This can be confusing for provider developers since attributes (nested or not) do not work the same way. Consider these situations for unconfigured blocks:

- Using a data model type and calling `Get()` will always return a known, empty collection value
- Calling `GetAttribute()` will always return a known, empty collection value
- Using configuration validators cannot just check the collection type `IsNull()` method without understanding the semantic difference with blocks. Instead, underlying elements (objects) would have to be checked for length.
- Similarly, using resource plan modifiers cannot just check the collection type `IsNull()` method

Ideally, working with attribute and block values that are unconfigured should be very similar in this particular case.

### Proposal

The `internal/fromproto*` logic for data handling may be able to catch the scenario of a block collection type being empty and convert the collection type to null automatically. This would smooth over the differences between attribute data and block data in the framework while (theoretically?) not losing _semantic_ accuracy. Given the framework is a higher level SDK, this automatic value translation should be an acceptable pragmatic exception to the general design preference of the framework to not introduce special behaviors on top of the provider protocol.

Terraform automatically will "fix" provider data that returns null collection types for blocks, so reverse handling for responses in `internal/toproto*` is not strictly necessary.

### References

- https://github.com/hashicorp/terraform/pull/32424
- https://github.com/hashicorp/terraform-plugin-framework-validators/issues/91
- https://github.com/hashicorp/terraform-plugin-framework/issues/600

### LINKED ISSUE #606: List/Set Blocks with DeprecationMessage Always Return Diagnostic
### Module version

```
v1.0.1
```

```go
func (r ThingResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
	resp.Schema = schema.Schema{
		// ...
		Blocks: map[string]schema.Block{
			"example_block": schema.ListNestedBlock{
				NestedObject: schema.NestedBlockObject{
						"example_attribute": schema.BoolAttribute{
							Optional: true,
						},
					},
				},
				DeprecationMessage: "oh no",
			},
		},
	}
}
```

```hcl
resource "examplecloud_thing" "example" {
  # no example_block configuration :)
}
```

### Expected Behavior

No deprecation warning diagnostic if the list/set block is not configured.

### Actual Behavior

Deprecation warning diagnostic is always returned to practitioners, regardless of the list/set block being configured or not.

The reason this occurs is because the framework logic uses the following check to generate the diagnostic:

https://github.com/hashicorp/terraform-plugin-framework/blob/51bb0a7f377cb95a513501119ce01be87f8351fa/internal/fwserver/block_validation.go#L161-L168

While Terraform always sends list/set blocks as known (with zero elements when unconfigured). We already have unit testing for this behavior:

https://github.com/hashicorp/terraform-plugin-framework/blob/51bb0a7f377cb95a513501119ce01be87f8351fa/internal/fwserver/block_validation_test.go#L30-L195

But the assumption about how Terraform sends the data over the protocol is why this was implemented incorrectly the first time.

### Steps to Reproduce

I confirmed this behavior via an acceptance test and found the logs:

```
2023-01-05T16:24:01.527-0500 [TRACE] sdk.proto: Received downstream response: diagnostic_warning_count=1 tf_req_duration_ms=0 tf_req_id=8ad3f97f-6ac3-b6c2-fdfb-8cc6cdbfc0fb tf_provider_addr=registry.terraform.io/examplecloud/examplecloud  tf_proto_version=6.3 tf_rpc=ValidateResourceConfig tf_resource_type=examplecloud_thing diagnostic_error_count=0
2023-01-05T16:24:01.527-0500 [WARN]  sdk.proto: Response contains warning diagnostic: tf_proto_version=6.3 diagnostic_summary="Block Deprecated" tf_req_id=8ad3f97f-6ac3-b6c2-fdfb-8cc6cdbfc0fb tf_provider_addr=registry.terraform.io/examplecloud/examplecloud_thing tf_rpc=ValidateResourceConfig tf_resource_type=examplecloud_thing diagnostic_detail="oh no" diagnostic_severity=WARNING diagnostic_attribute="AttributeName(\"example_block\")"
```

But any regular `terraform plan` of an installed provider with this schema setup should trigger the diagnostic in the human readable output.

### References

- Would be fixed by https://github.com/hashicorp/terraform-plugin-framework/issues/601 / https://github.com/hashicorp/terraform-plugin-framework/pull/604

bflad:
Closes #601
Closes #606 

Please note that this change requires additional _design_ consideration.

It may be desirable since:

- Partially increases parity with terraform-plugin-sdk, that automatically handled this conversion in its shims.
- Consistency between attribute and block values, removing Terraform's lower level data differences which should be inconsequential for provider developers.
- Workarounds for situations such as https://github.com/hashicorp/terraform-plugin-framework-validators/issues/91 should not be required.

It may be undesirable since:

- Increases differences between the protocol and what is surfaced to providers.
- Providers using custom type slices for `Get()` or `GetAttribute()` may/will receive a `nil` instead of an empty slice, which in some limited code situations may introduce a new provider logic panic. Provider logic using `for ... range` loops or `len()` checks will not have behavior changes though.


internal/fwschemadata/data_nullify_collection_blocks.go
```diff
@@ -0,0 +1,81 @@
...
+
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/internal/fromtftypes"
+	"github.com/hashicorp/terraform-plugin-framework/internal/fwschema"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+)
+
+// NullifyCollectionBlocks returns a copy of the Data where list and set block
+// empty values are converted to null values.
+func (d Data) NullifyCollectionBlocks(ctx context.Context) (Data, diag.Diagnostics) {
```
bendbennett, line 14: Believe that `Data` has methods on both value and pointer receivers.

bendbennett (approved):
LGTM 👍

internal/fwschemadata/data_nullify_collection_blocks.go
bflad: Updated to a pointer receiver method and removed the copying/return.

mvantellingen:
Just upgraded an in-progress provider to the latest version of git (`6139030b2d0e`) but now running into the following error:

```
attribute representing a list of nested blocks must be empty to indicate no blocks, not null.
```

I've figured this change might be responsible. Running Terraform v1.3.6

bflad:
@mvantellingen thank you for the heads up. Could you open a bug report with how to produce that error? We might need to implement the reverse logic for this when we send responses, which is unexpected after discussing this previously with the Terraform maintainers.

PR #608, opened 2023-01-06
.github/workflows: Fix label name in issue comment workflow

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/labels/waiting-response

The `stale` label is there should we ever want to use actions/stale (similar to other HashiCorp projects).


bendbennett (approved):
LGTM 👍

PR #610, opened 2023-01-06
ci: Enable paralleltest linter

bflad:
Reference: https://github.com/kunwardeep/paralleltest

The upstream issue with map-based `range` false positives appears to be resolved.


bendbennett (approved):
LGTM 👍

PR #614, opened 2023-01-09
types/basetypes: Fixed Float64Type validation to handle valid, stringified numbers from Terraform

### LINKED ISSUE #613: Float64 Type Validation Error (123 and 123.5 work, but 123.2 fails)
I'm trying to write my own provider with the plugin framework, but the validation on `schema.Float64Attribute` fails.

### Module version

```
github.com/hashicorp/terraform-plugin-framework v1.0.1
```

```go
"latitude": schema.Float64Attribute{
	Description:         "Latitude of the resource location",
	MarkdownDescription: "Latitude of the resource location",
	Optional:            true,
},
```

```hcl
latitude  = "123.2"
```

### Expected Behavior

The expected result should be, that floating point numbers like latitude and longitude in this case should be accepted.

### Actual Behavior

Terraform validation fails, when I try to set a value, which is not a “whole” number or a float ending with “.5” So these numbers work (123, 123.5), but all other numbers like (123.1, 123.2, etc.) fail. 

Error message:
```
Float64 Type Validation Error: Value %!s(*big.Float=123.6) cannot be represented as a 64-bit floating point.
```

### References

I posted this on the Hashicorp discuss forum as well: 
https://discuss.hashicorp.com/t/terraform-plugin-framework-float64-type-validation-error/48645

bflad:
Closes #613

Numbers in Terraform's type system can be represented with up to 512 bits of base 10 precision. When encoding and decoding across the protocol, the type system or MessagePack may switch the representation to be stringified, which is valid and expected.

The prior `Float64Type` type `Validate` method logic was only checking the `big.Accuracy` return from the value, however in certain cases the stringified number was saved with a `big.Below` or `big.Above` accuracy, even though the value itself could be successfully converted into a `float64` type. This change updates the validation logic to follow the `(*big.Float).Float64()` method documentation to check the returned value, along with the `big.Accuracy`.

Previously before logic updates:

```
--- FAIL: TestFloat64TypeValidate (0.00s)
    --- FAIL: TestFloat64TypeValidate/positive-string-float (0.00s)
        /Users/bflad/src/github.com/hashicorp/terraform-plugin-framework/types/basetypes/float64_test.go:98: unexpected difference:   diag.Diagnostics(
            - 	{
            - 		diag.withPath{
            - 			Diagnostic: diag.ErrorDiagnostic{
            - 				detail:  "Value %!s(*big.Float=123.2) cann"...,
            - 				summary: "Float64 Type Validation Error",
            - 			},
            - 			path: s"test",
            - 		},
            - 	},
            + 	nil,
              )
    --- FAIL: TestFloat64TypeValidate/negative-string-float (0.00s)
        /Users/bflad/src/github.com/hashicorp/terraform-plugin-framework/types/basetypes/float64_test.go:98: unexpected difference:   diag.Diagnostics(
            - 	{
            - 		diag.withPath{
            - 			Diagnostic: diag.ErrorDiagnostic{
            - 				detail:  "Value %!s(*big.Float=-123.2) can"...,
            - 				summary: "Float64 Type Validation Error",
            - 			},
            - 			path: s"test",
            - 		},
            - 	},
            + 	nil,
              )
    --- FAIL: TestFloat64TypeValidate/MaxFloat64-above (0.00s)
        /Users/bflad/src/github.com/hashicorp/terraform-plugin-framework/types/basetypes/float64_test.go:98: unexpected difference:   diag.Diagnostics(
            - 	{
            - 		diag.withPath{
            - 			Diagnostic: diag.ErrorDiagnostic{
            - 				detail:  "Value %!s(*big.Float=2.797693135"...,
            - 				summary: "Float64 Type Validation Error",
            - 			},
            - 			path: s"test",
            - 		},
            - 	},
            + 	nil,
              )
    --- FAIL: TestFloat64TypeValidate/SmallestNonzeroFloat64-below (0.00s)
        /Users/bflad/src/github.com/hashicorp/terraform-plugin-framework/types/basetypes/float64_test.go:98: unexpected difference:   diag.Diagnostics(
            - 	{
            - 		diag.withPath{
            - 			Diagnostic: diag.ErrorDiagnostic{
            - 				detail:  "Value %!s(*big.Float=3.940656458"...,
            - 				summary: "Float64 Type Validation Error",
            - 			},
            - 			path: s"test",
            - 		},
            - 	},
            + 	nil,
              )
```


bendbennett (approved):
LGTM 👍

PR #621, opened 2023-01-12
internal/toproto*: Convert null list/set block values to empty

### LINKED ISSUE #620: internal/toproto*: Convert Null List/Set Blocks to Empty
### Module version

```
faab4e35bd4e13c6f0bff2c8223c9c52996304c2 (post 1.0.1; pre 1.1.0)
```

Any schemas containing `ListNestedBlock` or `SetNestedBlock`.

Any configuration without any of said blocks.

### Expected Behavior

Providers should not receive any errors when working with null/empty collection blocks in the framework.

### Actual Behavior

https://github.com/hashicorp/terraform-plugin-framework/pull/604#issuecomment-1376340914

```
attribute representing a list of nested blocks must be empty to indicate no blocks, not null.
```

This error from Terraform core originates in the plan validity logic, which is explicitly checking list/set nesting mode blocks for null values. This logic differs from some of the Terraform core _state_ logic, which can automatically perform some of the null to empty conversions.

### Steps to Reproduce

Presumably can happen on any `terraform plan` or `terraform apply` in the given situation.

### References

- https://github.com/hashicorp/terraform-plugin-framework/issues/601
- https://github.com/hashicorp/terraform-plugin-framework/pull/604

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/pull/604
Closes #620

This logic implements the inverse of what was added to `internal/fromproto*` where empty block values are converted to null. While Terraform implements some logic to automatically convert these for state responses from `ReadResource` and others, the same null values can cause errors with Terraform's plan validity logic.

This fixes the potential new error caused by only introducing the `internal/fromproto*` logic, which was not released and why it does not contain a changelog entry in that regard, However, this change is considered an overall enhancement prior and including those changes since it means the provider developers will no longer be burdened with understanding Terraform's implementation details with null versus empty collection blocks when setting values for responses.


bendbennett (approved):
LGTM 👍

PR #632, opened 2023-01-20
ci: added `forcetypeassert` lint rule + resolved all lint warnings

austinvalle:
contributes to: https://github.com/hashicorp/terraform-providers-devex-internal/issues/102

I used the surrounding code's context as clues to how to handle the type assertions, lmk if there are more appropriate ways to handle them.

### DIFF
--- a/internal/fwserver/server_planresourcechange.go
+++ b/internal/fwserver/server_planresourcechange.go
@@ -44,7 +44,7 @@ func (s *Server) PlanResourceChange(ctx context.Context, req *PlanResourceChange
 		return
 	}
 
-	if _, ok := req.Resource.(resource.ResourceWithConfigure); ok {
+	if resourceWithConfigure, ok := req.Resource.(resource.ResourceWithConfigure); ok {
 		logging.FrameworkTrace(ctx, "Resource implements ResourceWithConfigure")
 
 		configureReq := resource.ConfigureRequest{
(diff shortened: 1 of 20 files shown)


internal/fwserver/server_planresourcechange.go
```diff
@@ -277,7 +277,7 @@ func MarkComputedNilsAsUnknown(ctx context.Context, config tftypes.Value, resour
 		if err != tftypes.ErrInvalidStep && err != nil {
 			logging.FrameworkError(ctx, "error walking attribute path")
 			return val, err
-		} else if err != tftypes.ErrInvalidStep && !configVal.(tftypes.Value).IsNull() {
+		} else if err != tftypes.ErrInvalidStep && !configVal.(tftypes.Value).IsNull() { //nolint:forcetypeassert // Not sure the best comment for here? Should we handle this?
```
austinvalle, line 280: I wasn't sure how to handle this one, any thoughts on an appropriate log or skip message?

bflad: It was likely written this way because `tftypes.WalkAttributePath()` on a `tftypes.Value` should be guaranteed to return a `tftypes.Value`: https://pkg.go.dev/github.com/hashicorp/terraform-plugin-go/tftypes#WalkAttributePath (there's a lot of overly generic code in `tftypes` 🙁 )

I think its fine to fix up the logic to be a little cleaner for type safety and present an error if the unexpected situation occurs:

```go
configValIface, _, err := tftypes.WalkAttributePath(config, path)

if err != nil && err != tftypes.ErrInvalidStep {
	logging.FrameworkError(ctx,
		"Error walking attribute/block path during unknown marking",
		map[string]any{
			logging.KeyError: err.Error(),
		},
	)

	return val, fmt.Errorf("error walking attribute/block path during unknown marking: %w", err)
}

configVal, ok := configValIface.(tftypes.Value)

if !ok {
	return val, fmt.Errorf("Unexpected type during unknown marking: %T", configValIface)
}

if !configVal.IsNull() {
	logging.FrameworkTrace(ctx, "Attribute/block not null in configuration, not marking unknown")

	return val, nil
}
```

bflad (approved):
Looks good to me 🚀

PR #645, opened 2023-02-02
add better reflect error message

ryan-berger:


bflad (commented):
Hi @ryan-berger 👋 Thank you for this submission!

There is currently no unit testing covering this specific error messaging, which is totally okay, but I wonder if it is removing the `name` from the message, since the `path` should already include that information from line 168.

What do you think?

ryan-berger:
Hi @bflad ! That sounds fine to me.

In reference to your [comment](https://github.com/hashicorp/terraform-plugin-framework/pull/645#pullrequestreview-1281163328) does this fix then need to happen somewhere in here? https://github.com/hashicorp/terraform-plugin-framework/blob/6139030b2d0ed6df585faf8682310746b147bbfa/diag/attribute_error_diagnostic.go#L8

bflad:
Oh sorry, @ryan-berger, for my unclear ask. 😅 I was intending to ask whether you thought it would be acceptable if we adjusted the details here to:

```go
err := fmt.Errorf("couldn't find type information for attribute at %s in supplied attr.Type %T", path, typ)
```

So the sentence will potentially be less redundant:

```text
couldn't find type information for attribute at example_single_nested_attr.nested_attr in supplied attr.Type basetypes.ObjectType
```

With `name` in there as it is currently in the PR, it would show as:

```text
couldn't find type information for attribute nested_attr at example_single_nested_attr.nested_attr in supplied attr.Type basetypes.ObjectType
```

No need to worry about adjusting anything else. A lot of times when we adjust error messaging in this codebase one of the existing unit tests will pick up the sentence changes, but it seems there's nothing yet explicitly triggering this message in a unit test. If you are interested in adding something like that, it would be adding something in `internal/reflect/struct_test.go`, but totally optional because the reflection part of the codebase is much more difficult to work on for an outsider.

ryan-berger:
No problem @bflad ! I wasn't sure if there was some sort of systemic thing that needed to be changed so that the `path` was shown for all `AttributeErrorDiagnostic`s.

I agree it is a redundant message, so I've removed the `name` and just have the `path`. Thanks for your help!

bflad:
Ah ha okay -- changing it globally, while a good idea, could make other existing messages confusing (or themselves redundant if they already include path information) so I think if we were going to embark on that level of change we'd want to think about the problem space and make a more formal proposal first. 👍

bflad (approved):
Looks good to me, thanks @ryan-berger! 🚀

PR #648, opened 2023-02-03
types/basetypes: Fixed Float64Type value conversion to handle stringified numbers

### LINKED ISSUE #647: Float64 - Only whole and half (.5) numbers are accepted, others `cannot be represented as a 64-bit floating point`.
This bug is related or (almost) the same as #613 

### Module version

```
github.com/hashicorp/terraform-plugin-framework v1.1.1
```

```go
"latitude": schema.Float64Attribute{
	Description:         "Latitude of the resource location",
	MarkdownDescription: "Latitude of the resource location",
	Optional:            true,
},
```

```hcl
latitude  = "123.2"
```

### Expected Behavior

The expected result should be, that floating point numbers like latitude and longitude in this case should be accepted.

### Actual Behavior

Terraform validation fails, when I try to set a value, which is not a “whole” number or a float ending with “.5” So these numbers work (123, 123.5), but all other numbers like (123.1, 123.2, etc.) fail. 

Error message:
```
Error: Value %!s(*big.Float=51.12) cannot be represented as a 64-bit floating point.
```

### References

- #613 

I posted this on the Hashicorp discuss forum as well: 
https://discuss.hashicorp.com/t/terraform-plugin-framework-float64-type-validation-error/48645

austinvalle:
closes #647 

This is a continuation of the fix implemented in #614 
> _Numbers in Terraform's type system can be represented with up to 512 bits of base 10 precision. When encoding and decoding across the protocol, the type system or MessagePack may switch the representation to be stringified, which is valid and expected._

The `ValueFromTerraform` function has the same accuracy checking logic, so this is just bringing the logic into sync with the `Validate` function.

I checked the rest of the codebase for similar accuracy checks and the only other one seems to be handling it correctly: https://github.com/hashicorp/terraform-plugin-framework/blob/e23d2856504bc02aa596ea7d7be93405b1b100b7/internal/reflect/number.go#L186

### DIFF
--- a/types/basetypes/float64_test.go
+++ b/types/basetypes/float64_test.go
@@ -147,6 +147,39 @@ func TestFloat64TypeValueFromTerraform(t *testing.T) {
 			input:       tftypes.NewValue(tftypes.String, "oops"),
 			expectedErr: "can't unmarshal tftypes.String into *big.Float, expected *big.Float",
 		},
+		// Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/647
+		"zero-string-float": {
+			input:       tftypes.NewValue(tftypes.Number, testMustParseFloat("0.0")),
+			expectation: NewFloat64Value(0.0),
+		},
+		"positive-string-float": {
+			input:       tftypes.NewValue(tftypes.Number, testMustParseFloat("123.2")),
+			expectation: NewFloat64Value(123.2),
+		},
+		"negative-string-float": {
+			input:       tftypes.NewValue(tftypes.Number, testMustParseFloat("-123.2")),
+			expectation: NewFloat64Value(-123.2),
+		},
+		// Reference: https://pkg.go.dev/math/big#Float.Float64
+		// Reference: https://pkg.go.dev/math#pkg-constants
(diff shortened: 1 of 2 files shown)


austinvalle:
Here's a test I ran with my sandbox provider:

#### Schema
```go
"float64_attribute": schema.Float64Attribute{
   Optional: true,
}
```

#### Config
```terraform
resource "example_resource" "example" {
  float64_attribute = "1234.2"
}
```

#### Output
```bash
 $ terraform apply -auto-approve

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the
following symbols:
  + create

  # example_resource.example will be created
  + resource "example_resource" "example" {
      + float64_attribute = 123.2
      + id                = (known after apply)
    }

Plan: 1 to add, 0 to change, 0 to destroy.
example_resource.example: Creating...
example_resource.example: Creation complete after 3s [id=example-id]

Apply complete! Resources: 1 added, 0 changed, 0 destroyed.
```

bflad (approved):
Looks good to me 🚀

bendbennett (approved):
Nice 👍

Issue #659, opened 2023-02-08
Add more prescriptive documentation for mapping between SDKv2 and the Framework 


bendbennett:
### Module version
```
v1.1.1
```

### Use-cases

When migrating a provider from SDKv2 to the Framework, it can be a little ambiguous as to how specific field should be mapped. For instance, SDKv2 allows defining [RequiredWith](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema#Schema.RequiredWith) as a field on [schema.Schema](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema#Schema).

Benefit could be derived from providing more prescriptive guidance during the migration of a provider from SDKv2 to the Framework as to how specific [schema.Schema](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema#Schema) fields defined within an SDKv2 version of a provider should be mapped into a provider written with the Framework.  

### Attempted Solutions

Searching the Framework documentation for [Migrating from SDK](https://developer.hashicorp.com/terraform/plugin/framework/migrating).

### Proposal

Update the Framework documentation for [Migrating from SDK](https://developer.hashicorp.com/terraform/plugin/framework/migrating) to include prescriptive guidance.

### References

- #657

Issue #657, opened 2023-02-07
Add documentation for RequiredWith migration from SDKv2 to plugin framework


remyleone:
### Module version

```
...
```

bflad:
Hi @remyleone 👋 Thank you for raising this. If you have any specific suggestions, please let us know.

There's currently the [SDK to framework migration guide](https://developer.hashicorp.com/terraform/plugin/framework/migrating), but once it gets to individual `schema.Schema`  fields, the documentation does split up a little bit depending on what you're trying to do. For example with `RequiredWith`, it is mentioned on this migration page: https://developer.hashicorp.com/terraform/plugin/framework/migrating/attributes-blocks/validators-predefined

But it doesn't say the exact solutions available for that particular field:

- If desired to leave the validation per-attribute, then it is implementing one of the type-specific `AlsoRequires` validators, such as https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework-validators@v0.9.0/stringvalidator#AlsoRequires
- If desired to move the validation to be declared at the resource level, then it is implementing https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework-validators@v0.9.0/stringvalidator#AlsoRequires

@bendbennett do you think we should have https://developer.hashicorp.com/terraform/plugin/framework/migrating/attributes-blocks/fields and/or https://developer.hashicorp.com/terraform/plugin/framework/migrating/attributes-blocks/validators-predefined be a little bit more prescriptive about the mapping between the old and new? It may be useful to have tables, such as::

| SDK Attribute Field | Framework Attribute Validator | Framework Resource Validator |
|---|---|---|
| `ConflictsWith` | `{TYPE}validator.ConflictsWith()` | `resourcevalidator.Conflicting()` |

etc.

remyleone:
It would be advantageous to have a table that lists each field and how you should translate it into the framework. I've been looking around many pages to see how I should try out to translate a given pattern. In particular, having a linter tool that offers automated suggestions/helps even if it is very simplistic would help give confidence that it could be achieved. While waiting for an automated tool to appear a table would be very much appreciated and offer great help to map possible migration.

bendbennett:
[...]

Issue #650, opened 2023-02-03
Add documentation for migrating from `Required` block to new validators


austinvalle:
### Module version

```
v1.1.1
```

### Use-cases

Currently, blocks that are written in SDKv2 can have a `Required` field applied, but there is no equivalent field or validator for `SingleNestedBlock`, `SetNestedBlock`, or `ListNestedBlock`. We are adding validators to solve this issue, but we should document this in our migration guide.

### Proposal

- After creating the new validators, document in: https://developer.hashicorp.com/terraform/plugin/framework/migrating/attributes-blocks/blocks
- Mention the validators that can be used with each block type as well as plug to not use blocks if possible 😄 

### References
- https://github.com/hashicorp/terraform-plugin-framework-validators/issues/105

bflad:
Now that terraform-plugin-framework-validators v0.10.0 is released with the functionality for this, it may be worth including this as part of #660

PR #660, opened 2023-02-08
Documentation for mapping between SDKv2 and the Framework

bendbennett:
Closes: #664 
Closes: #659 
Closes: #657 
Closes: #650


website/docs/plugin/framework/migrating/attributes-blocks/validators-predefined.mdx
```diff
@@ -13,6 +13,20 @@ the Framework documentation for details. Refer to the
...
+[resource validator](/plugin/framework/resources/validate-configuration) for declaring validation at the level of the
+data source, provider or resource, respectively.
+
+| SDK Attribute Field | Framework Attribute Validator   | Framework Data Source, Provider and Resource Validators                                                                                                                             |
+|---------------------|---------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------|
+| AtLeastOneOf        | {TYPE}validator.AtLeastOneOf()  | [datasourcevalidator.AtLeastOneOf()](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework-validators/datasourcevalidator#AtLeastOneOf), [providervalidator.AtLeastOneOf()](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework-validators/providervalidator#AtLeastOneOf), [resourcevalidator.AtLeastOneOf()](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework-validators/resourcevalidator#AtLeastOneOf)         |
```
remyleone, line 25: Would some html be useful here to have a list for the cases where there are several translation available: https://stackoverflow.com/questions/19950648/how-to-write-lists-inside-a-markdown-table

bflad: I'm not sure how well our Markdown rendering engine for the website plays with intermixed HTML/Markdown syntax. Another option here could be having separate table columns for data source, provider, and resource to prevent the listing in general.

bflad:
I'm wondering if we might also want a table on https://developer.hashicorp.com/terraform/plugin/framework/migrating/attributes-blocks/fields that accounts for all the `schema.Schema` since that may be where folks first wind up when looking for this type of information. That framework side of that table could point folks to either the direct answer (e.g. `Required` for `Required`) or link to the other migration pages:

| SDK Schema Field | Framework |
|---|---|
| `Required` | `Required` field on attribute |
| `ValidateFunc` | [Predefined Validators](/terraform/plugin/framework/migrating/attributes-blocks/validators-predefined) and [Custom Validators](/terraform/plugin/framework/migrating/attributes-blocks/validators-custom) | 

It may also be worth considering something similar for https://developer.hashicorp.com/terraform/plugin/framework/migrating/attributes-blocks/blocks since they have a differing migration path -- e.g. https://github.com/hashicorp/terraform-plugin-framework/issues/650

website/docs/plugin/framework/migrating/attributes-blocks/validators-predefined.mdx
bendbennett: Have split into 3 columns for data source, provider and resource.

website/docs/plugin/framework/migrating/attributes-blocks/fields.mdx
```diff
@@ -9,6 +9,38 @@ description: >-
...
+| Description           | `Description` field on attribute                                                                                                                                                                                                                                                                                                                                                                                                                                       |
+| InputDefault          | [PlanModifiers](/framework/migrating/resources/plan-modification#framework) field on attribute or implementation of [ResourceWithModifyPlan](/plugin/framework/migrating/resources/plan-modification#framework) interface                                                                                                                                                                                                                                              |
```
bflad, line 27:
```suggestion
| InputDefault          | N/A - no longer valid |
```

website/docs/plugin/framework/migrating/attributes-blocks/fields.mdx
```diff
@@ -9,6 +9,38 @@ description: >-
...
+| ComputedWhen          | [PlanModifiers](/framework/migrating/resources/plan-modification#framework) field on attribute or implementation of [ResourceWithModifyPlan](/plugin/framework/migrating/resources/plan-modification#framework) interface                                                                                                                                                                                                                                              |
```
bflad, line 33:
```suggestion
| ComputedWhen          | N/A - no longer valid |
```

website/docs/plugin/framework/migrating/attributes-blocks/fields.mdx
```diff
@@ -9,6 +9,38 @@ description: >-
...
+| StateFunc             | Requires implementation of bespoke logic before storing state, for instance in resource [Create method](plugin/framework/migrating/resources/crud#framework-1)                                                                                                                                                                                                                                                                                                         |
+| Elem                  | `ElementType` on [ListAttribute](/plugin/framework/migrating/attributes-blocks/types), [MapAttribute](/plugin/framework/migrating/attributes-blocks/types) or [SetAttribute](/plugin/framework/migrating/attributes-blocks/types)                                                                                                                                                                                                                                      |
```
bflad, line 29: I think it might be good to call out that any `schema.Resource` in `Elem` should refer to the blocks documentation instead. 👍

website/docs/plugin/framework/migrating/attributes-blocks/blocks.mdx
```diff
@@ -14,6 +14,17 @@ This page explains how to migrate nested blocks that are not computed (i.e., do
 [Blocks with Computed Fields](/plugin/framework/migrating/attributes-blocks/blocks-computed) for more details
 about migrating nested blocks that contain fields that are computed.
 
+Refer to [Attribute Fields](/plugin/framework/migrating/attributes-blocks/fields) for details regarding the relationship
```
bflad, line 17: Since there are a variety of differences between blocks and attributes, I think this page might warrant its own table. The prior SDK allowed developers to do unsupported things, like declare `Sensitive` on blocks (https://github.com/hashicorp/terraform-plugin-sdk/pull/819) and we already have special callouts in paragraph form about `Computed` and `Required`. Therefore I think we should walk through the various fields and how they relate to blocks:

| SDK Schema Field | Framework |
|---|---|
| Computed | Link to computed block page |
| Optional | N/A - no implementation required |
| Required | `listvalidator.IsRequired()` and `setvalidator.IsRequired()` (let's omit any discussion on single nested blocks, they weren't possible to declare before) |
| Sensitive | N/A - only supported on attributes |

etc. 😄

bflad (commented):
I think this is shaping up nicely so far!

website/docs/plugin/framework/migrating/attributes-blocks/fields.mdx
bendbennett: Updated to include call out.

website/docs/plugin/framework/migrating/attributes-blocks/blocks.mdx
bendbennett: Have added a table into the Blocks page.

website/docs/plugin/framework/migrating/attributes-blocks/blocks.mdx
```diff
@@ -14,6 +14,38 @@ This page explains how to migrate nested blocks that are not computed (i.e., do
...
+| Default               | [PlanModifiers](/framework/migrating/resources/plan-modification#framework) field on attribute within block or implementation of [ResourceWithModifyPlan](/plugin/framework/migrating/resources/plan-modification#framework) interface                                                                                                        |
+| DefaultFunc           | [PlanModifiers](/framework/migrating/resources/plan-modification#framework) field on attribute within block or implementation of [ResourceWithModifyPlan](/plugin/framework/migrating/resources/plan-modification#framework) interface                                                                                                        |
+| Description           | `Description` field on attribute                                                                                                                                                                                                                                                                                                              |
```
bflad, line 31:
```suggestion
| Description           | `Description` field on block                                                                                                                                                                                                                                                                                                                  |
```

website/docs/plugin/framework/migrating/attributes-blocks/blocks.mdx
```diff
@@ -14,6 +14,38 @@ This page explains how to migrate nested blocks that are not computed (i.e., do
...
+| InputDefault          | N/A - no longer valid                                                                                                                                                                                                                                                                                                                         |
+| StateFunc             | Requires implementation of bespoke logic before storing state, for instance in resource [Create method](plugin/framework/migrating/resources/crud#framework-1)                                                                                                                                                                                |
+| Elem                  | N/A - define `Type` as [ListNestedBlock](/plugin/framework/migrating/attributes-blocks/blocks) or [SetNestedBlock](/plugin/framework/migrating/attributes-blocks/blocks)                                                                                                                                                                      |
```
bflad, line 34: I wonder if it might be good to phrase it this way?

```suggestion
| Elem                  | `NestedObject` within block                                                                                                                                                                      |
```

bflad (approved):
This looks really good to me 🚀 Nice work mapping everything out.

PR #666, opened 2023-02-09
Switch to Changie for changelog automation

SBGoods:


bflad (approved):
Looks good to me 🚀 Merging so we can start using it immediately. 😄

PR #667, opened 2023-02-09
Replace custom reflect.Struct logic with logic from encoding/json

### LINKED ISSUE #242: "Follow" Embedded Structs
### terraform-plugin-go version

```
v0.4.0
```

### Use cases

Consider the following type 👇🏼 
```
type Port struct {
	FromPort int `json:"fromPort" tfsdk:"from_port"`
	ToPort   int `json:"toPort" tfsdk:"to_port"`
}

type Source struct {
	Address string `json:"address" tfsdk:"address"`
	MAC     string `json:"mac" tfsdk:"mac"`
	*Port
}
```
In `JSON`, `address, mac, fromPort, and toPort` would all be at the _same level_. I want this identical behavior for `tdsdk`. Currently, it'll complain that the embedded `*Port` field doesn't contain a tag.

### Attempted solutions

I have a workaround worth about 100 lines of code that implements `FromTerraform5Value` and `ToTerraform5Value`.

### Proposal

Implement the same behavior as `encoding/json` so that...
```
"destination": {
		"from_port": {
			Type:     types.NumberType,
			Required: true,
		},
		"to_port": {
			Type:     types.NumberType,
			Required: true,
		},
		"address": {
			Type:     types.StringType,
			Optional: true,
		},
	}),
	Required: true,
},
```
...will _natively_ be _unmarshaled_ into the aforementioned type.

skirsten:
This PR replaces the existing custom struct reflection logic with the logic from encoding/json.

The code is taken from https://cs.opensource.google/go/go/+/refs/tags/go1.20:src/encoding/json/encode.go;l=1212. It was simplified a bit to remove unneeded code and fields. The core logic is still untouched.

Pros:
- Standard handling according to Go visibility rules
- Supports embedded structs
- Support untagged fields
- "If it works in encoding/json it should work here" :tm: 

Cons:
- Silent errors (just like encoding/json)
  For example, if you supply the same tag twice, the new logic will drop it completely. Luckily, this will be caught by the logic in the upper layers ("Object defines fields not found in struct: ...") so this might not be such a big deal?
- Can panic (just like encoding/json) if supplied wrong types etc.

There are definitely some trade-offs here. I just needed it to support embedded structs (closes #242) but found that implementing that would be far more difficult than just replacing the logic with the existing logic from the stdlib.

Let me know what you think!

### DIFF
--- a/internal/reflect/helpers.go
+++ b/internal/reflect/helpers.go
@@ -1,13 +1,10 @@
 package reflect
 
 import (
-	"context"
-	"fmt"
 	"reflect"
 	"regexp"
+	"sort"
 	"strings"
-
-	"github.com/hashicorp/terraform-plugin-framework/path"
 )
 
 // trueReflectValue returns the reflect.Value for `in` after derefencing all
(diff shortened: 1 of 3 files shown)


bflad:
Hi @skirsten 👋 Thank you very much for this contribution.

To be upfront with expectations: the maintainer team is unfortunately not actively working in this complex area of the codebase in the near future, so I'm not sure when we might get around to properly reloading context in this area's design or reviewing this contribution. Among trying to continually improve the documentation to get the ecosystem migrated away from terraform-plugin-sdk, we already have work queued up for the next two minor version releases (with new major features) that will likely lead into the April or later timeframe.

That said, I can say that any change which introduces more chances of runtime panics will likely not be accepted. I'm not saying that this does because I'm not actively reviewing the code or design here, but wanted to ensure that was mentioned out loud since it may cause the need for approaching potential solutions in a completely different way.

Thanks again.

drakkan:
Hello,

I'm writing a Terraform provider for SFTPGo and embedded structs support will help a lot for example user and group models share a lot of fields and having support for emebedded structs will allow me to easily avoid code duplication (it can also be donw without supporting embedded structs but it will be a bit more complex).

Besides this PR do you plan to support embedded structs?

bflad:
> Besides this PR do you plan to support embedded structs?

This PR is community-contributed and may not be reviewed or merged for the reasons outlined in https://github.com/hashicorp/terraform-plugin-framework/pull/667#issuecomment-1426136439. The team that maintains this Go module is not actively designing or implementing changes in the reflection code for at least the next few weeks. I would say we are hopeful that something can be supported for this type of functionality, but cannot make any guarantees at this time while we are focused on other work at the moment. The best place to track discussions and effort updates would likely be the associated issue: https://github.com/hashicorp/terraform-plugin-framework/issues/242.

PR #670, opened 2023-02-10
docs: Various documentation fixes while going through migration guide

austinvalle:
- No preexisting issues, just some old framework code snippets I found while migrating `cloudinit`


bendbennett (approved):
LGTM 👍

docs/design/plan-modification.md
```diff
@@ -215,9 +215,9 @@ Provider code:
 ```go
 type fileResourceType struct{}
 
-// GetSchema returns the schema for this resource.
-func (f fileResourceType) GetSchema(_ context.Context) (schema.Schema, []*tfprotov6.Diagnostic) {
```
bflad, line 219: FYI, the design documentation which was written early in the framework's implementation phases was intended to be a snapshot of the decision making around that time, rather than something intended to be perpetually updated since that changes the framing of the decision making.

If these documentation artifacts are more confusing for existing, we should consider removing them.

austinvalle: Ah, good catch, I did not see that. Reverting in follow-up #671

PR #672, opened 2023-02-14
website: adds content-check command

BRKalow:
Adds the `content-check` command to run content checks locally.


bflad (approved):
Verified locally with `cd website && npm ci && npm run content-check`

PR #680, opened 2023-03-02
Update release workflow to work with new branch protections

austinvalle:
- Contributes to: https://github.com/hashicorp/terraform-providers-devex-internal/issues/136


bflad (approved):
Looks good to me 🚀

PR #684, opened 2023-03-13
Add deprecation documentation (from Best Practices - SDKv2)

austinvalle:
contributes to https://github.com/hashicorp/terraform-plugin-sdk/issues/1151

**Related PRs:**
- https://github.com/hashicorp/terraform-website/pull/2474
- https://github.com/hashicorp/terraform-docs-common/pull/295
- https://github.com/hashicorp/terraform-plugin-sdk/pull/1170

## Notes
- There are some notes about drift already here: https://developer.hashicorp.com/terraform/plugin/framework/resources/read#recommendations and the original docs seemed very specific to SDK, so I did not migrate that page over -> https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/detecting-drift
- The deprecation page was transcribed from the original SDKv2 documentation and was my best effort on porting to framework. Looking for comments on if these examples I wrote are correct (I tested them in a sandbox provider on my local)


austinvalle:
Should we stack this new page under a section? I initially removed the section because there is only one page right now:
image

website/docs/plugin/framework/deprecations.mdx
```diff
@@ -0,0 +1,1282 @@
...
+1. Add conditional logic in the `Create` and `Update` functions of the data source or resource to handle both attributes. Generally, this involves using [`{type}.IsNull()`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/types/basetypes).
+1. Follow the rest of the procedures in the [Provider Attribute Removal section](#provider-attribute-removal). When the old attribute is removed, update the schema definition and documentation of the new attribute back to `Required`, and remove the `{type}validator.ExactlyOneOf` validator.
+
+#### Example Renaming of a Required Attribute
+
+Given this sample resource:
+
+```go
+package provider
```
austinvalle, line 79: Not sure if I correctly transcribed the original SDK deprecation code examples over, this is a first attempt 😄

website/docs/plugin/framework/deprecations.mdx
```diff
@@ -0,0 +1,1282 @@
...
+		// ... other configuration ...
+
+		Attributes: map[string]schema.Attribute{
+			// ... other attributes ...
+		},
+		DeprecationMessage: "use example_thing resource instead",
+	}
+}
+
+func (e *exampleWidgetResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
```
austinvalle, line 1043: Same as above, I transcribed the original SDK deprecation code examples over, but is there a better place to return this `diag` when deprecating an entire resource?

Original example: https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/deprecations#provider-data-source-or-resource-removal

website/docs/plugin/framework/deprecations.mdx
```diff
@@ -0,0 +1,1282 @@
...
+The recommended process is as follows:
+
+1. Duplicate the code of the existing resource, renaming (and potentially modifying) functions as necessary.
+1. Duplicate the documentation of the existing resource, renaming (and potentially modifying) as necessary.
+1. Add a [`DeprecationMessage`](/terraform/plugin/framework/handling-data/schemas#deprecationmessage) to the schema definition of the existing (now the "old") resource, noting to use the new resource in the message.
+1. Add `~> This resource is deprecated and will be removed in the next major version` to the documentation of the existing (now the "old") resource, noting to use the new resource.
+1. Add the new resource to the provider [`Resources`](/terraform/plugin/framework/providers#resources) function
+1. Follow the rest of the procedures in the [Provider Data Source or Resource Removal section](#provider-data-source-or-resource-removal).
+
+### Example Resource Renaming
```
austinvalle, line 1093: Same as above, I transcribed the original SDK deprecation code examples over, but is there a better place to return this `diag` when deprecating an entire resource?

Original example: https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/deprecations#provider-data-source-or-resource-rename

website/docs/plugin/framework/deprecations.mdx
```diff
@@ -0,0 +1,1282 @@
...
+1. Duplicate the documentation of the existing attribute, renaming one of them with the new attribute name.
+1. Add a [`DeprecationMessage`](/terraform/plugin/framework/handling-data/schemas#deprecationmessage-1) to the schema definition of the existing (now the "old") attribute, noting to use the new attribute in the message.
+1. Add `**Deprecated**` to the documentation of the existing (now the "old") attribute, noting to use the new attribute.
+1. Add a note to the documentation that either the existing (now the "old") attribute or new attribute must be configured.
+1. Add the type-specific [validator](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework-validators) `{type}validator.ExactlyOneOf` to the schema definition of the new attribute, with a path expression matching the old attribute. This will ensure at least one of the attributes is configured, but present an error to the operator if both are configured at the same time. For example, an attribute of type string would use the [`stringvalidator.ExactlyOneOf`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator#ExactlyOneOf) validator.
```
austinvalle, line 70: I initially had this as the `ConflictsWith` validator, not sure which one is recommended

website/docs/plugin/framework/deprecations.mdx
```diff
@@ -0,0 +1,1282 @@
...
+func (e *exampleWidgetResource) Schema(ctx context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		// ... other configuration ...
+
+		Attributes: map[string]schema.Attribute{
+			// ... other attributes ...
+
+			"existing_attribute": schema.StringAttribute{
+				Computed:           true,
+				DeprecationMessage: "use new_attribute instead",
```
bendbennett, line 777: Is it worth pointing out that `DeprecationMessage` on computed-only attributes are not surfaced through the CLI?

austinvalle: I can add some explanation and maybe a link to the schema/computed docs: https://developer.hashicorp.com/terraform/plugin/framework/handling-data/schemas#deprecationmessage-1

Are `DeprecationMessage`s for `Computed` fields **never** surfaced to the CLI as a warning? This doc above combines the messaging with a note about earlier terraform versions and `Required`/`Optional`, but they seem to be separate notes.

bendbennett: > Are DeprecationMessages for Computed fields never surfaced to the CLI as a warning?

I believe that's correct but perhaps @bflad would like to comment?

bflad: The [`DeprecationMessage` field Go documentation](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework@v1.1.1/resource/schema#BoolAttribute.DeprecationMessage) tries to go into more detail here:

```
	// In Terraform 1.2.7 and later, this warning diagnostic is displayed any
	// time a practitioner attempts to configure a value for this attribute and
	// certain scenarios where this attribute is referenced.
	//
	// In Terraform 1.2.6 and earlier, this warning diagnostic is only
	// displayed when the Attribute is Required or Optional, and if the
	// practitioner configuration sets the value to a known or unknown value
	// (which may eventually be null). It has no effect when the Attribute is
	// Computed-only (read-only; not Required or Optional).
	//
	// Across any Terraform version, there are no warnings raised for
	// practitioner configuration values set directly to null, as there is no
	// way for the framework to differentiate between an unset and null
	// configuration due to how Terraform sends configuration information
	// across the protocol.
	//
	// Additional information about deprecation enhancements for read-only
	// attributes can be found in:
	//
	//  - https://github.com/hashicorp/terraform/issues/7569
	//
```

austinvalle: Thanks for the context! I added a blurb in the beginning of the documentation noting to reference the schema docs for more info on when warnings will be raised: [16f88fe](https://github.com/hashicorp/terraform-plugin-framework/pull/684/commits/16f88fe2b16e54aab3d154397b5bf3c2d464debe)

website/docs/plugin/framework/deprecations.mdx
bendbennett: I believe `ConflictsWith` or `ExactlyOneOf` can be used with equivalent results but I think I would lean towards the latter as seems from a semantic point of view to be more aligned with the intent.

PR #687, opened 2023-03-14
website: Enhance diagnostics page with references to diag.New functions and more examples

### LINKED ISSUE #675: Enhancements for Custom Diagnostic Types Documentation
### Module version

```
v1.1.1
```

### Use-cases

In a forum I cannot publicly link to, there was some confusion about the https://developer.hashicorp.com/terraform/plugin/framework/diagnostics#custom-diagnostics-types documentation. In particular, it is not clear that (at least currently) these custom diagnostics must align to the Terraform-understood warning/error severity. Providing a clear use case example would also be helpful there.

### Proposal

- Add a sentence talking about the severity constraint (e.g. we don't particularly enable making "hidden from Terraform" diagnostics)
- Add real use case examples (e.g. a error diagnostic type which also stores the underlying error)

```go
// UnderlyingErrorDiagnostic is an error diagnostic
// which also stores the underlying error.
type UnderlyingErrorDiagnostic struct {
  Detail          string
  Summary         string
  UnderlyingError error
}

func (d UnderlyingErrorDiagnostic) Detail() string {
  return d.Detail
}

func (d UnderlyingErrorDiagnostic) Equal(o SpecialDiagnostic) bool {
  if d.Detail() != o.Detail() {
    return false
  }

  if d.Summary() != o.Summary() {
    return false
  }

  if d.UnderlyingError == nil {
    return o.UnderlyingError == nil
  }

  if o.UnderlyingError == nil {
    return false
  }

  if d.UnderlyingError.Error() != o.UnderlyingError.Error() {
    return false
  }

  return true
}

func (d UnderlyingErrorDiagnostic) Severity() diag.Severity {
  return diag.SeverityError
}

func (d UnderlyingErrorDiagnostic) Summary() string {
  return d.Summary
}
```

Another example option may be one that mainly hardcodes summary/detail information except for a particular last bit of information:

```go
type APIErrorDiagnostic struct {
  Detail string
}

func (d APIErrorDiagnostic) Detail() string {
  return "An unexpected error occurred while calling the API: " + d.Detail
}

func (d APIErrorDiagnostic) Equal(o SpecialDiagnostic) bool {
  if d.Detail() != o.Detail() {
    return false
  }

  return true
}

func (d APIErrorDiagnostic) Severity() diag.Severity {
  return diag.SeverityError
}

func (d APIErrorDiagnostic) Summary() string {
  return "Unexpected API Error"
}
```

bflad:
Closes #675

### DIFF
--- a/website/docs/plugin/framework/diagnostics.mdx
+++ b/website/docs/plugin/framework/diagnostics.mdx
@@ -247,7 +247,7 @@ func (s exampleType) Validate(ctx context.Context, in tftypes.Value, path path.P
 
 ### Consistent Diagnostic Creation
 
-Create a helper function in your provider code to generate consistent diagnostics for types of errors/warnings. It is also possible to use [custom diagnostics types](#custom-diagnostics-types) to accomplish this same goal.
+Create a helper function in your provider code using the diagnostic creation functions available in the [`diag` package](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/diag) to generate consistent diagnostics for types of errors/warnings. It is also possible to use [custom diagnostics types](#custom-diagnostics-types) to accomplish this same goal.
 
 The [`diag` package](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/diag) provides these functions to create various diagnostics:
 
--- a/website/docs/plugin/framework/diagnostics.mdx
+++ b/website/docs/plugin/framework/diagnostics.mdx
@@ -48,7 +48,7 @@ working with diagnostics, using functionality from the available
 
 ### Severity
 
-`Severity` specifies whether the diagnostic is an error or a warning.
+`Severity` specifies whether the diagnostic is an error or a warning. Terraform, nor the framework, supports other severity levels. Use [logging](/terraform/plugin/log/writing) for debugging or informational purposes.
 
 - An **error** will be displayed to the practitioner and halt Terraform's
   execution, not continuing to apply changes to later resources in the graph.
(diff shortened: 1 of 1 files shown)


website/docs/plugin/framework/diagnostics.mdx
```diff
@@ -245,12 +245,55 @@ func (s exampleType) Validate(ctx context.Context, in tftypes.Value, path path.P
...
 ```
 
-#### Custom Diagnostics Types
+### Consistent Diagnostic Creation
 
-Advanced provider developers may need to further differentiate or store
-additional data in error and warning types. The
-[`diag.Diagnostic` type](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/diag#Diagnostic)
-is an interface type that can be implemented with these methods:
+Create a helper function in your provider code to generate consistent diagnostics for types of errors/warnings. It is also possible to use [custom diagnostics types](#custom-diagnostics-types) to accomplish this same goal.
```
austinvalle, line 250: > _Create a helper function in your provider code to generate consistent diagnostics for types of errors/warnings._

This first line reads like a recommendation to create your own helper function, but the following content describes the existing helpers we provide. Could restructure like: "you can utilize X helper functions we provide (list functions), or you can write your own custom helper function like [this example](https://github.com/hashicorp/terraform-plugin-framework/pull/687/files#diff-288f191e52a4d5293f97d279c6253ea1770f5b0ae793e6ad1c9f6b5d49d9fedeR261-R272)"

austinvalle (approved):
Just one note, looks great 👍🏻

website/docs/plugin/framework/diagnostics.mdx
bflad: Happy to adjust this -- just not sure what the best verbiage might be. In the code example below, `APIErrorDiagnostic()` is what I was intending to convey as the "create a helper function" part here. Is this suggested change maybe better here?

```suggestion
Create a helper function in your provider code using the diagnostic creation functions available in the [`diag` package](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/diag) to generate consistent diagnostics for types of errors/warnings. It is also possible to use [custom diagnostics types](#custom-diagnostics-types) to accomplish this same goal.
```

austinvalle: Ah, I understand the intention now, that change clears it up for me! 👍🏻

bendbennett (approved):
LGTM 🎸

PR #688, opened 2023-03-14
attr: Fix verbiage for Type and Value type Equal methods

### LINKED ISSUE #679: attr.Value.Equal() : *semantically* equal?
### Module version

```
github.com/hashicorp/terraform-plugin-framework v1.1.2-0.20230214195151-b784296a0acb
```

While writing tests for [PR 110 for terraform-framework-validators](https://github.com/hashicorp/terraform-plugin-framework-validators/pull/110) I encountered a situation where the `Equal()` method on implementations of `attr.Value` return false when asked to make a comparison with another `attr.Value` with a different underlying type.

Specifically, `z` winds up `false` here:

```go
	x := types.Int64Value(42)
	y := types.NumberValue(big.NewFloat(float64(42)))
	z := x.Equal(y)
``` 

Documentation for `attr.Value.Equal()` uses the term *semantically equal*, which suggests to me that `z` should wind up `true` in the snippet above.

```go
type Value interface {
	// Equal must return true if the Value is considered semantically equal
	// to the Value passed as an argument.
	Equal(Value) bool
```

Maybe I've misunderstood?

I was going to raise this question in the conversation about the validator PR, but thought it better to address it separately.

At this point, I'm not sure whether we're talking about (a) confusion on my part (b) implementation bugs in `Equal()` or (c) a documentation bug in the `Value` interface.

In any event, clarifying this (apparent?) discrepancy should drive the answer to the test failure I mentioned in the conversation about the validator PR.

Thanks!

bflad:
Closes #679

This documentation may have stemmed from an early design goal to include semantic equality within types. The initial implementation did not include that functionality, which is being tracked with https://github.com/hashicorp/terraform-plugin-framework/issues/70.

### DIFF
--- a/attr/type.go
+++ b/attr/type.go
@@ -31,8 +31,15 @@ type Type interface {
 	// and improving error diagnostics.
 	ValueType(context.Context) Value
 
-	// Equal must return true if the Type is considered semantically equal
-	// to the Type passed as an argument.
+	// Equal should return true if the Type is considered equivalent to the
+	// Type passed as an argument.
+	//
+	// Most types should verify the associated Type is exactly equal to prevent
+	// potential data consistency issues. For example:
+	//
+	//  - basetypes.Number is inequal to basetypes.Int64 or basetypes.Float64
+	//  - basetypes.String is inequal to a custom Go type that embeds it
+	//
 	Equal(Type) bool
 
 	// String should return a human-friendly version of the Type.
--- a/attr/value.go
+++ b/attr/value.go
@@ -27,8 +27,22 @@ type Value interface {
 	// a tftypes.Value.
 	ToTerraformValue(context.Context) (tftypes.Value, error)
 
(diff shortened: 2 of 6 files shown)


chrismarget-j:
Thanks for this @bflad.

The same "semantically equal" language appears in the various implementation of these methods (that's where I first encountered it). Seems like it should change there too? You probably already have that in mind, but I wanted to mention in case you'd overlooked it.

bflad:
@chrismarget-j ah ha, okay. While the "primitive" types seem to say the right thing, I wasn't aware that some of the collection types `Equal()` also used the incorrect verbiage. Will fix those now.

bflad:
Please reach out if https://github.com/hashicorp/terraform-plugin-framework/pull/688/commits/ddcac2ba1fd70a0220a1b03d653098b7b893c8cb is not what was expected.

austinvalle (approved):
👍🏻

PR #689, opened 2023-03-14
types: Go pointer value constructors and getters for Bool, Float64, Int64, and String

### LINKED ISSUE #649: Consider Go Pointer Value Constructors and Getters for Primitive Types
### Module version

```
v1.1.1
```

### Use-cases

Provider developers with terraform-plugin-framework are typically encouraged to work with `types` package types for a consistent data model across Terraform configuration, plan, and state data. Configuration and plan data may contain unknown values, which an important distinction from other type systems, including Go's built-in types.

However, API-based SDKs opposite of Terraform may prefer pointer-based Go types for data handling. For example:

- [AWS SDK for Go](https://aws.amazon.com/sdk-for-go/) (e.g. [`ec2.CreateVpcInput`](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ec2#CreateVpcInput) / [`ec2.CreateVpcOutput`](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ec2#CreateVpcOutput))
- [HashiCorp Go Azure SDK](https://github.com/hashicorp/go-azure-sdk) (e.g. [`servers.Server`](https://pkg.go.dev/github.com/hashicorp/go-azure-sdk@v0.20230202.1112019/resource-manager/mysql/2021-05-01/servers#Server))
- [go-github](https://github.com/google/go-github) (e.g. [`github.Repository`](https://pkg.go.dev/github.com/google/go-github/v50/github#Repository))

Given the current `types` package design which has methodologies for working directly with non-pointer Go types, it can therefore require extra pointer-based logic to correctly work with data between Terraform and the API SDK.

### Attempted Solutions

These examples assume `*bool` API client value handling and `schema.BoolAttribute` / `types.Bool` on the Terraform side.

```go
if client.Value == nil {
  data.Value = types.BoolNull()
}

data.Value = types.BoolValue(*client.Value)
```

```go
if data.Value.IsNull() {
  client.Value = nil
} else {
  value := data.Value.ValueBool()
  client.Value = &value
}
```

### Proposal

This proposal covers the following types in the framework type system:

| Framework Type (FT) | Go Type (GT) |
|---|---|
| `basetypes.Bool` | `bool` |
| `basetypes.Float64` | `float64` |
| `basetypes.Int64` | `int64` |
| `basetypes.String` | `string` |

In the `types/basetypes` package, create `New{FT}PointerValue(*GT)` value creator functions. For example:

```go
// NewBoolPointerValue creates a Bool with a null value if nil or a known value.
func NewBoolPointerValue(value *bool) BoolValue {
	if value == nil {
		return NewBoolNull()
	}

	return NewBoolValue(*value)
}
```

In the `types` package, wrapper functions would be created to use to their basetypes equivalent similar to the other value creator function aliases. For example:

```go
// BoolPointerValue creates a Bool with a null value if nil or a known value.
func BoolPointerValue(value *bool) basetypes.BoolValue {
	return basetypes.NewBoolPointerValue(value)
}
```

In the `types/basetypes` package, create `func (t {FT}) Value{GT}Pointer() *{GT}` getter methods. For example:

```go
// ValueBoolPointer returns a pointer to the known bool value, nil for a null bool value,
// or a pointer to false for an unknown bool value.
func (b BoolValue) ValueBoolPointer() *bool {
	if b.IsNull() {
		return nil
	}

	return &b.value
}
```

Note that this method implementation is similar to the existing `func (t {FT}) Value{GT}() {GT}` getter methods, where the design is intentionally skewed to provide a potentially invalid answer in the name of simplicity rather than introduce a second `diag.Diagnostics` return which must always be handled (or worse and forbidden in our typical design decisions, a panic). The non-pointer getter methods return the Go type's zero-value in the case of null and unknown. Provider developers are expected to check against a value's unknown state prior to these methods and Terraform will typically raise its own error should there be a data consistency issue.

These new functions and methods should enable provider code such as:

```go
data.Value = types.BoolPointerValue(client.Value)
// and conversely
client.Value = data.Value.ValueBoolPointer()
```

bflad:
Closes #649

### DIFF
--- a/types/basetypes/int64.go
+++ b/types/basetypes/int64.go
@@ -48,6 +48,16 @@ func NewInt64Value(value int64) Int64Value {
 	}
 }
 
+// NewInt64PointerValue creates a Int64 with a null value if nil or a known
+// value. Access the value via the Int64 type ValueInt64Pointer method.
+func NewInt64PointerValue(value *int64) Int64Value {
+	if value == nil {
+		return NewInt64Null()
+	}
+
+	return NewInt64Value(*value)
+}
+
 // Int64Value represents a 64-bit integer value, exposed as an int64.
 type Int64Value struct {
 	// state represents whether the value is null, unknown, or known. The
@@ -129,6 +139,16 @@ func (i Int64Value) String() string {
 // 0.
 func (i Int64Value) ValueInt64() int64 {
 	return i.value
+}
+
+// ValueInt64Pointer returns a pointer to the known int64 value, nil for a
+// null value, or a pointer to 0.0 for an unknown value.
+func (i Int64Value) ValueInt64Pointer() *int64 {
+	if i.IsNull() {
+		return nil
+	}
+
+	return &i.value
 }
 
 // ToInt64Value returns Int64.
--- a/types/basetypes/int64.go
+++ b/types/basetypes/int64.go
@@ -142,7 +142,7 @@ func (i Int64Value) ValueInt64() int64 {
 }
 
 // ValueInt64Pointer returns a pointer to the known int64 value, nil for a
-// null value, or a pointer to 0.0 for an unknown value.
+// null value, or a pointer to 0 for an unknown value.
 func (i Int64Value) ValueInt64Pointer() *int64 {
 	if i.IsNull() {
 		return nil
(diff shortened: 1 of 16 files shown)


types/basetypes/int64.go
```diff
@@ -131,6 +141,16 @@ func (i Int64Value) ValueInt64() int64 {
 	return i.value
 }
 
+// ValueInt64Pointer returns a pointer to the known int64 value, nil for a
+// null value, or a pointer to 0.0 for an unknown value.
```
austinvalle, line 145:
```suggestion
// null value, or a pointer to 0 for an unknown value.
```

austinvalle (approved):
One note, otherwise LGTM 🚀

bendbennett (approved):
LGTM 🎸

PR #690, opened 2023-03-15
types/basetypes: Lift and shift type versus value code

bflad:
This change cleans up a small piece of technical debt by aligning all the code associated with the base types to follow the convention `{TYPE}_type.go`, `{TYPE}_type_test.go`, `{TYPE}_value.go`, and `{TYPE}_value_test.go`. Previously, only some of the types followed this convention. No logic changes were made.


bendbennett (approved):
LGTM 🎸

PR #693, opened 2023-03-17
Use generics for states and plans

miton18:
I was asking me if this project is ready to accept generics.

This does not change any working behaviors except the syntax.

From:
```go
var plan MyResourcePlan

	return
}
```

to:
```go
plan, diags := req.Plan.Get[MyResourcePlan](ctx)

	return
}
```

### DIFF
--- a/tfsdk/plan.go
+++ b/tfsdk/plan.go
@@ -11,14 +11,15 @@ import (
 )
 
 // Plan represents a Terraform plan.
-type Plan struct {
+type Plan[T any] struct {
 	Raw    tftypes.Value
 	Schema fwschema.Schema
 }
 
 // Get populates the struct passed as `target` with the entire plan.
-func (p Plan) Get(ctx context.Context, target interface{}) diag.Diagnostics {
-	return p.data().Get(ctx, target)
+func (p Plan[T]) Get(ctx context.Context) (T, diag.Diagnostics) {
+	var target T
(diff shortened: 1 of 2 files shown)


bflad:
Hi @miton18 👋 Thank you for submitting this. Please note that per our [contributing guide](https://github.com/hashicorp/terraform-plugin-framework/blob/main/.github/CONTRIBUTING.md#new-pull-request), it is highly recommended to create issues before raising pull requests with this project, so we can go through a design phase and discuss tradeoffs from both the maintainer and provider developer perspectives.

In this particular case, I think we would need to understand the true benefits of requiring a syntax change for provider developers. There may be a few valid reasons, but also some invalid reasons to switch usage to generics. These are details that would typically come through a feature request or bug report, such as explaining how certain code cannot be written given the current types/methods. From a code perspective, we also cannot break code compatibility until a new major version, so therefore if something were to be introduced today, it would need to be done with separate types/methods. I'm not sure if this is something where we would want to enable two conflicting code paths for provider developers.

Since this is not a change we would consider as-is and we have no set plans for a major version update anytime soon, I'm going to close this since it feels pretty in-actionable. However, I would encourage creating a new issue with a full explanation of why this change is necessary or beneficial so we can discuss its merits and work through a potential path forward before working on any potential code implementation. Thanks again.

PR #697, opened 2023-03-21
website: Fix table on Default page

bflad:
Reference: https://developer.hashicorp.com/terraform/plugin/framework/resources/default

It looks like the production Markdown renderer will cause the page to not be loaded when a table is incorrectly defined, rather than fallback to an errant display of that table as it does locally.


austinvalle (approved):
🚀

Issue #705, opened 2023-03-28
Enhance Attribute Name Validation to Handle Leading Numerics


bflad:
### Module version

```
v1.2.0
```

### Use-cases

The framework's attribute name validation is currently the regular expression `^[a-z0-9_]+$`. The Terraform configuration language forbids identifiers (how attribute names are handled) from beginning with numeric characters.

### Proposal

Adjust the attribute name validation regular expression to `^[a-z_][a-z0-9_]*$`. Technically the validation should also permit hyphens (`-`), however the provider ecosystem has conventionally never used hyphens in attribute names, so the framework should remain opinionated here to prevent introducing practitioner confusion.

### References

- https://developer.hashicorp.com/terraform/language/syntax/configuration#identifiers

Issue #704, opened 2023-03-28
Add for_each to Reserved Root Attribute Name Validation


bflad:
### Module version

```
v1.2.0
```

### Use-cases

The schema validation for reserved root attribute names includes the following:

- `connection`
- `count`
- `depends_on`
- `lifecycle`
- `provider`
- `provisioner`

However, it is missing `for_each`.

### Proposal

Add `for_each` to the reserved root attribute name validation.

### References

- https://developer.hashicorp.com/terraform/language/meta-arguments/for_each

Issue #574, opened 2022-12-12
Handle Nested Attributes and Nested Blocks Using List, Set and Map When Validating Attribute Field Names


bendbennett:
### Module version
```
v0.17
```

### Use-cases

```go
for k, v := range attributes {
	d := validateAttributeFieldName(path.AtName(k), k, v)
```

```go
for k, v := range blocks {
	d := validateBlockFieldName(path.AtName(k), k, v)
```

Since the nested attributes might be done using list, map, or set mode, we'll also need to account for adding those intermediate steps (AtListIndex, AtMapKey, AtSetValue) to the path. 

### Attempted Solutions

N/A

### Proposal

Fix up the path information and consider including it in the diagnostic details in case it is not in Terraform's output. Maybe we can use something like the zero-value (e.g. AtListIndex(0), AtMapKey(""), AtSetValue(/* ? */)) for describing those intermediate path steps for static validation.

### References

- #136 
- #548

PR #706, opened 2023-03-29
all: Refactor schema validation to support type-specific logic and prevent AttributeTypes/ElementType field panics

### LINKED ISSUE #699: schema.MapAttribute struct without ElementType member causes segmentation fault
### Module version
```
1.1.1
```

```go
schema.Schema{
    // generic example following
    "foo": schema.MapAttribute{
      Description: "A Map attribute",
      Required:    true,
    },
  },
}
```

Adding e.g. `ElementType: types.StringType,` to the schema above fixes the issue as expected.

### Expected Behavior
A clean runtime error thrown that the struct is missing the `ElementsAs` member.

### Actual Behavior
```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x6483df]

goroutine 83 [running]:
github.com/hashicorp/terraform-plugin-framework/types/basetypes.MapType.TerraformType(...)
	/home/matt/go/pkg/mod/github.com/hashicorp/terraform-plugin-framework@v1.1.1/types/basetypes/map.go:66
github.com/hashicorp/terraform-plugin-framework/internal/toproto6.SchemaAttribute({0xf950c0, 0xc000111080}, {0xe1d5c1, 0x6}, 0xc000112120, {0xf99408?, 0xc000262000?})
	/home/matt/go/pkg/mod/github.com/hashicorp/terraform-plugin-framework@v1.1.1/internal/toproto6/schema_attribute.go:26 +0x158
github.com/hashicorp/terraform-plugin-framework/internal/toproto6.Schema({0xf950c0, 0xc000111080}, {0xf999c0, 0xc00011e3c0})
	/home/matt/go/pkg/mod/github.com/hashicorp/terraform-plugin-framework@v1.1.1/internal/toproto6/schema.go:26 +0x2e6
github.com/hashicorp/terraform-plugin-framework/internal/toproto6.GetProviderSchemaResponse({0xf950c0, 0xc000111080}, 0xc00018f610)
	/home/matt/go/pkg/mod/github.com/hashicorp/terraform-plugin-framework@v1.1.1/internal/toproto6/getproviderschema.go:51 +0x52e
github.com/hashicorp/terraform-plugin-framework/internal/proto6server.(*Server).GetProviderSchema(0xc000254580, {0xf950c0?, 0xc000110f90?}, 0x1546e10)
	/home/matt/go/pkg/mod/github.com/hashicorp/terraform-plugin-framework@v1.1.1/internal/proto6server/server_getproviderschema.go:23 +0xc7
github.com/hashicorp/terraform-plugin-go/tfprotov6/tf6server.(*server).GetProviderSchema(0xc00031d680, {0xf950c0?, 0xc0001104e0?}, 0x0?)
```

There are several more lines, but their stack trace pertains to the gRPC server and are omitted for brevity.

### Steps to Reproduce
Executing an acceptance test would be the easiest reproduction step.

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/574
Closes #699
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/704
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/705

The goal of this change is to enable easier schema validation within the framework logic. This is achieved by implementing new internal interfaces that implement shared logic across all schema types, then introduce methods on attribute types which currently implement type-specific validation logic.

The `Schema` type `Validate()` method deprecations are necessary to ensure that the framework logic can pass the available `context.Context`. While no framework logic is currently dependent on the `context.Context`, its possible that future logic may want to use `tfsdklog`.

The new `ValidateImplementation()` methods on attribute types, while technically exported, cannot be implemented externally due to their usage of internal types. This follows the existing implementation of the framework where attribute types already cannot be extended due to the `Equal(fwschema.Attribute)` method requirement. Therefore these new `ValidateImplementation()` methods do not need to worry about compatibility promises and can be modified or removed in the future.

This change also includes some breadcrumbs for other schema validation requests.

### DIFF
--- a/.changes/unreleased/BUG FIXES-20230329-102850.yaml
+++ b/.changes/unreleased/BUG FIXES-20230329-102850.yaml
@@ -1,0 +1,7 @@
+kind: BUG FIXES
+body: 'datasource/schema: Raise errors with `ListAttribute`, `MapAttribute`, `ObjectAttribute`,
+  and `SetAttribute` implementations instead of panics when missing required `AttributeTypes` or
+  `ElementTypes` fields'
+time: 2023-03-29T10:28:50.525346-04:00
+custom:
+  Issue: "699"
--- a/.changes/unreleased/BUG FIXES-20230329-102851.yaml
+++ b/.changes/unreleased/BUG FIXES-20230329-102851.yaml
@@ -1,0 +1,7 @@
+kind: BUG FIXES
+body: 'provider/metaschema: Raise errors with `ListAttribute`, `MapAttribute`, `ObjectAttribute`,
+  and `SetAttribute` implementations instead of panics when missing required `AttributeTypes` or
+  `ElementTypes` fields'
+time: 2023-03-29T10:28:51.525346-04:00
+custom:
+  Issue: "699"
--- a/.changes/unreleased/BUG FIXES-20230329-102852.yaml
+++ b/.changes/unreleased/BUG FIXES-20230329-102852.yaml
@@ -1,0 +1,7 @@
+kind: BUG FIXES
+body: 'provider/schema: Raise errors with `ListAttribute`, `MapAttribute`, `ObjectAttribute`,
+  and `SetAttribute` implementations instead of panics when missing required `AttributeTypes` or
+  `ElementTypes` fields'
+time: 2023-03-29T10:28:52.525346-04:00
+custom:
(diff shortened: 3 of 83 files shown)


bendbennett (approved):
LGTM 🎸

PR #707, opened 2023-03-30
internal/fwschema: Validate for_each as a reserved resource attribute name

### LINKED ISSUE #704: Add for_each to Reserved Root Attribute Name Validation

bflad:
Closes #704

Similar to the other reserved resource attribute name checks, in the unlikely event provider developers need a method to opt out of the safety check, a feature request can be created.


austinvalle (approved):
👍🏻

PR #708, opened 2023-03-30
internal/fwschema: Validate leading numerics in attribute names

### LINKED ISSUE #705: Enhance Attribute Name Validation to Handle Leading Numerics

bflad:
Closes #705

The error diagnostic details will only contain this additional context if a numeric is detected.


austinvalle (approved):
🚀

PR #710, opened 2023-03-30
Move debug logs to trace

trodge:
These messages are appearing in our logs during acceptance test runs and make up the majority of the log without providing much information. We would prefer if they were logged at a lower level.

### DIFF
--- a/internal/fwschemadata/data_set_at_path.go
+++ b/internal/fwschemadata/data_set_at_path.go
@@ -68,9 +68,9 @@ func (d *Data) SetAtPath(ctx context.Context, path path.Path, val interface{}) d
 
 	if attrTypeWithValidate, ok := attrType.(xattr.TypeWithValidate); ok {
 		logging.FrameworkTrace(ctx, "Type implements TypeWithValidate")
-		logging.FrameworkDebug(ctx, "Calling provider defined Type Validate")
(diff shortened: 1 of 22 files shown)


bflad:
Hi @trodge 👋 Thank you for submitting this. While the maintainers generally agree that some of the logging (e.g. schema-based logic handoffs to provider-defined code) is likely a good candidate for log level demotion now that the framework code has stabilized since v1.0.0, per our [contributing guidelines](https://github.com/hashicorp/terraform-plugin-framework/blob/main/.github/CONTRIBUTING.md#new-pull-request) it feels like this type of developer-affecting change should be advertised and discussed in an issue first. We'll likely need to publish (and potentially try to enforce) the updated guidelines for logging levels as part of this, since the current precedent is that all handoffs to provider-defined logic should be DEBUG level. Thank you again.

bflad:
Please note that in the meantime, you may be able to tune your logging by setting the environment variable `TF_LOG_SDK_FRAMEWORK=INFO` in addition to other `TF_LOG*` environment variables already set.

rileykarson:
Could that target be added to https://developer.hashicorp.com/terraform/internals/debugging? We'd been trying to limit the log noise and found that these logs were coming through based on the `TF_LOG_PROVIDER` settings.

PR #711, opened 2023-03-31
resource/schema: Prevent UseStateForUnknown data misalignment by raising implementation error when under a list or set

bflad:
Closes #709
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/717

This change will adjust the `UseStateForUnknown` plan modifiers to return an implementation error if defined on an attribute below a list or set. In any cases where the list or set elements were rearranged or removed, the plan modifier result could cause invalid data being proposed in the plan or Terraform data consistency errors. These situations require additional value identifying information from the provider developer to ensure that the prior state for an element is correctly realigned, however the `UseStateForUnknown` plan modifier takes no parameters for such information and adding those parameters would represent a breaking change to the function signature. Future framework proposals will investigate whether a second plan modifier can be introduced to cover this situation.

The new plan modification unit tests reproduced the issue prior to the new implementation error:

```
--- FAIL: TestAttributeModifyPlan (0.00s)
    --- FAIL: TestAttributeModifyPlan/attribute-set-nested-nested-usestateforunknown-elements-rearranged (0.00s)
        attribute_plan_modification_test.go:1627: Unexpected response (-wanted, +got):   fwserver.ModifyAttributePlanResponse{
            - 	AttributePlan:   s`[{"nested_computed":"statevalue2","nested_required":"testvalue2"},{"nested_computed":"statevalue1","nested_required":"testvalue1"}]`,
            + 	AttributePlan:   s`[{"nested_computed":"statevalue1","nested_required":"testvalue2"},{"nested_computed":"statevalue2","nested_required":"testvalue1"}]`,
              	Diagnostics:     nil,
              	RequiresReplace: s"[]",
              	Private:         nil,
              }
    --- FAIL: TestAttributeModifyPlan/attribute-set-nested-nested-usestateforunknown-elements-removed (0.00s)
        attribute_plan_modification_test.go:1627: Unexpected response (-wanted, +got):   fwserver.ModifyAttributePlanResponse{
            - 	AttributePlan:   s`[{"nested_computed":"statevalue2","nested_required":"testvalue2"}]`,
            + 	AttributePlan:   s`[{"nested_computed":"statevalue1","nested_required":"testvalue2"}]`,
              	Diagnostics:     nil,
              	RequiresReplace: s"[]",
              	Private:         nil,
              }
    --- FAIL: TestAttributeModifyPlan/attribute-list-nested-nested-usestateforunknown-elements-removed (0.00s)
[...]

PR #718, opened 2023-04-07
resource/schema: Implement MatchElementStateForUnknown plan modifier

### LINKED ISSUE #717: Consider MatchElementStateForUnknown() Plan Modifiers
### Module version

```
v1.2.0
```

### Use-cases

Over in #709, we have discovered that the `UseStateForUnknown()` plan modifier (amongst other custom ones) can return unexpected values when a parent list or set has elements rearranged or removed. In particular:

* Attribute-based plan modifiers which attempt to read prior state (such as `UseStateForUnknown()`)
* Implementing those attribute-based plan modifiers on attributes underneath a `schema.ListNestedAttribute`, `schema.ListNestedBlock`, `schema.SetNestedAttribute`, or `schema.SetNestedBlock`
* The resource is planning/applying an in-place update which rearranges and/or removes those elements within the list/set

This is because the framework's list and set data structures are array index based and the planning implementation will align configuration, plan, and prior state values based on that index. The implementation itself was an design choice where other choices introduced their own potential tradeoffs, such as requiring provider developers to implement much more explicit logic into their schemas. It was particularly desirable to avoid the prior terraform-plugin-sdk model of set and plan difference handling, which consistently had bug reports which were difficult or not possible to fix without breaking changes. In any event, the framework's data handling for attribute plan modifiers cannot potentially be changed until a theoretical next major version, so that is not an option in this case.

This leaves the framework maintainers in an awkward position where `UseStateForUnknown()` can introduce unexpected plans or Terraform data consistency errors, which framework-native functionality should never do. It is possible to detect when the plan modifier is being called while under a list or set, which means that it can raise its own implementation error to tell developers to remove the plan modifier. This does mean however, that previously working cases such as unchanged lists/sets or ones where only new list/set elements were added will lose their ability to display a more known plan.

### Attempted Solutions

Provider developers implementing custom attribute plan modifiers with fairly complex data handling logic.

#### Parent Plan Modifier

One possible proposal would be a new list/set attribute plan modifier that enables provider developers to input tracking attributes which the plan modifier logic would realign the prior state values of any unknown computed values. As a design sketch:

```go
schema.ListNestedAttribute{
      "attr1": schema.StringAttribute{
         Required: true,
       },
       "attr2": schema.StringAttribute{
         Computed: true,
       },
       // potentially other configurable or non-configurable attributes to handle
    },
  },
  PlanModifiers: []planmodifier.List{
    listplanmodifier.MatchElementStateForUnknown(
      path.MatchRelative().AtAnyListIndex().AtName("attr1"),
      // potentially other configurable attributes to handle
    ),
  },
},
```

While requiring less provider developer implementation details especially in the case of many computed attributes and potentially being a slightly easier framework implementation, this approach does have a few immediate downsides:

- Action at a distance: Provider developers need to know this implementation goes in a different location than the affected attribute.
- Lack of ability to choose which element attributes it effects: Unless separate function parameters are introduced or a separate plan modifier is created, there's no ability to have the value preservation apply to a subset of computed attribute values. That could introduce Terraform data consistency errors between plan and apply depending on the provider implementation details.

#### Nested Plan Modifiers

Another possible proposal would be new typed attribute plan modifiers, such as in the `resource/schema/stringplanmodifer` package, where the provider developer can optionally realign the prior state based on identifying attributes. As a design sketch:

```go
schema.ListNestedAttribute{
      "attr1": schema.StringAttribute{
         Required: true,
       },
       "attr2": schema.StringAttribute{
         Computed: true,
           stringplanmodifier.MatchElementStateForUnknown(
             path.MatchRelative().AtParent().AtName("attr1"),
             // potentially other configurable attributes to handle
           ),
         },
       },
       // potentially other configurable or non-configurable attributes to handle
    },
  },
  // ...
},
```

This means that each underlying attribute can opt into the behavior as appropriate (action where its implemented and explicit choice of effect). There is slightly less discoverability concerns since the implementation would be next to `UseStateForUnknown()`.

The plan modifier logic will need to implement some rigorous validation where it raises errors if:

- The plan modifier is not implemented on an attribute below a list/set.
- It receives 0 path expressions.
- A given path expression goes outside the parent list/set.
- A given path expression goes deeper than nesting of the current path. (Provider developers likely need to implement their own custom plan modifiers for this level of planning complexity.)

### References

- #709
- #711

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/pull/711
Closes #717

This change implements a new plan modifier to cover cases where `UseStateForUnknown` will now raise an implementation error when under a list or set nested attribute or block. Provider developers can supply all configurable and identifying nested object attributes which can be used to align and preserve prior state data for this attribute.

An example schema implementation:

```go
schema.SetNestedAttribute{
			"computed_attr": schema.StringAttribute{
				Computed: true,
					// Preseve this computed value during updates.
					stringplanmodifier.MatchElementStateForUnknown(
						// Identify matching prior state value based on configurable_attr
						path.MatchRelative().AtParent().AtName("configurable_attr"),
						// ... potentially others ...
					),
				},
			},
			"configurable_attr": schema.StringAttribute{
				Required: true,
			},
		},
	},
	Optional: true,
}
```

### DIFF
--- a/internal/fwplanmodifier/doc.go
+++ b/internal/fwplanmodifier/doc.go
@@ -1,0 +1,5 @@
+// Package fwplanmodifier contains shared implementation details for framework-
+// defined plan modifiers. The framework should only ever contain critical and
+// generic plan modifiers that are applicable in all providers to implement core
+// Terraform planning functionality.
+package fwplanmodifier
--- a/internal/fwplanmodifier/framework_implementation_error_diagnostics.go
+++ b/internal/fwplanmodifier/framework_implementation_error_diagnostics.go
@@ -1,0 +1,35 @@
+package fwplanmodifier
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+)
+
+// FrameworkImplementationErrorDiag returns an error diagnostic intended for
+// when a framework-defined schema plan modifier reached an unexpected
+// implementation issue.
+func FrameworkImplementationErrorDiag(p path.Path, details string) diag.Diagnostic {
+	return diag.NewAttributeErrorDiagnostic(
+		p,
+		"Framework Plan Modifier Implementation Error",
+		"A framework-defined plan modifier encountered an unexpected implementation issue which could cause unexpected behavior or panics. "+
+			"This is always an issue with terraform-plugin-framework and should be reported to the provider developers.\n\n"+
+			fmt.Sprintf("Path: %s\n", p)+
+			"Details: "+details,
+	)
+}
+
+// PlanValueTypeAssertionDiag returns an error diagnostic intended for when a
+// schema plan modifier with a shared implementation did not return the expected
+// type in the response for the typed response.
+func PlanValueTypeAssertionDiag(p path.Path, requestType attr.Value, responseType attr.Value) diag.Diagnostic {
+	return FrameworkImplementationErrorDiag(
+		p,
+		"The shared implementation responded with an unexpected type.\n"+
+			fmt.Sprintf("Expected Type: %T\n", requestType)+
+			fmt.Sprintf("Response Type: %T", responseType),
+	)
+}
--- a/internal/fwplanmodifier/framework_implementation_error_diagnostics_test.go
+++ b/internal/fwplanmodifier/framework_implementation_error_diagnostics_test.go
@@ -1,0 +1,90 @@
+package fwplanmodifier_test
+
+import (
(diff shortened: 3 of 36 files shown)


bflad:
I do want to take a look at implementing this generically so its once-and-done for each list/set since that could solve a few different use cases and could enable a single `UseStateForUnknown` plan modifier to continue working, so I'm going to temporarily mark this as draft.

bflad:
https://github.com/hashicorp/terraform-plugin-framework/issues/720 is an alternate proposal

bflad:
This solution is not desirable for many reasons, so closing it out for now. GitHub will nicely keep the code differences though, should the maintainers actually want to go down this route.

Issue #721, opened 2023-04-11
Trace and debug logging is extremely expensive (x32 slower) for large resource diffs


lawrencejones:
Hey team 👋 

Was unsure if this would come under bug or enhancement, but given that this is preventing my provider from creating a plan, I went for bug.

tl;dr: we have a resource that accepts a large and deeply nested object. Trace and debug logs produced by the SDK around setting defaults/state modifications/etc are so verbose that it causes planning to become much more expensive (16s -> 527s, so 32x).

Having profiled the provider as it produces the plan, it appears the majority of the cost comes from the context based logger. Generally logging this much can be expensive, but particularly reflection-based runtime loggers that pull context.Value frequently, as the time adds up.

Attaching the profile below where you can see of the 28.52s sample, we spent 22.38s in `context.(*valueCtx).Value` as part of the logging library, and the rest in logging adjacent functions.

```
(pprof) top 100Showing nodes accounting for 28.08s, 98.46% of 28.52s totalDropped 105 nodes (cum <= 0.14s)
      flat  flat%   sum%        cum   cum%    15.34s 53.79% 53.79%     19.21s 67.36%  runtime.efaceeq     3.76s 13.18% 66.97%      3.77s 13.22%  runtime.strequal
     3.49s 12.24% 79.21%      3.49s 12.24%  syscall.syscall     3.16s 11.08% 90.29%     22.38s 78.47%  context.value     1.50s  5.26% 95.55%      1.50s  5.26%  runtime.madvise
     0.66s  2.31% 97.86%      0.66s  2.31%  runtime.memclrNoHeapPointers
     0.16s  0.56% 98.42%      0.20s   0.7%  runtime.scanobject
     0.01s 0.035% 98.46%      0.76s  2.66%  runtime.mallocgc
         0     0% 98.46%     22.38s 78.47%  context.(*valueCtx).Value
         0     0% 98.46%      4.21s 14.76%  github.com/hashicorp/go-hclog.(*intLogger).Trace
         0     0% 98.46%      4.29s 15.04%  github.com/hashicorp/go-hclog.(*intLogger).log
         0     0% 98.46%      0.80s  2.81%  github.com/hashicorp/go-hclog.(*intLogger).logJSON
         0     0% 98.46%      3.49s 12.24%  github.com/hashicorp/go-hclog.(*writer).Flush
[...]

PR #722, opened 2023-04-11
Shortcut logging when the level isn't active

lawrencejones:
This commit addresses this issue:
https://github.com/hashicorp/terraform-plugin-framework/issues/721

When working with large terraform resource attributes, logging can become the primary cost to running plan/apply. This is because the context logger used by tfsdklog is very expensive, the SDK framework itself is very verbose, and even the log level means Debug/Trace logs won't be emitted we still build the log.

As we log via the internal logging package anyway, we can avoid this performance issue by maintaining a package-local log-level which is initialised at the same time as we init the contextual logger. Then our logging package can avoid calling into the tfsdklog (which is where we pay the performance penalty) unless it believes the level is active, which in most cases avoids 99% of the log cost.

### DIFF
--- a/internal/logging/context.go
+++ b/internal/logging/context.go
@@ -3,7 +3,17 @@ package logging
 import (
 	"context"
 
+	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/terraform-plugin-log/tfsdklog"
+)
+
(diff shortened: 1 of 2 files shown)


bflad:
Please refer to https://github.com/hashicorp/terraform-plugin-framework/issues/721#issuecomment-1513971056.

PR #730, opened 2023-04-14
Update index.mdx

jbinko:
return &ThingResource{}


bflad (approved):
Thank you for this fix, @jbinko 🚀

PR #731, opened 2023-04-15
Update update.mdx

jbinko:
StringValue to ValueString


bflad (approved):
😎 Thanks so much for this fix, @jbinko 🚀

PR #745, opened 2023-06-01
[COMPLIANCE] Add copywrite config, GH workflow and headers

austinvalle:
Closes #656 

This PR introduces:
- `make generate` command that will auto-generate all copyright headers for this repo (hooks into `go generate`)
- GH workflow that will error if copyright headers aren't present
- Addition to the `CONTRIBUTING.md` on how to run license autogenerating commands


go.mod
austinvalle, line 1: Lots of dependencies coming from the `copywrite` tool 😞

website/scripts/should-build.sh
```diff
@@ -1,4 +1,7 @@
 #!/usr/bin/env bash
+# Copyright (c) HashiCorp, Inc.
```
bflad, line 2: I think this is okay to merge in since the digital team should be adding this to their management and we can flag it during review if the header is trying to be removed.

.github/workflows/compliance.yml
```diff
@@ -0,0 +1,16 @@
...
+  contents: read
+
+jobs:
+  # Reference: ENGSRV-059
+  copywrite:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
+      - uses: hashicorp/setup-copywrite@867a1a2a064a0626db322392806428f7dc59cb3e # v1.1.2
+      - run: copywrite headers --plan
```
bflad, line 16: Let's also enable `LICENSE` file checking while we're here 😄 

```suggestion
      - run: copywrite headers --plan
      - run: copywrite license
```

It will return a non-zero exit status if there's an issue with the file:

```
$ copywrite license
Licensing under the following terms: MPL-2.0
Using year of initial copyright: 2021
Using copyright holder: HashiCorp, Inc.

Validated file: LICENSE
Validating presence of license header
2023-06-02T09:50:08.804-0400 [ERROR] cli: license file has a copyright statement, but it is malformed; Expected to find: "Copyright (c) 2021 HashiCorp, Inc." Please resolve this manually
Error: license file has a copyright statement, but it is malformed; Expected to find: "Copyright (c) 2021 HashiCorp, Inc." Please resolve this manually
$ echo $?
1
```

bflad (approved):
One minor addition to the workflow, but otherwise looks great 🚀

.github/workflows/compliance.yml
austinvalle: Good call, added! ⚡

go.mod
austinvalle: This has been resolved by adding a dedicated `go.mod` just for the `/tools` directory

PR #749, opened 2023-06-02
Add `tools` folder to dependabot and add license dry run flag

austinvalle:
- Missed adding the `tools` folder to dependabot in #745
- Ensures that an error is thrown if the license is malformed or missing


bendbennett (approved):
LGTM 🎸

Issue #709, opened 2023-03-30
Unexpected Behavior With Plan Modifier (UseStateForUnknown) Prior State Under Lists and Sets


rschmied:
### Module version

```
github.com/hashicorp/terraform-plugin-framework v1.2.0
Terraform v1.4.2 on linux_amd64
```

### Relevant provider source code

```go
func LabGroup() map[string]schema.Attribute {
	return map[string]schema.Attribute{
		"id": schema.StringAttribute{
			Description: "Group ID (UUID).",
			Optional:    true,
			Computed:    true,
			PlanModifiers: []planmodifier.String{
				stringplanmodifier.UseStateForUnknown(),
			},
		},
		"name": schema.StringAttribute{
			Description: "Descriptive group name.",
			Computed:    true,
			PlanModifiers: []planmodifier.String{
				stringplanmodifier.UseStateForUnknown(),
			},
		},
		"permission": schema.StringAttribute{
			MarkdownDescription: "Permission, either `read_only` or `read_write`.",
			Optional:            true,
			Computed:            true,
			PlanModifiers: []planmodifier.String{
				stringplanmodifier.UseStateForUnknown(),
			},
			Validators: []validator.String{
				cmlvalidator.GroupPermission{},
			},
		},
	}
}

func Lab() map[string]schema.Attribute {
	return map[string]schema.Attribute{
		"id": schema.StringAttribute{
			Computed:    true,
			Description: "Lab identifier, a UUID.",
			PlanModifiers: []planmodifier.String{
				stringplanmodifier.UseStateForUnknown(),
			},
		},
		// attributes ommitted here
		"groups": schema.SetNestedAttribute{
			Optional:    true,
			Computed:    true,
			Description: "Groups assigned to the lab.",
			NestedObject: schema.NestedAttributeObject{
				Attributes: LabGroup(),
			},
			PlanModifiers: []planmodifier.Set{
				setplanmodifier.UseStateForUnknown(),
			},
		},
	}
}

```

### Terraform Configuration Files

```hcl
resource "cml2_group" "group1" {
	name       = "user_acc_lab_test_group1"
}

resource "cml2_group" "group2" {
	name       = "user_acc_lab_test_group2"
}

resource "cml2_lab" "test" {
	title       = "something"
	description = "something more"
	notes   = "somenotes"
	groups = [
		{
			id = cml2_group.group1.id
			permission = "read_only"
		},
		{
			id = cml2_group.group2.id
[...]

PR #750, opened 2023-06-06
resource/schema: Revert prevent UseStateForUnknown data misalignment by raising implementation error when under a list or set

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/709
Reference: https://github.com/hashicorp/terraform-plugin-framework/pull/711

This unblocks the v1.3.0 release by reverting the `UseStateForUnknown` implementation errors introduced as an interim solution to the originally reported data alignment issue with the plan modifier. Leaving the half state of the `UseStateForUnknown` implementation error would leave provider developers in an unfortunately much worse situation where existing implementations would break without a straightforward path to fixing their schema implementations. The framework needs additional reworking of plan data (potentially reimplementing some of the Terraform core logic in this regard) as a fuller solution for the issue, rather than trying to shoe-horn a solution only for framework-defined plan modifiers.

This changeset reverts a lot of the prior code changes that were part of the interim implementation error logic, however it does notably leave the additional unit testing that highlights the current (sometimes unfortunate) behavior of the `UseStateForUnknown` plan modifier so that future efforts can show the expected behavior changes of updating the framework planning logic.


bendbennett (approved):
LGTM 🎸

austinvalle (approved):
🚀

PR #753, opened 2023-06-09
Handle nested attribute and block definition equality with Equal() methods

### LINKED ISSUE #752: Attribute and Block Equal Methods Do Not Account for Nested Attributes/Blocks
### Module version

```
1.3.0
```

This type of logic would generally be found where shared libraries are attempting to manage custom schema definitions, then unit testing the result.

```go
func TestBlock(t *testing.T) {
	t.Parallel()

	type testCase struct {
		expected schema.Block
	}
	tests := map[string]testCase{
		"read": {
			expected: schema.SingleNestedBlock{
				CustomType: timeouts.Type{
					ObjectType: types.ObjectType{
							"read": types.StringType,
						},
					},
				},
				// Adjusting anything below will not cause test failure
					"read": schema.StringAttribute{
						Optional: true,
						Validators: []validator.String{
							validators.TimeDuration(),
						},
					},
				},
				// Adjusting anything above will not cause test failure
			},
		},
	}

	for name, test := range tests {
		name, test := name, test
			t.Parallel()
			actual := timeouts.Block(context.Background())

			if diff := cmp.Diff(actual, test.expected); diff != "" {
				t.Errorf("unexpected block difference: %s", diff)
			}
		})
	}
}
```

N/A

### Debug Output

N/A

### Expected Behavior

The unit test should fail if `Attributes` is not exactly equivalent to the test expectations.

### Actual Behavior

The unit test will pass silently. Each of the schema packages delegates to a shared `internal/fwschema` package functions which omit handling nested attributes for attributes and omit handling nested attributes/blocks for blocks.

### Steps to Reproduce

`go test`

### References

- https://github.com/hashicorp/terraform-plugin-framework-timeouts/pull/52

bflad:
Closes #752

Previously before logic updates:

```
--- FAIL: TestSingleNestedAttributeEqual (0.00s)
    --- FAIL: TestSingleNestedAttributeEqual/different-attribute-definitions (0.00s)
        /Users/bflad/src/github.com/hashicorp/terraform-plugin-framework/datasource/schema/single_nested_attribute_test.go:182: unexpected difference:   bool(
            - 	true,
            + 	false,
              )
```

### DIFF
--- a/internal/fwschema/nested_attribute.go
+++ b/internal/fwschema/nested_attribute.go
@@ -17,3 +17,18 @@ type NestedAttribute interface {
 	// does not represent nested attributes.
 	GetNestingMode() NestingMode
 }
+
+// NestedAttributesEqual is a helper function to perform equality testing on two
+// NestedAttribute. NestedAttribute Equal implementations should still compare
+// the concrete types in addition to using this helper.
+func NestedAttributesEqual(a, b NestedAttribute) bool {
+	if a.GetNestingMode() != b.GetNestingMode() {
+		return false
+	}
+
+	if !a.GetNestedObject().Equal(b.GetNestedObject()) {
+		return false
+	}
+
+	return AttributesEqual(a, b)
+}
--- a/.changes/unreleased/BUG FIXES-20230609-140931.yaml
+++ b/.changes/unreleased/BUG FIXES-20230609-140931.yaml
@@ -1,0 +1,6 @@
+kind: BUG FIXES
+body: 'datasource/schema: Ensure nested attribute and block Equal methods check nested
+  attribute and block definition equality'
+time: 2023-06-09T14:09:31.199452-04:00
+custom:
+  Issue: "752"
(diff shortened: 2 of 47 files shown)


internal/fwschema/nested_attribute.go
```diff
@@ -17,3 +17,18 @@ type NestedAttribute interface {
...
+func NestedAttributesEqual(a, b NestedAttribute) bool {
+	if a.GetNestingMode() != b.GetNestingMode() {
+		return false
+	}
+
+	if !a.GetNestedObject().Equal(b.GetNestedObject()) {
+		return false
+	}
+
+	return AttributesEqual(a, b)
```
austinvalle, line 33: Is there a reason that we first compare the nested objects before we compare the attributes? If I'm understanding the `BlocksEqual` function above it is the other way around (I don't think it matters either way, just curious)

austinvalle (approved):
Looks great! Just one clarifying question 🚀

internal/fwschema/nested_attribute.go
bflad: Nope and in fact, maybe we should compare the "easier" bits first to save a little on performance. Would you want to submit a followup PR to adjust these? (I don't think they need a benchmark test or anything)

austinvalle: sounds good, I have no problem submitting the followup PR 👍🏻

PR #762, opened 2023-06-12
Adjust `NestedAttributesEqual` to perform shallow compare first

austinvalle:
Follow-up to #753 

Adjusts the `NestedAttributesEqual` method to perform the compare on the "shallow" portions of the attribute before comparing the nested attributes, similar to `BlocksEqual`: https://github.com/hashicorp/terraform-plugin-framework/blob/532797d138849d5b4404b7711157f360fe0decc2/internal/fwschema/block.go#L71


bflad (approved):
Looks good to me 🚀

PR #769, opened 2023-06-13
docs: Additional type-based semantic equality information

bflad:
Per out-of-band feedback:

- Add mentions of custom types into SDK migration documentation for attribute fields
- Add mention in custom type semantic equality section about framework automatic collection/object behaviors with element/underlying attribute types.


austinvalle (approved):
Looks good! 🚀

PR #770, opened 2023-06-14
docs: Adjust semantic equality ordering and use stronger wording for overriding methods

bflad:
This quick change will bring the type embedding override lists for schema and value types closer to the example code. It also adjusts the wording of "should be overridden" to "must be overridden" to try to help prevent confusing situations or error messages.


austinvalle (approved):
🚀

Issue #754, opened 2023-06-09
Plan Modifiers Doesn't Work with Custom String Type


zliang-akamai:
### Module version

```
v1.3
```

### Relevant provider source code

```go

var frameworkResourceSchema = schema.Schema{
	Attributes: map[string]schema.Attribute{
		"expiry": schema.StringAttribute{
			Description: "When this token will expire. Personal Access Tokens cannot be renewed, so after " +
				"this time the token will be completely unusable and a new token will need to be generated. Tokens " +
				"may be created with 'null' as their expiry and will never expire unless revoked. Format: " +
				helper.TIME_FORMAT,
			Optional: true,
			PlanModifiers: []planmodifier.String{
				stringplanmodifier.RequiresReplaceIf(
					func(
						ctx context.Context,
						sr planmodifier.StringRequest,
						rrifr *stringplanmodifier.RequiresReplaceIfFuncResponse,
					) {
						rrifr.RequiresReplace = !helper.CompareTimeStrings(
							sr.PlanValue.ValueString(),
							sr.StateValue.ValueString(),
							time.RFC3339,
						)
					},
					RequireReplacementWhenScopesChangedDescription,
					RequireReplacementWhenScopesChangedDescription,
				),
			},
			Validators: []validator.String{
				helper.NewDateTimeStringValidator(time.RFC3339),
			},
			CustomType: customtypes.RFC3339TimeStringType{},
		},
		"created": schema.StringAttribute{
			Description: "The date and time this token was created.",
			Computed:    true,
			PlanModifiers: []planmodifier.String{
				stringplanmodifier.UseStateForUnknown(),
			},
			CustomType: customtypes.RFC3339TimeStringType{},
		},
		// ... other attrs
	},
}

```

### Terraform Configuration Files

```hcl
resource "linode_token" "foo" {
  scopes = "linodes:read_only"
  expiry = "2030-01-02T04:04:05+01:00"
}
```

### Expected Behavior
Successfully apply the configuration to the cloud.

### Actual Behavior
```
Planning failed. Terraform encountered an error while generating this plan.

╷
│ Error: Value Conversion Error
│ 
│   with linode_token.foo,
│   on main.tf line 16, in resource "linode_token" "foo":
│   16:   expiry = "2030-01-02T04:04:05Z"
│ 
│ An unexpected error was encountered while verifying an attribute value matched its expected type to prevent unexpected behavior or panics. This is always an error in the provider. Please
[...]

PR #715, opened 2023-04-05
internal/reflect: Raise framework errors instead of upstream errors or panics with mismatched types

bflad:
Closes #566

Previously in a provider where a `types.Object` zero value was used in `(tfsdk.State).Set()`:

```
panic: AttributeName("single_nested_attribute"): can't use tftypes.Object[] as tftypes.Object["nested_attribute":tftypes.String]

goroutine 448 [running]:
github.com/hashicorp/terraform-plugin-go/tftypes.NewValue(...)
	/Users/bflad/go/pkg/mod/github.com/hashicorp/terraform-plugin-go@v0.14.3/tftypes/value.go:273
github.com/hashicorp/terraform-plugin-framework/internal/reflect.FromStruct({0x1035b3e40, 0x140001fa870}, {0x1035b86f8?, 0x140001faae0?}, {0x103504480?, 0x140001ff400?, 0x10b4975c8?}, {{0x103b8fc60?, 0x50?, 0x1400006c900?}})
	/Users/bflad/go/pkg/mod/github.com/hashicorp/terraform-plugin-framework@v1.2.0/internal/reflect/struct.go:207 +0x73c
github.com/hashicorp/terraform-plugin-framework/internal/reflect.FromValue({0x1035b3e40, 0x140001fa870}, {0x1035b6d80?, 0x140001faae0}, {0x103504480, 0x140001ff400}, {{0x103b8fc60?, 0x0?, 0x14000462f28?}})
	/Users/bflad/go/pkg/mod/github.com/hashicorp/terraform-plugin-framework@v1.2.0/internal/reflect/outof.go:54 +0x6c0
github.com/hashicorp/terraform-plugin-framework/internal/reflect.FromPointer({0x1035b3e40, 0x140001fa870}, {0x1035b6d80?, 0x140001faae0?}, {0x10342a1e0?, 0x140001ff360?, 0x0?}, {{0x103b8fc60?, 0x0?, 0xffffffffffffffff?}})
	/Users/bflad/go/pkg/mod/github.com/hashicorp/terraform-plugin-framework@v1.2.0/internal/reflect/pointer.go:115 +0x198
github.com/hashicorp/terraform-plugin-framework/internal/reflect.FromValue({0x1035b3e40, 0x140001fa870}, {0x1035b6d80?, 0x140001faae0}, {0x10342a1e0, 0x140001ff360}, {{0x103b8fc60?, 0x0?, 0x10329f143?}})
[...]

PR #771, opened 2023-06-14
internal/fwserver: Add unit test for PlanResourceChange with CustomType and Default

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/pull/715
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/754

This adjusts the existing `PlanResourceChange` unit testing for schema-based default value handling to include a test case of an attribute with both `CustomType` and `Default` set. The framework now performs stricter value type checking and this test case can act as a canary should the framework's storage of data from Terraform be migrated to the framework type system. The testing passes as-is, unlike the original case of both `CustomType` and `PlanModifiers` being set after the validation logic introduction because the default value handling occurs in the underlying terraform-plugin-go type system so there is no need to re-instantiate custom value types.


austinvalle (approved):
Very nice! 🚀

PR #780, opened 2023-06-16
internal/fwschemadata: Ensure Default implementations receive request Path and have response Diagnostics handled

### LINKED ISSUE #778: Default Implementations Never Receive Path Field in Request Type
### Module version

```
v1.3.1
```

Implementing a [`resource/schema/defaults` package](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/resource/schema/defaults) interface, e.g. like this existing implementation:

```go
// StaticValue returns a static object value default handler.
//
// Use StaticValue if a static default value for a object should be set.
func StaticValue(defaultVal types.Object) defaults.Object {
	return staticValueDefault{
		defaultVal: defaultVal,
	}
}

// staticValueDefault is static value default handler that
// sets a value on a object attribute.
type staticValueDefault struct {
	defaultVal types.Object
}

// Description returns a human-readable description of the default value handler.
func (d staticValueDefault) Description(_ context.Context) string {
	return fmt.Sprintf("value defaults to %v", d.defaultVal)
}

// MarkdownDescription returns a markdown description of the default value handler.
func (d staticValueDefault) MarkdownDescription(_ context.Context) string {
	return fmt.Sprintf("value defaults to `%v`", d.defaultVal)
}

// DefaultObject implements the static default value logic.
func (d staticValueDefault) DefaultObject(ctx context.Context, req defaults.ObjectRequest, resp *defaults.ObjectResponse) {
	// req.Path here is problematic
	resp.PlanValue = d.defaultVal
}
```

N/A

### Debug Output

N/A

### Expected Behavior

The request types have a `Path` field, with a filled in `path.Path` value.

### Actual Behavior

The code in `internal/fwschemadata` does not populate the `Path` field appropriately.

### Steps to Reproduce

`go test`

### References

None

bflad:
Closes #778

This changeset also includes panic prevention for `defaults.List`, `defaults.Map`, and `defaults.Set` implementations which do not return `Diagnostics` or `PlanValue`.

Previously before logic update:

```
--- FAIL: TestDataDefault (0.00s)
    --- FAIL: TestDataDefault/bool-attribute-response-diagnostics (0.00s)
        /Users/bflad/src/github.com/hashicorp/terraform-plugin-framework/internal/fwschemadata/data_default_test.go:6959: unexpected diagnostics difference:   diag.Diagnostics(
            - 	nil,
            + 	{
            + 		diag.ErrorDiagnostic{detail: "test error detail", summary: "test error summary"},
            + 		diag.WarningDiagnostic{detail: "test warning detail", summary: "test warning summary"},
            + 	},
              )

--- FAIL: TestDataDefault (0.01s)
    --- FAIL: TestDataDefault/list-attribute-request-path (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x2 addr=0x30 pc=0x100a161ec]

goroutine 14 [running]:
testing.tRunner.func1.2({0x1010fe140, 0x10137c890})
	/opt/homebrew/Cellar/go/1.20.5/libexec/src/testing/testing.go:1526 +0x1c8
testing.tRunner.func1()
	/opt/homebrew/Cellar/go/1.20.5/libexec/src/testing/testing.go:1529 +0x384
panic({0x1010fe140, 0x10137c890})
	/opt/homebrew/Cellar/go/1.20.5/libexec/src/runtime/panic.go:884 +0x204
github.com/hashicorp/terraform-plugin-framework/types/basetypes.ListValue.ToTerraformValue({{0x0, 0x0, 0x0}, {0x0, 0x0}, 0x0}, {0x101154860, 0x1400009c018})
	/Users/bflad/src/github.com/hashicorp/terraform-plugin-framework/types/basetypes/list_value.go:208 +0x5c
github.com/hashicorp/terraform-plugin-framework/internal/fwschemadata.(*Data).TransformDefaults.func1(0x140004ecba0?, {{0x101156f08?, 0x14000165530?}, {0x0?, 0x0?}})
	/Users/bflad/src/github.com/hashicorp/terraform-plugin-framework/internal/fwschemadata/data_default.go:168 +0x1324
```


internal/fwschemadata/data_default.go
```diff
@@ -75,76 +75,229 @@ func (d *Data) TransformDefaults(ctx context.Context, configRaw tftypes.Value) d
...
+
+			defaultValue.DefaultObject(ctx, req, &resp)
+
+			diags.Append(resp.Diagnostics...)
+
+			if resp.Diagnostics.HasError() {
+				return tfTypeValue, nil
+			}
+
+			logging.FrameworkTrace(ctx, fmt.Sprintf("setting attribute %s to default value: %s", fwPath, resp.PlanValue))
```
austinvalle, line 246: curious q: Is there a reason why collection types need a warning nil check and the structural type, Object, doesn't?

Referring to the other `resp.PlanValue.ElementType(ctx) == nil` checks for map, set, list

austinvalle (approved):
One question that's mostly just for me 😆 

🚀 🚀 🚀

internal/fwschemadata/data_default.go
bflad: We could consider an object value check here to ensure that a provider defined object value's type matches the schema's defined/expected type and raise a warning/error in that case, rather than having it error/panic later in a different way.

bflad: I created a quick note for myself to double check the framework's current behavior in the above mentioned scenario and will raise an issue if it feels warranted to adjust that behavior.

PR #785, opened 2023-06-23
internal/fwserver: Delay resource schema caching to occur after GetProviderSchema RPC

### LINKED ISSUE #784: Delay Schema Caching Until Resource Operations
### Module version

```
v1.3.1
```

### Description

Currently, this Go module will cache data source and managed resource (collectively called "resource" from now on) schemas when the `GetProviderSchema` RPC is called. This caching was introduced to prevent the multiple RPCs associated with a single resource from continually invoking the `Schema` method for no benefit as by Terraform's definition, schemas must be static. For large providers, this is very problematic as it will actively keep unused schema information that is likely not needed again in memory. Providers do not have an opt-out for this framework behavior.

### Proposal

Rather than caching resource schemas during `GetProviderSchema` RPC, which in turn will keep all schemas in memory, only perform the caching when another resource-based operation RPC occurs. This should ensure that:

- Go garbage collection can occur after the `GetProviderSchema` RPC
- Actively used schemas still receive the slight performance benefit of caching rather than repeatedly invoking the `Schema` method

It'll be worth exploring with upstream if the schema fetching can be split across RPCs to prevent a large allocation for `GetProviderSchema` RPC response.

### References

- https://github.com/hashicorp/terraform-plugin-framework/issues/299
- https://github.com/hashicorp/terraform-provider-aws/issues/31722

bflad:
Closes #784

### DIFF
--- a/.changes/unreleased/BUG FIXES-20230623-172651.yaml
+++ b/.changes/unreleased/BUG FIXES-20230623-172651.yaml
@@ -1,0 +1,5 @@
+kind: BUG FIXES
+body: 'providerserver: Prevented caching of unused data and managed resource schemas'
+time: 2023-06-23T17:26:51.383868-04:00
+custom:
+  Issue: "784"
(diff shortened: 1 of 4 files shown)


bendbennett (approved):
LGTM 🚀

PR #790, opened 2023-06-29
internal/reflection: Migrate reflection tests for struct tags to exported methods

austinvalle:
This PR will make it easier for us to verify internal refactoring that is made to any of the struct reflection logic. That includes:
- Moving the existing `Struct` error tests into a single table-driven test (same as `FromStruct`)
- Moving all the internal testing of `getStructTags` to use the exported `Struct` method
   - The only test that was removed was `TestGetStructTags_notAStruct`, as the exported methods (`FromStruct` & `Struct`) already have tests for this scenario and the logic in `getStructTags` is mostly defensive.
- Added new tests to `FromStruct` for all the error scenarios from `getStructTags` + the success scenario with ignored tags


bflad (approved):
Looks good to me 🚀

Issue #775, opened 2023-06-16
Performance issue when run Terraform Plan with many Block Sets or NestedAttribute Sets


jeremmfr:
### Module version
```
v1.2.0 & v1.3.1
```

### Description

After migrating resources of my provider [jeremmfr/junos](https://github.com/jeremmfr/terraform-provider-junos) from SDK plugin to this framework plugin, the time to run a `terraform plan` with a config that have many block sets has been increased very significantly.

For 1000 items of block set, a `terraform plan` (with empty state) with provider including module `sdk/v2` take ~1,3 seconds and take ~ **70 seconds** with provider including `plugin-framework`.
A user of my provider has a time of 19 minutes to run terraform plan with the new version of the provider including `plugin-framework` instead of 52 seconds with the previous version of the provider including module `sdk/v2`

### Tests

I tested to remove all `PlanModifiers` and `Validators` fields on each attribute, and `ValidateConfig` function for resources, using the latest version of this module (v1.3.1) but there's no change.
The problem disappears if replace `schema.SetNestedBlock` with `schema.ListNestedBlock`: 70 seconds to 1 second.

I reproduce the problem with provider  `hashicups` and `hashicups_order` resources.
I tested 1000 blocks`items` on a resource `hashicups_order` with different modifications of `items` block.

[hashicorp/terraform-provider-hashicups](https://github.com/hashicorp/terraform-provider-hashicups) schema.TypeSet : 3,8 - 4 seconds
[hashicorp/terraform-provider-hashicups](https://github.com/hashicorp/terraform-provider-hashicups) schema.TypeList : 0,5 - 0,8 seconds
[hashicorp/terraform-provider-hashicups-pf](https://github.com/hashicorp/terraform-provider-hashicups-pf) schema.SetNestedBlock : **49,1 - 49,7** seconds
[hashicorp/terraform-provider-hashicups-pf](https://github.com/hashicorp/terraform-provider-hashicups-pf) schema.ListNestedBlock : 0,8 - 1,1 seconds
[hashicorp/terraform-provider-hashicups-pf](https://github.com/hashicorp/terraform-provider-hashicups-pf) schema.SetNestedAttribute : **49,0 - 49,4** seconds
[hashicorp/terraform-provider-hashicups-pf](https://github.com/hashicorp/terraform-provider-hashicups-pf) schema.ListNestedAttribute : 0,7 - 0,9 seconds

### Relevant provider source code

[...]

PR #791, opened 2023-07-03
types/basetypes: Reduce memory allocations of (ObjectValue).ToTerraformValue()

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/775

The prior implementation used the exported `AttributeTypes()` method, which in this case was unnecessarily creating a copy of the map when it was only be used for walking the existing data. At scale, such as handling 1000+ objects, this creates unnecessary pressure on the Go garbage collector.

`benchstat` comparison with new benchmark test:

```text
goos: darwin
goarch: arm64
pkg: github.com/hashicorp/terraform-plugin-framework/types/basetypes
                                   │  original   │              proposed               │
                                   │   sec/op    │   sec/op     vs base                │
ObjectValueToTerraformValue1000-10   408.4µ ± 1%   318.4µ ± 1%  -22.02% (p=0.000 n=10)

                                   │   original   │               proposed               │
                                   │     B/op     │     B/op      vs base                │
ObjectValueToTerraformValue1000-10   449.2Ki ± 0%   286.6Ki ± 0%  -36.19% (p=0.000 n=10)

                                   │  original   │              proposed              │
                                   │  allocs/op  │  allocs/op   vs base               │
ObjectValueToTerraformValue1000-10   2.051k ± 0%   2.020k ± 0%  -1.51% (p=0.000 n=10)
```

### DIFF
--- a/.changes/unreleased/BUG FIXES-20230703-123518.yaml
+++ b/.changes/unreleased/BUG FIXES-20230703-123518.yaml
@@ -1,0 +1,6 @@
+kind: BUG FIXES
+body: 'types/basetypes: Minor reduction of memory allocations for `ObjectValue` type
+  `ToTerraformValue()` method, which decreases provider operation durations at scale'
+time: 2023-07-03T12:35:18.484275-04:00
+custom:
+  Issue: "775"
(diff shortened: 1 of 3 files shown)


bendbennett (approved):
LGTM 🚀

PR #800, opened 2023-07-13
ci: Add TSCCR verification job to GitHub Actions workflow

bflad:
Reference: https://github.com/hashicorp/terraform-providers-devex-internal/issues/130


bflad:
I'm reaching out to the Security team to followup about this.

bflad:
This is being tracked internally -- this PR does not need to sit open in the meantime.

Issue #793, opened 2023-07-06
Value Conversion Error for timeouts attribute


aorith:
### Module version

```
github.com/hashicorp/terraform-plugin-framework v1.3.2
```
### Relevant provider source code

```go
func (r *siteResource) Schema(ctx context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
	resp.Schema = schema.Schema{
		Description:         "Manages company sites (domains)",
		MarkdownDescription: "Manages company sites (domains)",

		Attributes: map[string]schema.Attribute{
			"timeouts": timeouts.Attributes(ctx, timeouts.Opts{
				Create: true,
			}),

```

```go
type Site struct {
	Timeouts timeouts.Value `tfsdk:"timeouts"`
	ID       types.Int64    `tfsdk:"id"`
	Domain   types.String   `tfsdk:"domain"`
	Active   types.Bool     `tfsdk:"active"`
}
```

### Terraform Configuration Files

```hcl
resource "transparentedge_site" "all" {
  for_each = var.sites
  domain   = each.key
  timeouts = {
    create = "120s"
  }
}
```
### Debug Output

```
│ Expected type: types.ObjectType["create":basetypes.StringType]
│ Value type: types.ObjectType["create":basetypes.StringType]
│ Path: timeouts
╵
╷
│ Error: Value Conversion Error
│
│   with transparentedge_site.all["demo.transparentedge.eu"],
│   on main.tf line 38, in resource "transparentedge_site" "all":
│   38:   timeouts = {
│   39:     create = "120s"
│   40:   }
```

### Expected Behavior

The type is correct, it should not complain.

### Actual Behavior

Value conversion error

### Steps to Reproduce

1. `terraform init`
2. `terraform apply`

### References

- https://github.com/hashicorp/terraform-provider-aws/issues/32322

bflad:
Thank you for reporting this, @aorith, and sorry you are running into trouble here. The code snippets provided _seem_ correct at first glance, but we will take a deeper look.

bflad:
By the way, what version of terraform-plugin-framework-timeouts are you using?

aorith:
> By the way, what version of terraform-plugin-framework-timeouts are you using?

`v0.4.0`

I was updating from:

```
github.com/hashicorp/terraform-plugin-framework v1.2.0
github.com/hashicorp/terraform-plugin-framework-timeouts v0.3.1
```

[...]

PR #806, opened 2023-07-19
internal/reflect: Use underlying terraform-plugin-go types to validate provider given values

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/pull/715
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/793
Reference: https://github.com/hashicorp/terraform-provider-aws/issues/32441

The original intention of this logic was to raise a framework error instead of underlying terraform-plugin-go error or panic when providers returned an errant value from their logic. In practice though, comparing the framework types (which may include either base or custom types) was too strict. Technically, as long as the underlying terraform-plugin-go types are equivalent, then this check should pass. Terraform will catch any errant modifications of the actual values, should there be a difference between the base types and custom types.


bendbennett (approved):
LGTM 🚀

Issue #805, opened 2023-07-19
Attributes inside nested block are not validated


mkuzmin:
### Module version
```
1.3.2
```

### Terraform Configuration Files
This is an example with official provider to reproduce the issue easier. I get the same behavior with my custom code.

```hcl
resource "random_id" "server" {
  byte_length = 8
  keepers     = {
    # Unknown attribute
    qwerty = 123
    ami_id = "123"
  }
}
```

### Expected Behavior
Validation should fail:
```
Error: Unsupported argument
│ 
│   on project.tf line 9, in resource "random_id" "server":
│    9:   qwerty = 123
│ 
│ An argument named "qwerty" is not expected here.
```

### Actual Behavior
Error is ignored:
```
Terraform will perform the following actions:

  # random_id.server will be created
  + resource "random_id" "server" {
      + b64_std     = (known after apply)
      + b64_url     = (known after apply)
      + byte_length = 8
      + dec         = (known after apply)
      + hex         = (known after apply)
      + id          = (known after apply)
      + keepers     = {
          + "ami_id" = "123"
          + "qwerty" = "123"
        }
    }
```

austinvalle:
Hi there @mkuzmin 👋🏻 !

The `random_id.keepers` attribute you're referencing in your reproduction doesn't have validation on it's attributes as it's an arbitrary map of values, represented as a `MapAttribute` of `types.StringType`.

https://github.com/hashicorp/terraform-provider-random/blob/d62a15b2ba285278698fbf0dde65d8022124c2c8/internal/provider/resource_id.go#L58

> I get the same behavior with my custom code.

Can you share the schema, terraform config, and error message of your code that you're expecting to produce a validation error?

bflad:
Drive-by note to leave some additional breadcrumbs/linking/context: on the Terraform core side of things, https://developer.hashicorp.com/terraform/language/expressions/types#maps-objects is the relevant documentation here. It could potentially use some more verbiage about maps having arbitrary key names while objects have defined attribute names. That change would need to occur upstream. On the provider side of things, https://developer.hashicorp.com/terraform/plugin/framework/handling-data/attributes#map is potentially missing a word like "arbitrary" in front of "string keys". We do have plans to expand the framework type documentation via https://github.com/hashicorp/terraform-plugin-framework/issues/695 so its not all colocated on one page and maybe that separation via navigation could help.

chrismarget-j:
[...]

PR #807, opened 2023-07-20
docs: Use arbitrary with map keys and explicit with object attribute names

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/805


austinvalle (approved):
lgtm 🚀

PR #817, opened 2023-08-03
types/basetypes: Fix Float64Attribute precision comparisons

### LINKED ISSUE #815: Floating point precision differences after migrating from SDKv2
I recently migrated the `honeycombio_trigger` resource from the Plugin SDKv2 to the Plugin Framework and received a bug report that the way floating point numbers are being handled is differently.

I think this is not so much a 'bug' -- I can see increased precision as a feature! -- but I am looking for guidance on how to handle this case. A custom `planmodifier`?

### Module version

```
github.com/hashicorp/terraform-plugin-framework v1.3.3
```

https://github.com/honeycombio/terraform-provider-honeycombio/blob/main/internal/provider/trigger_resource.go#L132-L153

```hcl
data "honeycombio_query_specification" "test" {
  calculation {
    op     = "COUNT"
  }

  time_range = 1200
}

resource "honeycombio_trigger" "test" {
  name = "floating point test"

  dataset  = var.dataset
  query_id = data.honeycombio_query_specification.test.id

  threshold {
    op    = ">"
    value = 1 - 0.99
  }
}
```

### Debug Output

https://gist.github.com/jharley/90ec9c38c70858c0e1bdcaaea0707c05

### Expected Behavior

As with the SDKv2-based resource, it parses `threshold[0].value` as 0.01

### Actual Behavior

```
       Error: Provider produced invalid plan
        
        Provider "registry.terraform.io/hashicorp/honeycombio" planned an invalid
        value for honeycombio_trigger.test.threshold[0].value: planned value
        cty.NumberFloatVal(0.01) does not match config value
        cty.MustParseNumberVal("0.010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000003").
        
        This is a bug in the provider, which should be reported in the provider's own
        issue tracker.
```

### Steps to Reproduce

`terraform plan`

Test to reproduce is on this branch: https://github.com/honeycombio/terraform-provider-honeycombio/commit/467e603fdc05ebb2879a1c62c1315a4a04cba8ed

### References

Reported bug: https://github.com/honeycombio/terraform-provider-honeycombio/issues/335

austinvalle:
Closes #815 
Related corner tests: https://github.com/hashicorp/terraform-provider-corner/pull/170

### Context
Terraform core's `number` type can store up to [512 bits of precision](https://github.com/zclconf/go-cty/blob/main/docs/types.md#ctynumber) for floating numbers. This level of precision is typically only encountered when performing an arithmetic operation like below:
```bash
 $ terraform console
> 1 - 0.99
0.010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000003
```

The `Float64Attribute` underlying value of Go's built-in `float64` type was losing this precision and responding with invalid plan value errors seen in #815 .

### Solution
This PR replaces the underlying value of `Float64Attribute` with a `*big.Float` type, which is already supported in [`terraform-plugin-go`](https://github.com/hashicorp/terraform-plugin-go/blob/81f0d496aaa39e43cc185f5c961df5fcd86192da/tftypes/primitive.go#L176). Semantic equality logic has also been implemented to ensure that `Computed` values do not replace more precise prior values.

### DIFF
--- a/.changes/unreleased/BUG FIXES-20230803-120411.yaml
+++ b/.changes/unreleased/BUG FIXES-20230803-120411.yaml
@@ -1,0 +1,6 @@
+kind: BUG FIXES
+body: 'types/basetypes: Fix Float64Attribute bug by preserving precision from number
+  values'
+time: 2023-08-03T12:04:11.996955-04:00
+custom:
+  Issue: "817"
--- a/.changes/unreleased/BUG FIXES-20230803-120411.yaml
+++ b/.changes/unreleased/BUG FIXES-20230803-120411.yaml
@@ -1,6 +1,6 @@
 kind: BUG FIXES
-body: 'types/basetypes: Fix Float64Attribute bug by preserving precision from number
-  values'
+body: 'types/basetypes: Prevented Float64Value Terraform data consistency errors for
+  numbers with high precision floating point rounding errors'
 time: 2023-08-03T12:04:11.996955-04:00
 custom:
   Issue: "817"
--- a/types/basetypes/float64_type.go
+++ b/types/basetypes/float64_type.go
@@ -140,28 +140,24 @@ func (t Float64Type) ValueFromTerraform(ctx context.Context, in tftypes.Value) (
 		return nil, err
 	}
 
-	// Copy the float retrieved from tftypes.Value to avoid any unexpected pass-by-reference shenanigans
-	copiedBigF := new(big.Float)
-	copiedBigF.Copy(bigF)
-
-	f, accuracy := copiedBigF.Float64()
+	f, accuracy := bigF.Float64()
 
 	// Underflow
 	// Reference: https://pkg.go.dev/math/big#Float.Float64
 	if f == 0 && accuracy != big.Exact {
-		return nil, fmt.Errorf("Value %s cannot be represented as a 64-bit floating point.", copiedBigF)
+		return nil, fmt.Errorf("Value %s cannot be represented as a 64-bit floating point.", bigF)
 	}
 
 	// Overflow
 	// Reference: https://pkg.go.dev/math/big#Float.Float64
 	if math.IsInf(f, 0) {
-		return nil, fmt.Errorf("Value %s cannot be represented as a 64-bit floating point.", copiedBigF)
+		return nil, fmt.Errorf("Value %s cannot be represented as a 64-bit floating point.", bigF)
 	}
 
 	// Underlying *big.Float values are not exposed with helper functions, so creating Float64Value via struct literal
 	return Float64Value{
 		state: attr.ValueStateKnown,
-		value: copiedBigF,
+		value: bigF,
 	}, nil
 }
 
--- a/types/basetypes/float64_type.go
+++ b/types/basetypes/float64_type.go
@@ -136,6 +136,7 @@ func (t Float64Type) ValueFromTerraform(ctx context.Context, in tftypes.Value) (
 
 	var bigF *big.Float
 	err := in.As(&bigF)
+
 	if err != nil {
 		return nil, err
 	}
--- a/types/basetypes/float64_type.go
+++ b/types/basetypes/float64_type.go
@@ -136,26 +136,33 @@ func (t Float64Type) ValueFromTerraform(ctx context.Context, in tftypes.Value) (
 
 	var bigF *big.Float
 	err := in.As(&bigF)
-
 	if err != nil {
 		return nil, err
 	}
 
-	f, accuracy := bigF.Float64()
+	// Copy the float retrieved from tftypes.Value to avoid any unexpected pass-by-reference shenanigans
+	copiedBigF := new(big.Float)
+	copiedBigF.Copy(bigF)
+
+	f, accuracy := copiedBigF.Float64()
 
 	// Underflow
 	// Reference: https://pkg.go.dev/math/big#Float.Float64
 	if f == 0 && accuracy != big.Exact {
-		return nil, fmt.Errorf("Value %s cannot be represented as a 64-bit floating point.", bigF)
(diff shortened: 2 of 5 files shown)


types/basetypes/float64_value.go
```diff
@@ -68,7 +70,7 @@ func NewFloat64Unknown() Float64Value {
 func NewFloat64Value(value float64) Float64Value {
 	return Float64Value{
 		state: attr.ValueStateKnown,
-		value: value,
+		value: big.NewFloat(value),
```
austinvalle, line 73: This will panic if [`math.Nan()`](https://pkg.go.dev/math#NaN) is provided here. This problem currently exists with the underlying value as `float64`, the panic will just come later when `tftypes.ValidateValue` is called.

Are there any non-breaking changes we can make to smooth this out?

.changes/unreleased/BUG FIXES-20230803-120411.yaml
```diff
@@ -0,0 +1,6 @@
+kind: BUG FIXES
+body: 'types/basetypes: Fix Float64Attribute bug by preserving precision from number
+  values'
```
bflad, line 3: Nit: Maybe this is clearer for developers?

```suggestion
body: 'types/basetypes: Prevented Float64Value Terraform data consistency errors for
  numbers with high precision floating point rounding errors'
```

types/basetypes/float64_type.go
```diff
@@ -136,26 +136,33 @@ func (t Float64Type) ValueFromTerraform(ctx context.Context, in tftypes.Value) (
...
 	if err != nil {
 		return nil, err
 	}
 
-	f, accuracy := bigF.Float64()
+	// Copy the float retrieved from tftypes.Value to avoid any unexpected pass-by-reference shenanigans
+	copiedBigF := new(big.Float)
+	copiedBigF.Copy(bigF)
+
+	f, accuracy := copiedBigF.Float64()
```
bflad, line 147: Turns out I was being overly cautious when describing this and tftypes already does the copying: https://github.com/hashicorp/terraform-plugin-go/blob/81f0d496aaa39e43cc185f5c961df5fcd86192da/tftypes/value.go#L449

types/basetypes/float64_value.go
bflad: Sorry I'm being dense, are you saying this is a new problem or existing problem? I'm not sure we could do anything different here, other than potentially raise our own panic message.

types/basetypes/float64_value.go
```diff
@@ -108,7 +132,7 @@ func (f Float64Value) Equal(other attr.Value) bool {
 		return true
 	}
 
-	return f.value == o.value
+	return f.value.Cmp(o.value) == 0
```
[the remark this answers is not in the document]
bflad: Defensive coding is probably preferable even if its not possible to create a known value with a `nil` `*big.Float` currently.

.changes/unreleased/BUG FIXES-20230803-120411.yaml
austinvalle: Sounds good to me 👍🏻

types/basetypes/float64_value.go
austinvalle: This is an existing problem today, if today you write:

```go
data.FloatAttribute = types.NewFloat64Value(math.Nan())
```

You will get a panic later when `tftypes` attempts to use that float64, I think the first place it happens is with [`tftypes.ValidateValue`](https://github.com/hashicorp/terraform-plugin-framework/blob/f937d7456b05f45e8e8d7fb8dcb6bda806fd9861/types/basetypes/float64_value.go#L118).

Panic raised here:
https://github.com/hashicorp/terraform-plugin-go/blob/81f0d496aaa39e43cc185f5c961df5fcd86192da/tftypes/primitive.go#L347 


When we change this logic, a developer would receive a panic immediately:
```go
// Panic right away! :O
data.FloatAttribute = types.NewFloat64Value(math.Nan())
```

austinvalle: So the same result (a panic error), but at different points. I can create a separate issue for this if we don't want to handle it now

types/basetypes/float64_value.go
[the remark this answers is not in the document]
austinvalle: Fixed in dc93cee ⚡

types/basetypes/float64_type.go
austinvalle: Good catch, removed copy in 4425e1c ⚡

types/basetypes/float64_value.go
```diff
@@ -108,7 +132,12 @@ func (f Float64Value) Equal(other attr.Value) bool {
 		return true
 	}
 
-	return f.value == o.value
+	// Not possible to create a known Float64Value with a nil value, but check anyways
+	if f.value == nil || o.value == nil {
+		return false
```
bflad, line 137: I'm not sure if its better or worse to treat both values being `nil` as `true` or not -- feel free to discard if you disagree. But this should still also prevent the potential panic.

```suggestion
	if f.value == nil {
		return o.value == nil
```

bflad (approved):
Looks good to me 🚀

types/basetypes/float64_value.go
austinvalle: I was unsure as well, but I think if I have to choose, I'd rather do what you proposed. ⚡

bflad:
Oh wow, surprised that `(*big.Float).Cmp()` will panic when passed a `nil` value -- never mind about that I guess.

types/basetypes/float64_value.go
austinvalle: Sounds good, added doc note in 8672e73 ⚡

bflad (approved):

PR #824, opened 2023-08-17
website: Use hcl code language fences instead of terraform/tf

bflad:
The developer website rendering currently only supports `hcl` -- discussing with the maintainers about either adding a content check to suggest `hcl` code language instead or implementing `terraform` code language support similar to linguist and other renderers that support it already.


austinvalle (approved):
Quick look at the codebase, we have some straggling ```tf code fences as well:

-[website/docs/plugin/framework/providers/index.mdx](https://github.com/hashicorp/terraform-plugin-framework/blob/9d461607c2590a0d321c9462e176c8e071b66ba5/website/docs/plugin/framework/providers/index.mdx#L211)
- [website/docs/plugin/framework/handling-data/terraform-concepts.mdx](https://github.com/hashicorp/terraform-plugin-framework/blob/9d461607c2590a0d321c9462e176c8e071b66ba5/website/docs/plugin/framework/handling-data/terraform-concepts.mdx#L99)

PR #830, opened 2023-08-28
website: Move acceptance testing id requirement to troubleshooting section

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-testing/issues/84

terraform-plugin-testing version 1.5.0 and later will no longer require managed resources and data resources to implement the `id` attribute.


austinvalle (approved):
🚀

PR #834, opened 2023-09-05
Update mux documentation

akinross:
While making use of the examples for muxing noticed small adjustments I had to make in order to get examples to work.

1. providerserver.NewProtocol6() should get `provider.Provider` an not get a `func() provider.Provider` passed
2. ProtoV6ProviderFactories requires tfprotov6.ProviderServer not tfprotov5.ProviderServer


bflad (approved):
Thank you for these fixes, @akinross 🚀

PR #841, opened 2023-09-14
Correct Site Doc Links To "Write State" Page

tgoodsell-tempus:
Fixes locations where the incorrect reference `terraform/plugin/framework/handing-data/writing-state` is used, replaced with the correct `terraform/plugin/framework/handling-data/writing-state`.


austinvalle (approved):
Much appreciated @tgoodsell-tempus ! 🚀

PR #852, opened 2023-10-06
internal/fwserver - Fix `DataSource/Resource Type Not Found` errors when `GetProviderSchema` is not called

### LINKED ISSUE #853: `Data Source Type Not Found` or `Resource Type Not Found` when using Terraform 1.6
This is a Framework bug affecting users of Terraform 1.6 using providers built using Plugin Framework 1.4.0. The root cause is the recent caching functionality being enabled in #828.

### Module version

```
github.com/hashicorp/terraform-plugin-framework v1.4.0
```

Any resource/data source schema will have this problem, sample:
```go
resp.Schema = schema.Schema{
		"configurable_attribute": schema.StringAttribute{
			MarkdownDescription: "Example configurable attribute",
			Required:            true,
		},
	},
}
```

```hcl
resource "examplecloud_thing" "this" {
  configurable_attribute = "test"
}

data "examplecloud_thing" "this" {
  configurable_attribute = "test"
}
```

### Expected Behavior
Resource/Data Source type is found and RPCs correctly validate/plan/apply

### Actual Behavior

An error diagnostic is returned:

```shell
 $ terraform1.6 apply -auto-approve
╷
│ Error: Resource Type Not Found
│ 
│   with examplecloud_thing.this,
│   on resource.tf line 8, in resource "examplecloud_thing" "this":
│    8: resource "examplecloud_thing" "this" {
│ 
│ No resource type named "examplecloud_thing" was found in the provider.
╵
╷
│ Error: Data Source Type Not Found
│ 
│   with data.examplecloud_thing.this,
│   on resource.tf line 12, in data "examplecloud_thing" "this":
│   12: data "examplecloud_thing" "this" {
│ 
│ No data source type named "examplecloud_thing" was found in the provider.
╵
```

- For practitioners, they can downgrade to use Terraform `1.5.x`

- For provider developers, they can downgrade `terraform-plugin-framework` to `v1.3.5` AND `terraform-plugin-go` to `v0.18.0`.
   - If using `terraform-plugin-mux`, downgrade to `v0.11.2` to avoid build errors from `terraform-plugin-go`
   - If using `terraform-plugin-sdk/v2`, downgrade to `v2.28.0` to avoid build errors from `terraform-plugin-go`
#### Example of build error if you don't downgrade `terraform-plugin-go` and co.
```bash
# github.com/hashicorp/terraform-plugin-framework/internal/proto5server
../../go/pkg/mod/github.com/hashicorp/terraform-plugin-framework@v1.3.5/internal/proto5server/serve.go:14:34: cannot use &Server{} (value of type *Server) as tfprotov5.ProviderServer value in variable declaration: *Server does not implement tfprotov5.ProviderServer (missing method GetMetadata)
# github.com/hashicorp/terraform-plugin-framework/internal/proto6server
../../go/pkg/mod/github.com/hashicorp/terraform-plugin-framework@v1.3.5/internal/proto6server/serve.go:14:34: cannot use &Server{} (value of type *Server) as tfprotov6.ProviderServer value in variable declaration: *Server does not implement tfprotov6.ProviderServer (missing method GetMetadata)
```

### References

- https://github.com/hashicorp/terraform/issues/33990
- https://github.com/hashicorp/terraform/issues/34001

austinvalle:
Closes #853 

Terraform 1.6 no longer calls `GetProviderSchema` before any of the other RPCs, which currently populates `s.providerTypeName`. Without this populated, data sources + resources are added to their respective caches without the provider prefix, i.e. `_thing` instead of the correct `examplecloud_thing`, resulting in a confusing `Not Found` error.

### Notes

Unfortunately, this can't be recreated easily in CI because `terraform-plugin-testing` manages the plugin processes similar to [debug mode](https://developer.hashicorp.com/terraform/plugin/debugging), so the plugin keeps the provider running, whereas terraform core does not. Therefore, during a test run, `s.providerTypeName` happens to be populated due to a prior `GetProviderSchema` call populating it.

See: https://developer.hashicorp.com/terraform/plugin/debugging#debugging-caveats

### DIFF
--- a/.changes/unreleased/BUG FIXES-20231006-163235.yaml
+++ b/.changes/unreleased/BUG FIXES-20231006-163235.yaml
@@ -1,6 +1,0 @@
-kind: BUG FIXES
-body: 'internal/fwserver: Fix `Data Source Type Not Found` errors when `GetProviderSchema`
-  is not called'
-time: 2023-10-06T16:32:35.47541-04:00
-custom:
-  Issue: "853"
--- a/.changes/unreleased/BUG FIXES-20231006-163235.yaml
+++ b/.changes/unreleased/BUG FIXES-20231006-163235.yaml
@@ -1,0 +1,6 @@
+kind: BUG FIXES
+body: 'internal/fwserver: Fix `Data Source Type Not Found` errors when `GetProviderSchema`
+  is not called'
+time: 2023-10-06T16:32:35.47541-04:00
+custom:
+  Issue: "853"
--- a/internal/fwserver/server.go
+++ b/internal/fwserver/server.go
@@ -87,6 +87,10 @@ type Server struct {
 	// implemented the Metadata method. Access this field with the Provider.ProviderTypeName() method.
 	providerTypeName string
 
+	// providerTypeNameMutex is a mutex to protect concurrent providerTypeName
+	// access from race conditions.
+	providerTypeNameMutex sync.Mutex
+
 	// resourceSchemas is the cached Resource Schemas for RPCs that need to
 	// convert configuration data from the protocol. If not found, it will be
 	// fetched from the ResourceType.GetSchema() method.
@@ -288,6 +292,10 @@ func (s *Server) DataSourceSchemas(ctx context.Context) (map[string]fwschema.Sch
 
 // ProviderTypeName returns the TypeName associated with the Provider. The TypeName is cached on first use.
 func (s *Server) ProviderTypeName(ctx context.Context) string {
+	logging.FrameworkTrace(ctx, "Checking ProviderTypeName lock")
+	s.providerTypeNameMutex.Lock()
+	defer s.providerTypeNameMutex.Unlock()
+
 	if s.providerTypeName != "" {
 		return s.providerTypeName
 	}
--- a/internal/fwserver/server.go
+++ b/internal/fwserver/server.go
@@ -83,8 +83,8 @@ type Server struct {
 	// access from race conditions.
 	providerMetaSchemaMutex sync.Mutex
 
-	// providerTypeName is the type name of the provider, if the provider
-	// implemented the Metadata method.
+	// providerTypeName is the cached type name of the provider, if the provider
+	// implemented the Metadata method. Access this field with the Provider.ProviderTypeName() method.
 	providerTypeName string
 
 	// resourceSchemas is the cached Resource Schemas for RPCs that need to
@@ -140,6 +140,7 @@ func (s *Server) DataSourceFuncs(ctx context.Context) (map[string]func() datasou
 		return s.dataSourceFuncs, s.dataSourceTypesDiags
 	}
 
+	providerTypeName := s.ProviderTypeName(ctx)
 	s.dataSourceFuncs = make(map[string]func() datasource.DataSource)
 
 	logging.FrameworkTrace(ctx, "Calling provider defined Provider DataSources")
@@ -150,7 +151,7 @@ func (s *Server) DataSourceFuncs(ctx context.Context) (map[string]func() datasou
 		dataSource := dataSourceFunc()
 
 		dataSourceTypeNameReq := datasource.MetadataRequest{
-			ProviderTypeName: s.providerTypeName,
+			ProviderTypeName: providerTypeName,
 		}
 		dataSourceTypeNameResp := datasource.MetadataResponse{}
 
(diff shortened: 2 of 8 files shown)


internal/fwserver/server.go
```diff
@@ -285,6 +286,24 @@ func (s *Server) DataSourceSchemas(ctx context.Context) (map[string]fwschema.Sch
 	return dataSourceSchemas, diags
 }
 
+// ProviderTypeName returns the TypeName associated with the Provider. The TypeName is cached on first use.
+func (s *Server) ProviderTypeName(ctx context.Context) string {
```
bendbennett, line 290: Is it worth having a mutex here?

```go
	s.providerTypeNameMutex.Lock()
	defer s.providerTypeNameMutex.Unlock()
```

internal/fwserver/server_getmetadata.go
```diff
@@ -44,15 +42,6 @@ func (s *Server) GetMetadata(ctx context.Context, req *GetMetadataRequest, resp
 	resp.Resources = []ResourceMetadata{}
 	resp.ServerCapabilities = s.ServerCapabilities()
 
-	metadataReq := provider.MetadataRequest{}
```
bendbennett, line 47: Should we call `s.ProviderTypeName()` to set `providerTypeName` in order to preserve original behaviour?

internal/fwserver/server_getproviderschema.go
```diff
@@ -31,15 +29,6 @@ type GetProviderSchemaResponse struct {
 func (s *Server) GetProviderSchema(ctx context.Context, req *GetProviderSchemaRequest, resp *GetProviderSchemaResponse) {
 	resp.ServerCapabilities = s.ServerCapabilities()
 
-	metadataReq := provider.MetadataRequest{}
```
bendbennett, line 34: Should we call `s.ProviderTypeName()` to set `providerTypeName` in order to preserve original behaviour?

bendbennett (commented):
Looks good.
Just a couple of minor comments.

internal/fwserver/server_getmetadata.go
austinvalle: I also wondered this, it felt weird to make the call here to populate a cache used somewhere else. If we do want to preserve this, we could maybe change the naming up:

`s.SetProviderTypeName()`

But I went for the "lazy-load" approach 😆

internal/fwserver/server.go
austinvalle: Yeah we will have to introduce that I believe to avoid a race condition 👍🏻

austinvalle: Added in f147a4e ⚡

internal/fwserver/server_getmetadata.go
austinvalle: Just a clarification, the overall behavior is still preserved as there already exists a test for this: https://github.com/hashicorp/terraform-plugin-framework/blob/f147a4e18b39cfbee34b1d95adb0b5cfd12ec0a8/internal/fwserver/server_getmetadata_test.go#L156

The implementation detail has just been moved closer to where it's needed for the other RPCs to consume as well 👍🏻 

An alternative approach could be to set `s.ProviderTypeName()` at the top of every RPC

austinvalle:
I'm not sure the best way to add a CI test for this functionality (due to the way this bug is exposed), any suggestions for improving the test coverage of this is appreciated.

I'm just been running manual tests with a sandbox provider comparing `v1.4.0` to the latest commit of this repo 8a47174:
```shell
 $ go list -m github.com/hashicorp/terraform-plugin-framework/...
github.com/hashicorp/terraform-plugin-framework v1.4.0

 $ terraform1.6 -chdir=examples/resources apply -auto-approve    
╷
│ Warning: Provider development overrides are in effect
│ 
│ The following provider development overrides are set in the CLI configuration:
│  - austinvalle/sandbox in /Users/austin.valle/go/bin
│ 
│ The behavior may therefore not match any released version of the provider and applying changes may
│ cause the state to become incompatible with published releases.
╵
╷
│ Error: Resource Type Not Found
│ 
│   with examplecloud_thing.this,
│   on resource.tf line 8, in resource "examplecloud_thing" "this":
│    8: resource "examplecloud_thing" "this" {
│ 
│ No resource type named "examplecloud_thing" was found in the provider.
╵
╷
│ Error: Data Source Type Not Found
│ 
│   with data.examplecloud_thing.this,
│   on resource.tf line 12, in data "examplecloud_thing" "this":
│   12: data "examplecloud_thing" "this" {
│ 
│ No data source type named "examplecloud_thing" was found in the provider.
╵

 $ go get -u github.com/hashicorp/terraform-plugin-framework@8a471749f9833cc8e7bb0355d436a23782e1c9a4
go: upgraded github.com/hashicorp/terraform-plugin-framework v1.4.0 => v1.4.1-0.20231006203353-8a471749f983

 $ go mod tidy

 $ go install .

 $ terraform1.6 -chdir=examples/resources apply -auto-approve 
╷
│ Warning: Provider development overrides are in effect
│ 
│ The following provider development overrides are set in the CLI configuration:
│  - austinvalle/sandbox in /Users/austin.valle/go/bin
│ 
│ The behavior may therefore not match any released version of the provider and applying changes may
│ cause the state to become incompatible with published releases.
╵
data.examplecloud_thing.this: Reading...
examplecloud_thing.this: Refreshing state...
data.examplecloud_thing.this: Read complete after 0s

No changes. Your infrastructure matches the configuration.

Terraform has compared your real infrastructure against your configuration and found no differences, so
no changes are needed.

Apply complete! Resources: 0 added, 0 changed, 0 destroyed.
```

.changes/unreleased/BUG FIXES-20231006-163235.yaml
```diff
@@ -0,0 +1,6 @@
+kind: BUG FIXES
+body: 'internal/fwserver: Fix `Data Source Type Not Found` errors when `GetProviderSchema`
```
bflad, line 2: It might be good to call this out from the perspective of when the problem can occur and the exported "providerserver" package can act as a catch-all category for internal provider server bits.

```markdown
providerserver: Prevented `Data Source Type Not Found` and `Resource Type Not Found` errors with Terraform 1.6 and later
```

bflad (approved):
Looks good to me 🚀

.changes/unreleased/BUG FIXES-20231006-163235.yaml
austinvalle: That message sounds good to me 👍🏻 - thanks for the suggestion

PR #864, opened 2023-10-23
resource: Prevent private state data loss on Delete method error

### LINKED ISSUE #863: Private State is lost after a failed Deletion
### Module version

```
v1.3.4
```

```go
func (r *MyResource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
	result, err, d := performOperation(...)
	resp.Diagnostics.Append(d...)
	if err != nil {
	        return
	}

	d = applyState(&resp.State, result)
	resp.Diagnostics.Append(d...)
	        return
	}

	d = resp.Private.SetKey(ctx, createResultKey, result)
	resp.Diagnostics.Append(d...)
}

func (r *MyResource) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
        // Do stuff, has access to 'createResultKey' via req.Private.GetKey
}

func (r *MyResource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
        // Do stuff, has access to 'createResultKey' via req.Private.GetKey
}

func (r *MyResource) Delete(ctx context.Context, req resource.DeleteRequest, resp *resource.DeleteResponse) {
	ctx = tflog.SetField(ctx, rpcField, "delete")
	ctx = tflog.SetField(ctx, actionResourceNameField, r.name)
	r.performLinkOperationFromScratch(ctx, r.delete, req.Private.GetKey, req.Private.SetKey, req.State, &resp.State, &resp.Diagnostics)
		return
	}
	tflog.Info(ctx, "resource deleted")
}
```

```hcl
...
```

bflad:
Closes #863

The previous implementation assumed that Terraform core would preserve private state similar to "regular" state when a provider returned an error diagnostic while destroying a resource. The core implementation instead always uses the response data, which the framework was previously always discarding. This change corrects the errant design while also enabling provider logic to potentially update the private state in this situation.

### DIFF
--- a/.changes/unreleased/BUG FIXES-20231023-165712.yaml
+++ b/.changes/unreleased/BUG FIXES-20231023-165712.yaml
@@ -1,0 +1,6 @@
+kind: BUG FIXES
+body: 'resource: Add `Private` field to `DeleteResource` type, which was missing to
+  allow provider logic to update private state on errors'
+time: 2023-10-23T16:57:12.447739-04:00
+custom:
+  Issue: "863"
--- a/.changes/unreleased/BUG FIXES-20231023-165814.yaml
+++ b/.changes/unreleased/BUG FIXES-20231023-165814.yaml
@@ -1,0 +1,6 @@
+kind: BUG FIXES
+body: 'resource: Prevented private state data loss if resource destruction returned
+  an error'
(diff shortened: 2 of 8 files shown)


bendbennett (approved):
LGTM 🚀

PR #870, opened 2023-11-02
types/basetypes: Add Tuple type support

austinvalle:
Ref: #54 

This PR introduces `basetypes.TupleType`, `basetypes.TupleValue`, and tuple helper functions/aliases in the `types` package. This PR **does not** introduce any accompanying `TupleAttribute` implementations, so currently there is no way to describe this schema to Terraform core.

Eventually when #147 is implemented, the `TupleType` will be necessary to fully handle dynamic attributes, as then Terraform core could possible send a [tuple](https://developer.hashicorp.com/terraform/language/expressions/type-constraints#tuple) to Plugin Framework.

### Notes
- I updated the reflection rules to create an explicit message describing that reflection from Go slices into `types.Tuple` is not supported. Also added tests around the code that was refactored relating to sets/lists
   - There weren't explicit tests for `Into` and `FromValue`, so I created table tests for them and added tests for just what I refactored. We can continually update the tests as we make changes to `internal/reflect`
- I wasn't 100% sure if we needed to implement `(TupleType).ApplyTerraform5AttributePathStep` for dynamic attributes, so I can update the implementation to return an error if that's not needed.

### DIFF
--- a/.changes/unreleased/FEATURES-20231102-160132.yaml
+++ b/.changes/unreleased/FEATURES-20231102-160132.yaml
@@ -1,5 +1,5 @@
-kind: FEATURES
-body: 'types/basetypes: Added TupleType and TupleValue implementations'
+kind: ENHANCEMENTS
+body: 'types/basetypes: Added `TupleType` and `TupleValue` implementations, which are only necessary for dynamic value handling'
 time: 2023-11-02T16:01:32.166746-04:00
 custom:
   Issue: "870"
--- a/.changes/unreleased/FEATURES-20231102-160132.yaml
+++ b/.changes/unreleased/FEATURES-20231102-160132.yaml
@@ -1,0 +1,5 @@
+kind: FEATURES
+body: 'types/basetypes: Added TupleType and TupleValue implementations'
+time: 2023-11-02T16:01:32.166746-04:00
+custom:
+  Issue: "870"
--- a/internal/reflect/into_test.go
+++ b/internal/reflect/into_test.go
@@ -21,9 +21,6 @@ import (
 func TestInto_Slices(t *testing.T) {
 	t.Parallel()
 
-	// Just using this to get the reflected type for the odd diagnostics :-)
-	var stringSlice []string
-
 	testCases := map[string]struct {
 		typ           attr.Type
 		value         tftypes.Value
@@ -73,7 +70,7 @@ func TestInto_Slices(t *testing.T) {
 							tftypes.NewValue(tftypes.String, "hello"),
 							tftypes.NewValue(tftypes.String, "world"),
 						}),
-						TargetType: reflect.TypeOf(stringSlice),
+						TargetType: reflect.TypeOf([]string{}),
 						Err:        errors.New("cannot reflect tftypes.Tuple[tftypes.String, tftypes.String] using type information provided by basetypes.TupleType, reflection support is currently not implemented for tuples"),
 					},
 				),
@@ -89,7 +86,7 @@ func TestInto_Slices(t *testing.T) {
 					path.Empty(),
 					refl.DiagIntoIncompatibleType{
 						Val:        tftypes.NewValue(tftypes.String, "hello"),
-						TargetType: reflect.TypeOf(stringSlice),
+						TargetType: reflect.TypeOf([]string{}),
 						Err:        errors.New("can't unmarshal tftypes.String into *[]tftypes.Value expected []tftypes.Value"),
 					},
 				),
--- a/internal/reflect/into_test.go
+++ b/internal/reflect/into_test.go
@@ -1,0 +1,119 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package reflect_test
+
+import (
+	"context"
+	"errors"
+	"reflect"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	refl "github.com/hashicorp/terraform-plugin-framework/internal/reflect"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+)
+
+func TestInto_Slices(t *testing.T) {
+	t.Parallel()
+
+	// Just using this to get the reflected type for the odd diagnostics :-)
+	var stringSlice []string
+
+	testCases := map[string]struct {
+		typ           attr.Type
+		value         tftypes.Value
+		target        []string
+		expected      []string
+		expectedDiags diag.Diagnostics
+	}{
+		"list-to-go-slice": {
+			typ: types.ListType{ElemType: types.StringType},
+			value: tftypes.NewValue(tftypes.List{
+				ElementType: tftypes.String,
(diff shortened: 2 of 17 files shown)


austinvalle:
For some reason I forgot that the documentation existed in this repo 😖, updated the PR with the tuple docs ⚡

.changes/unreleased/FEATURES-20231102-160132.yaml
```diff
@@ -0,0 +1,5 @@
+kind: FEATURES
+body: 'types/basetypes: Added TupleType and TupleValue implementations'
```
bflad, line 2: Nit: I wonder if we should potentially call this out as an enhancement rather than a feature and describe it a little more, since ultimately we don't anticipate provider developers to interact with this directly too much.

```suggestion
kind: ENHANCEMENTS
body: 'types/basetypes: Added `TupleType` and `TupleValue` implementations, which are only necessary for dynamic value handling'
```

internal/reflect/into_test.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+)
+
+func TestInto_Slices(t *testing.T) {
+	t.Parallel()
+
+	// Just using this to get the reflected type for the odd diagnostics :-)
+	var stringSlice []string
```
bflad, line 25: Pretty sure you can skip this via `reflect.TypeOf([]string{})`

https://go.dev/play/p/tfisl0oH-0s

Less mental redirection for the win!

internal/reflect/outof_test.go
```diff
@@ -0,0 +1,112 @@
...
+	refl "github.com/hashicorp/terraform-plugin-framework/internal/reflect"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+func TestFromValue_go_types(t *testing.T) {
+	t.Parallel()
+
+	// Used for testing nil -> null attr.Value
+	var nilSlice []string
```
bflad, line 22: Pretty sure you can use `([]string)(nil)` or `new([]string)` directly for these 👍

types/basetypes/tuple_value.go
```diff
@@ -0,0 +1,254 @@
...
+// Elements returns a copy of the ordered list of known values for the Tuple.
+func (v TupleValue) Elements() []attr.Value {
+	// Ensure callers cannot mutate the internal elements
+	result := make([]attr.Value, 0, len(v.elements))
+	result = append(result, v.elements...)
+
+	return result
+}
+
+// ElementTypes returns a copy of the ordered list of element types for the Tuple.
```
bflad, line 134: To match the current implementation.

```suggestion
// ElementTypes returns the ordered list of element types for the Tuple.
```

website/docs/plugin/framework/handling-data/types/index.mdx
```diff
@@ -60,3 +60,15 @@ This type is associated with:
 | Type | Use Case |
 |----------------|----------|
 | [Object](/terraform/plugin/framework/handling-data/types/object) | Mapping of explicit attribute names to values |
+
+### Tuple Type
+
+Type that defines an ordered collection of elements where each element has it's own type.
+
+<Note>
+This type intentionally includes less functionality than other types in the type system as it has limited real world application and therefore is not exposed to provider developers.
```
bflad, line 69: I think we can go ahead and make this forward statement now so we don't have to update it later:

```suggestion
This type intentionally includes less functionality than other types in the type system as it has limited real world application and therefore is not exposed to provider developers except when working with dynamic values.
```

website/docs/plugin/framework/handling-data/types/list.mdx
```diff
@@ -83,7 +83,7 @@ Call one of the following to create a `types.List` value:
...
 * [`types.ListValueFrom(context.Context, attr.Type, any) (types.List, diag.Diagnostics)`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/types#ListValueFrom): A known value with the given element type and values. This can convert the source data from standard Go types into framework types as noted in the documentation for each element type, such as giving `[]*string` for a `types.List` of `types.String`.
-* [`types.ListValueMust(map[string]attr.Type, map[string]attr.Value) types.List`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/types#ListValueMust): A known value with the given element type and values. Any diagnostics are converted to a runtime panic. This is recommended only for testing or exhaustively tested logic.
+* [`types.ListValueMust(attr.Type, []attr.Value) types.List`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/types#ListValueMust): A known value with the given element type and values. Any diagnostics are converted to a runtime panic. This is recommended only for testing or exhaustively tested logic.
```
bflad, line 86: ❤️

website/docs/plugin/framework/handling-data/types/tuple.mdx
```diff
@@ -0,0 +1,53 @@
+---
+page_title: 'Plugin Development - Framework: Tuple Type'
+description: >-
+  Learn the tuple value type in the provider development framework.
+---
+
+<Note>
+The tuple type intentionally includes less functionality than other types in the type system as it has limited real world application and therefore is not exposed to provider developers.
```
bflad, line 8:
```suggestion
The tuple type intentionally includes less functionality than other types in the type system as it has limited real world application and therefore is not exposed to provider developers except when working with dynamic values.
```

website/docs/plugin/framework/handling-data/types/tuple.mdx
```diff
@@ -0,0 +1,53 @@
...
+
+# Tuple Type
+
+Tuple types store an ordered collection of elements where each element has it's own type. Values must have **exactly** the same number of elements (no more and no fewer), and the value in each position must match the specified type for that position.
+
+The tuple type is used to express Terraform's [tuple type constraint](/terraform/language/expressions/type-constraints#tuple).
+
+## Schema Definitions
+
+The tuple type is currently not supported in schema definitions of provider, data sources, or managed resources.
```
bflad, line 19:
```suggestion
The tuple type is not supported in schema definitions of provider, data sources, or managed resources as it has limited real world application.
```

bflad (approved):
This looks great to me 🚀 Just a few minor suggestions.

> I wasn't 100% sure if we needed to implement (TupleType).ApplyTerraform5AttributePathStep for dynamic attributes, so I can update the implementation to return an error if that's not needed.

Adding the correct implementation now (which it looks like you did), rather than potentially discovering that as an issue later, is definitely appreciated!

.changes/unreleased/FEATURES-20231102-160132.yaml
austinvalle: Yeah this makes sense, I was between the two 😆 - ⚡

internal/reflect/into_test.go
austinvalle: ⚡

internal/reflect/outof_test.go
austinvalle: Ah! Thank you for this, I'm not sure why my brain couldn't come up with this lol

website/docs/plugin/framework/handling-data/types/list.mdx
austinvalle: I definitely did NOT find this because I was copy pasting the docs 😅

PR #877, opened 2023-11-21
docs: Further clarify SDK migration testing of data sources

### LINKED ISSUE #875: Testing Migration (from SDKv2) of Datasources does not fail when fields are missing
### Module version

```
v1.4.2
```

We are currently in the process of migration our provider to the plugin-framework. To test that we included every field in the new data source implementation, we wrote a test similar to the docs at: https://developer.hashicorp.com/terraform/plugin/framework/migrating/testing#example

```go
func TestAccHcloudDataSourceDatacenters_UpgradePluginFramework(t *testing.T) {
	tmplMan := testtemplate.Manager{}

	datacentersD := &datacenter.DDataList{}
	datacentersD.SetRName("ds")

	resource.ParallelTest(t, resource.TestCase{
		PreCheck: e2etests.PreCheck(t),
		Steps: []resource.TestStep{
			{
				ExternalProviders: map[string]resource.ExternalProvider{
					"hcloud": {
                                                # Last version before migrating this specific data source
						VersionConstraint: "1.44.1",
						Source:            "hetznercloud/hcloud",
					}
				},

				Config: tmplMan.Render(t,
					"testdata/d/hcloud_datacenters", datacentersD,
				),
			},
			{
				ProtoV6ProviderFactories: e2etests.ProtoV6ProviderFactories(),

				Config: tmplMan.Render(t,
					"testdata/d/hcloud_datacenters", datacentersD,
				),

				PlanOnly: true,
			},
		},
	})
}
```

Both test steps  use the same configuration (templated):

```hcl
data "hcloud_datacenters" "ds" {}
```

### Expected Behavior

We would expect any missing or different fields between the two steps to cause the test to fail. We need this behaviour to test that we added all fields with the same values as the previous implementation. Especially as this is the documented/recommended way to test the migration.

### Actual Behavior

The test does not fail if we remove a field in the code in the new plugin-framework implementation.

We tested this for _Resources_ and that works as expected, only _Datasources_ are allowed to have fields missing.

### Steps to Reproduce

Not sure if there is an easy way to reproduce this. If wanted, I can push a branch to our repository that demonstrates the issue.

### References

We noticed this while testing https://github.com/hetznercloud/terraform-provider-hcloud/pull/779

We ended up using a [`null_resource` that gets the datasource `jsonencoded`](https://github.com/hetznercloud/terraform-provider-hcloud/pull/779/commits/ba63d62b6b6d13ea0452163a8b1624261d033e92) in a trigger to work around the issue. Would be nice if we can drop this in the future.

bflad:
Closes #875

The purpose of this additional documentation is two-fold:

- Showing how to perform migration testing of full data source data, using the `terraform_data` managed resource
- Recommending that developers should instead prefer individual attribute value testing for data sources

### DIFF
--- a/website/docs/plugin/framework/migrating/testing.mdx
+++ b/website/docs/plugin/framework/migrating/testing.mdx
@@ -32,7 +32,7 @@ the Framework.
 
 ### Example
 
-This example shows how you can use external providers to generate a state file with a previous version of the provider
+These examples show how you can use external providers to generate state with a previous version of the provider
 and then verify that there are no planned changes after migrating to the Framework.
 
 - The first `TestStep` uses `ExternalProviders` to cause `terraform apply` to execute with a previous version of the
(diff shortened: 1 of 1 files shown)


bendbennett (approved):
LGTM 🚀

PR #899, opened 2023-12-28
website: Use ConfigPlanChecks instead of PlanOnly in migration testing

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-testing/issues/256


bendbennett (approved):
LGTM 🚀

Issue #902, opened 2024-01-08
Consider Explicit Error Diagnostics When ReadResource/UpgradeResourceState Response State Includes Unknown Value


bflad:
### Module version

```
v1.4.1
```

### Relevant provider source code

`ReadResource` case:

```go
package provider

import (
	"context"

	"github.com/hashicorp/terraform-plugin-framework/resource"
	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
	"github.com/hashicorp/terraform-plugin-framework/types"
)

var (
	_ resource.Resource = &Testing262Resource{}
)

func NewTesting262Resource() resource.Resource {
	return &Testing262Resource{}
}

type Testing262Resource struct{}

func (r *Testing262Resource) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
	resp.TypeName = req.ProviderTypeName + "_testing262"
}

func (r *Testing262Resource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
	resp.Schema = schema.Schema{
		Attributes: map[string]schema.Attribute{
			"required_string": schema.StringAttribute{
				Required: true,
			},
		},
	}
}

func (r Testing262Resource) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
	var data Testing262ResourceModel

	resp.Diagnostics.Append(req.Plan.Get(ctx, &data)...)

	if resp.Diagnostics.HasError() {
		return
	}

	resp.Diagnostics.Append(resp.State.Set(ctx, data)...)
}

func (r Testing262Resource) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
	var data Testing262ResourceModel

	resp.Diagnostics.Append(req.State.Get(ctx, &data)...)

	if resp.Diagnostics.HasError() {
		return
	}

	// intentionally invalid implementation for example code
	data.RequiredString = types.StringUnknown()

	resp.Diagnostics.Append(resp.State.Set(ctx, data)...)
}

func (r Testing262Resource) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
	var data Testing262ResourceModel

	resp.Diagnostics.Append(req.Plan.Get(ctx, &data)...)

	if resp.Diagnostics.HasError() {
		return
	}

	resp.Diagnostics.Append(resp.State.Set(ctx, data)...)
}

[...]

PR #903, opened 2024-01-09
website: Document managed resource Read and UpgradeState unknown value errors

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/902
Reference: https://github.com/hashicorp/terraform/issues/34502
Reference: https://github.com/hashicorp/terraform/issues/34503

Unknown values are never valid in resource state and while Terraform nor the framework actually raise these errors right now, one or both will in the near future. This documents the errant implementation detail for developers in the meantime since it otherwise causes confusing Terraform behaviors for practitioners.


bendbennett (approved):
LGTM 🚀

PR #911, opened 2024-01-26
privatestate: Support nil or zero-length value in SetKey to remove key

### LINKED ISSUE #910: Proposal: Allow removing key from `Private` data
### Module version

```
github.com/hashicorp/terraform-plugin-framework v1.5.0
```

### Use-cases

Resource `Private` data may sometimes no longer be needed. There is currently no way to remove it.

### Attempted Solutions

Setting a key to `nil` or `""` causes an error, such as

> The value being supplied for key "aws_lb_listener.importing" is invalid.
> Please verify that the value is valid JSON.

### Proposal

Add a corresponding function `RemoveKey` to `*privatestate.ProviderData`.

bflad:
Closes #910

This change ensures that developers can fully remove private state keys by setting the value to `nil` or `[]byte{}`. Attempting to set the value to either value would previously generate an error. The closest available workaround was to set keys to the literal JSON `null` and handle that value during unmarshaling, rather than purely checking existence via `GetKey()` returning `nil`.

### DIFF
--- a/.changes/unreleased/ENHANCEMENTS-20240126-073810.yaml
+++ b/.changes/unreleased/ENHANCEMENTS-20240126-073810.yaml
@@ -1,6 +1,6 @@
 kind: ENHANCEMENTS
-body: 'privatestate: Added `nil` or zero-length value in `SetKey()` method to fully
-  remove key'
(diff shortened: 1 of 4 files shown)


bendbennett (approved):
LGTM 🚀

PR #917, opened 2024-02-01
resource: Initial MoveResourceState RPC support

bflad:
This change adds initial support for the `MoveResourceState` RPC to the framework, including:

* Adding the framework shared server implementation for the new RPC
* Adding the protocol version 5 and 6 server implementations for the new RPC
* Adding the type conversion logic for the framework types to/from the protocol types
* Exposing a new `resource.ResourceWithMoveState` interface that providers can implement to support the new RPC
* Adding a website documentation page for the new functionality

A state move using the new RPC occurs when a Terraform 1.8 and later configuration includes a `moved` configuration block such as the following:

```terraform
moved {
    from = "examplecloud_source.XXX"
    to   = "examplecloud_target.XXX"
}
```

There are no restrictions on the source resource types, but target resource types must opt into support to prevent data loss. Target resources can support moves from multiple, differing source resources, so the framework implementation is exposed as an ordered list to developers.

The framework implementation for the new RPC is as follows:

* If no state move support is defined for the resource, the framework returns an error diagnostic.
* If state move support is defined for the resource, each provider-defined state move implementation is called until one responds with error diagnostics or state data.
* If all provider-defined state move implementations return without error diagnostics and state data, the framework returns an error diagnostic.

The protocol server unit testing shows the end-to-end handling of the new RPC, including the provider-defined resource implementations, the type conversion logic, and the framework shared server implementation. The exposed `resource.ResourceWithMoveState` implementation is intentionally similar to the existing `resource.ResourceWithUpgradeState` handling, both in internal details and the exposed API. This is to provide a smoother experience for provider developers familiar with the other functionality and maintainers of the framework.

This initial implementation exposes a lot of the request/response handling as-is, however there is likely potential for introducing native helper functionality for developers, such as implementing one to simplify "aliasing"/"renaming" an existing `resource.Resource` within the same provider codebase.

### DIFF
--- a/internal/fwserver/server_moveresourcestate.go
+++ b/internal/fwserver/server_moveresourcestate.go
@@ -163,6 +163,14 @@ func (s *Server) MoveResourceState(ctx context.Context, req *MoveResourceStateRe
 			// since both the developer and calling practitioner cannot avoid
 			// the situation. Instead, developers will still have a nil
 			// SourceState and they can investigate any error as logged here.
+			//
+			// It is also important to note that the error generally only occurs
+			// if the source schema declared incompatible types. The
+			// IgnoreUndefinedAttributes option above can cause the error to be
+			// nil and the SourceState to be populated with null values. It is
+			// always recommended for StateMover implementations to check the
+			// other request fields (SourceTypeName, SourceProviderAddress,
+			// etc.) instead of relying on SourceState to be populated or not.
 			if err != nil {
 				logging.FrameworkDebug(
 					ctx,
--- a/internal/fwserver/server_moveresourcestate.go
+++ b/internal/fwserver/server_moveresourcestate.go
@@ -1,0 +1,187 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package fwserver
+
+import (
+	"context"
+	"strconv"
+
+	"github.com/hashicorp/terraform-plugin-go/tfprotov6"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/internal/fwschema"
+	"github.com/hashicorp/terraform-plugin-framework/internal/logging"
+	"github.com/hashicorp/terraform-plugin-framework/internal/privatestate"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
+)
+
+// MoveResourceStateRequest is the framework server request for the
+// MoveResourceState RPC.
+type MoveResourceStateRequest struct {
+	SourcePrivate         *privatestate.Data
+	SourceProviderAddress string
+	SourceSchemaVersion   int64
+	SourceRawState        *tfprotov6.RawState
+	SourceTypeName        string
+
+	TargetResource       resource.Resource
+	TargetResourceSchema fwschema.Schema
+	TargetTypeName       string
+}
+
+// MoveResourceStateResponse is the framework server response for the
+// MoveResourceState RPC.
+type MoveResourceStateResponse struct {
+	Diagnostics   diag.Diagnostics
+	TargetPrivate *privatestate.Data
+	TargetState   *tfsdk.State
+}
+
+// MoveResourceState implements the framework server MoveResourceState RPC.
+func (s *Server) MoveResourceState(ctx context.Context, req *MoveResourceStateRequest, resp *MoveResourceStateResponse) {
+	if req == nil {
+		return
+	}
+
+	if req.SourceRawState == nil {
+		resp.Diagnostics.AddError(
+			"Missing Source Resource State",
+			"The source resource state was not provided to the provider for the MoveResourceState operation. "+
+				"This is always an issue in Terraform and should be reported to the Terraform maintainers.",
+		)
+
+		return
+	}
+
+	resourceWithMoveState, ok := req.TargetResource.(resource.ResourceWithMoveState)
+
+	if !ok {
+		resp.Diagnostics.AddError(
+			"Unable to Move Resource State",
+			"The target resource implementation does not include move resource state support. "+
+				"The resource implementation can be updated by the provider developers to include this support with the ResourceWithMoveState interface.\n\n"+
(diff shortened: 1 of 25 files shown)


internal/fwserver/server_moveresourcestate_test.go
austinvalle, line 1: Do we want/need to have tests for some of the error "fail-forward" conditions when there are multiple implementations? Like:
- StateMover 1 schema doesn't match source, StateMover 2 schema matches + returns target state == success
- StateMover 1 schema doesn't match source, StateMover 2 doesn't define schema + returns target state == success
- StateMover 1 schema matches + returns target state, StaveMover 2 schema matches + would return target state == first is called, second is not

resource/move_state.go
```diff
@@ -0,0 +1,105 @@
...
+	// copied automatically.
+	//
+	// This is advanced functionality for providers wanting to skip the full
+	// redeclaration of source schemas and instead use lower level handlers to
+	// transform data. A typical implementation for working with this data will
+	// call the Unmarshal() method.
+	//
+	// TODO: Create framework defined type that is not protocol specific.
+	// Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/340
+	SourceRawState *tfprotov6.RawState
```
austinvalle, line 49: I mentioned this in another comment, but this issue is closed now. Do we need to re-open it or remove this comment?

resource/move_state.go
```diff
@@ -0,0 +1,105 @@
...
+	SourceSchemaVersion int64
+
+	// SourceState is the source resource state if the [StateMover.SourceSchema]
+	// was set. When available, this allows for easier data handling such as
+	// calling Get() or GetAttribute().
+	//
+	// If this request matches the intended implementation, the implementation
+	// logic must set [MoveStateResponse.TargetState] as it is intentionally not
+	// copied automatically.
+	SourceState *tfsdk.State
```
austinvalle, line 64: I made a comment above about this field. If it's possible that the `StateMover` implementation may be called with this field as `nil` when `StateMover.SourceSchema` doesn't match, we may want to document that here.

resource/state_mover.go
```diff
@@ -0,0 +1,63 @@
...
+	// SourceSchema is an optional schema for the intended source resource state
+	// and schema version. While not required, setting this will populate
+	// [MoveStateRequest.SourceState] when possible similar to other [Resource]
+	// types. State conversion errors based on this schema are only logged at
+	// DEBUG level as there may be multiple [StateMover] implementations on the
+	// same target resource for differing source resources.
+	//
+	// If not set, source state data is only available in
+	// [MoveStateRequest.SourceRawState].
+	SourceSchema *schema.Schema
```
austinvalle, line 41: Should we mention explicitly that state conversion errors will not cause the `StateMover` function to be skipped? It's kind of implied but was something that came to my mind when reading the interface.

website/docs/plugin/framework/resources/state-move.mdx
```diff
@@ -0,0 +1,264 @@
...
+        },
+    }
+}
+```
+
+## Caveats
+
+Note these caveats when implementing the `MoveState` method:
+
+* An error is returned if the response state contains unknown values. Set all attributes to either null or known values in the response.
```
austinvalle, line 263: Curious q: Is this validation implemented on the Terraform core side?

website/docs/plugin/framework/resources/state-move.mdx
austinvalle, line 1: Docs look great ❤️

austinvalle (approved):
Looks great! I had a couple notes and questions but nothing major. :shipit:

internal/fwserver/server_moveresourcestate.go
```diff
@@ -0,0 +1,187 @@
...
+					"Error unmarshalling SourceRawState using the provided SourceSchema for source "+
+						req.SourceProviderAddress+" resource type "+
+						req.SourceTypeName+" with schema version "+
+						strconv.FormatInt(req.SourceSchemaVersion, 10)+". "+
+						"This is not a fatal error since resources can support multiple source resources which cause this type of error to be unavoidable, "+
+						"but due to this error the SourceState will not be populated for the implementation.",
+					map[string]any{
+						logging.KeyError: err,
+					},
+				)
```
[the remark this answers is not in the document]
bflad: I think you may have found out later in reading the implementation but yeah, the comment at the beginning of this error message conditional here could probably use a little more verbiage. This logic essentially will always run if `SourceSchema` is set but before its ever passed off to `StateMover` provider-defined logic which might "skip" based on the other request fields (source resource type, source provider address, etc.).

We _could_ technically support some upfront "filtering" via framework logic before it reaches here by adding `SourceTypeName`, etc. fields next to `SourceSchema` in `StateMover` that have some opinionated behaviors, but I was hesitant to immediately introduce that. 😄

internal/fwserver/server_moveresourcestate.go
```diff
@@ -0,0 +1,187 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
+)
+
+// MoveResourceStateRequest is the framework server request for the
+// MoveResourceState RPC.
+type MoveResourceStateRequest struct {
+	SourcePrivate         *privatestate.Data
+	SourceProviderAddress string
+	SourceSchemaVersion   int64
+	SourceRawState        *tfprotov6.RawState
```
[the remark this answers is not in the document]
bflad: Adding a comment here is an excellent suggestion, will do.

bflad (commented):
Will take care of these things shortly (likely tomorrow 😄 ).

internal/fwserver/server_moveresourcestate.go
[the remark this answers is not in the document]
austinvalle: > I think you may have found out later in reading the implementation

Ah yes! My main issue with reviewing PRs is not returning to previous in-progress comments lol.. 😅 

This all makes sense to me and the additional docs on `SourceState` I think closes the loop nicely with this log message 👍🏻

website/docs/plugin/framework/resources/state-move.mdx
bflad: Looks like this should be good! https://github.com/hashicorp/terraform/blob/48148eda5f5111fe4d8fdd28882472076227691a/internal/refactoring/cross_provider_move.go#L165-L173

Issue #914, opened 2024-01-30
Convert to number error when storing a value in float64 during execution of provider-defined function


bendbennett:
### Module version

```
v1.5.0
```

### Relevant provider source code

```go
var _ function.Function = Float64Function{}

func NewFloat64Function() function.Function {
	return &Float64Function{}
}

type Float64Function struct{}

func (f Float64Function) Metadata(ctx context.Context, req function.MetadataRequest, resp *function.MetadataResponse) {
	resp.Name = "float64"
}

func (f Float64Function) Definition(ctx context.Context, req function.DefinitionRequest, resp *function.DefinitionResponse) {
	resp.Definition = function.Definition{
		Parameters: []function.Parameter{
			function.Float64Parameter{},
		},
		Return: function.Float64Return{},
	}
}

func (f Float64Function) Run(ctx context.Context, req function.RunRequest, resp *function.RunResponse) {
	var arg float64

	resp.Diagnostics.Append(req.Arguments.Get(ctx, &arg)...)

	resp.Diagnostics.Append(resp.Result.Set(ctx, arg)...)
}
```

```go
func TestFloat64Function_known(t *testing.T) {
	resource.UnitTest(t, resource.TestCase{
		TerraformVersionChecks: []tfversion.TerraformVersionCheck{
			tfversion.SkipBelow(tfversion.Version1_7_0),
		},
		ProtoV6ProviderFactories: map[string]func() (tfprotov6.ProviderServer, error){
			"framework": providerserver.NewProtocol6WithError(New()),
		},
		Steps: []resource.TestStep{
			{
				Config: `
				output "test" {
					value = provider::framework::float64(1.23)
				}`,
				ConfigStateChecks: []statecheck.StateCheck{
					statecheck.ExpectKnownOutputValue("test", knownvalue.Float64Exact(1.23)),
				},
			},
		},
	})
}
```

### Terraform Configuration Files

Following is the configuration used during execution of the test.

```hcl
output "test" {
	value = provider::framework::float64(1.23)
}
```

### Debug Output

https://gist.github.com/bendbennett/ea6aadc53b2e53e66eda630f6b638e64

### Expected Behavior

Test should run sucessfully.

### Actual Behavior

Following error is generated:

```shell
=== RUN   TestFloat64Function_known
    float64_function_test.go:20: Step 1/1 error: Error running pre-apply plan: exit status 1
        
        Error: Error in function call
        
[...]

PR #919, opened 2024-02-06
reflect: Determine equivalency of float32 or float64, and *big.Float via string representation  

bendbennett:
Closes: #914

### DIFF
--- a/internal/reflect/number_test.go
+++ b/internal/reflect/number_test.go
@@ -12,13 +12,14 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+
 	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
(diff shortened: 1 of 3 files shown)


internal/reflect/number_test.go
```diff
@@ -124,25 +103,6 @@ func TestNumber_int(t *testing.T) {
 	}
 }
 
-func TestNumber_intOverflow(t *testing.T) {
```
bflad, line 127: Rather than removing tests, does it make sense to flip their expectations so we still have regression testing for these scenarios?

bflad (commented):
Quick question!

internal/reflect/number_test.go
bendbennett: Apologies if I've misunderstood, but by "flip their expectations" do you mean assert that errors are generated when there is an overflow? If so, my understanding was that tests such as `TestNumber_intOverflowError` were covering these cases.

bflad: Ah okay, so these tests being removed were just covering the `AllowRoundingNumbers` usage to prevent errors which wasn't obvious from the prior test naming. Carry on!

bflad (approved):
Looks good to me 🚀

PR #923, opened 2024-02-16
function: Switch the representation of variadic arguments to `types.Tuple`

austinvalle:
This PR contains a breaking change to provider-defined function handling for variadic parameters, switching the framework type representation from `types.List` to `types.Tuple`, where each element is the same element type.

If provider developers are utilizing slice reflection to retrieve variadic argument values, there will be no changes required due to the updates to support tuple reflection, noted below in `Reflection Changes`:

```go
func (f *ExampleFunction) Run(ctx context.Context, req function.RunRequest, resp *function.RunResponse) {
    var boolArg bool
    var stringVarg []string

    // Despite being a tuple[string,string,etc.] under the hood, reflection will still work here
    resp.Diagnostics.Append(req.Arguments.Get(ctx, &boolArg, &stringVarg)...)
}
```

## Context

Initially, we grouped all variadic arguments for a provider-defined function into a `types.List` to emulate a similar experience to [Go variadic function arguments](https://gobyexample.com/variadic-functions) which are represented by a Go slice. However Terraform core does not handle variadic arguments as a `cty.List`, but rather a `[]cty.Value`, with the type constraint of the variadic parameter applied to each argument individually.

The difference in this handling is specifically problematic when introducing [dynamic type support](https://github.com/hashicorp/terraform-plugin-framework/issues/147) for variadic function parameters, because Terraform will not attempt to conform all variadic arguments to the same type (like a `cty.List` would), eventually causing a decoding error in `terraform-plugin-go` that a provider developer cannot avoid.

While the framework type system doesn't have the same decoding issue as `tftypes`/`terraform-plugin-go`, we need to convert values into `tftypes` to utilize the framework reflection logic 🙂. Switching to `types.Tuple` will allows both `tftypes` and the framework type system to be utilized.

## Reflection Changes
This PR also contains an implementation of limited reflection support for `TupleType` (off the interface `attr.TypeWithElementTypes`). These changes are only intended to support reflection use-cases for tuples where all element types are the same, which will allow provider functions like the following to work properly:
```go
func (f *ExampleFunction) Definition(ctx context.Context, req function.DefinitionRequest, resp *function.DefinitionResponse) {
        // ... other fields ...
            function.BoolParameter{},
        },
        VariadicParameter: function.StringParameter{
			Name: "variadic_param",
        },
    }
}

func (f *ExampleFunction) Run(ctx context.Context, req function.RunRequest, resp *function.RunResponse) {
    var boolArg bool
    var stringVarg []string

    resp.Diagnostics.Append(req.Arguments.Get(ctx, &boolArg, &stringVarg)...)

    // ... other logic ...
}
```

### Notes
- Only the `reflect.Into` changes are necessary for provider-defined function support, however I implemented the `BuildValue` equivalent for tuples for completeness 😄

### DIFF
--- a/.changes/unreleased/NOTES-20240216-090121.yaml
+++ b/.changes/unreleased/NOTES-20240216-090121.yaml
@@ -1,0 +1,8 @@
+kind: NOTES
+body: 'function: Changed the framework type for variadic parameters to `types.TupleType`,
+  where each element is the same element type. Provider-defined functions using a
+  `types.List` for retrieving variadic argument data will need to update their code
+  to use `types.Tuple`.'
+time: 2024-02-16T09:01:21.079352-05:00
+custom:
+  Issue: "923"
--- a/internal/fromproto5/arguments_data.go
+++ b/internal/fromproto5/arguments_data.go
@@ -133,20 +133,18 @@ func ArgumentsData(ctx context.Context, arguments []*tfprotov5.DynamicValue, def
 	}
 
 	if definition.VariadicParameter != nil {
-		// MAINTAINER NOTE:
-		//		Variadic parameters are represented as individual arguments in the CallFunction RPC and Terraform core applies the variadic parameter
-		//		type constraint to each argument individually. For developer convenience, the framework logic below, groups the variadic arguments into a
-		// 		framework Tuple where each element type of the tuple matches the variadic parameter type.
+		// MAINTAINER NOTE: Variadic parameters are represented as individual arguments in the CallFunction RPC and Terraform core applies the variadic parameter
+		// type constraint to each argument individually. For developer convenience, the framework logic below, groups the variadic arguments into a
+		// framework Tuple where each element type of the tuple matches the variadic parameter type.
 		//
-		//		Previously, this logic utilized a framework List with an element type that matched the variadic parameter type. Using a List presented an issue with dynamic
-		//		variadic parameters, as each argument was allowed to be any type "individually", rather than having a single type constraint applied to all dynamic elements,
-		//		like a cty.List in Terraform. This eventually results in an error attempting to create a tftypes.List with multiple element types (when unwrapping from a framework
-		//		dynamic to a tftypes concrete value).
+		// Previously, this logic utilized a framework List with an element type that matched the variadic parameter type. Using a List presented an issue with dynamic
+		// variadic parameters, as each argument was allowed to be any type "individually", rather than having a single type constraint applied to all dynamic elements,
+		// like a cty.List in Terraform. This eventually results in an error attempting to create a tftypes.List with multiple element types (when unwrapping from a framework
+		// dynamic to a tftypes concrete value).
 		//
-		//		While a framework List type can handle multiple dynamic values of different types (due to it's wrapping of dynamic values), `terraform-plugin-go` and `tftypes.List` cannot.
-		// 		Currently, the logic for retrieving argument data is dependent on the tftypes package to utilize the framework reflection logic, requiring us to apply a type constraint
-		//		that is valid in Terraform core and `terraform-plugin-go`, which we are doing here with a Tuple.
-		//
+		// While a framework List type can handle multiple dynamic values of different types (due to it's wrapping of dynamic values), `terraform-plugin-go` and `tftypes.List` cannot.
+		// Currently, the logic for retrieving argument data is dependent on the tftypes package to utilize the framework reflection logic, requiring us to apply a type constraint
+		// that is valid in Terraform core and `terraform-plugin-go`, which we are doing here with a Tuple.
 		variadicType := definition.VariadicParameter.GetType()
 		tupleTypes := make([]attr.Type, len(variadicValues))
 		tupleValues := make([]attr.Value, len(variadicValues))
--- a/internal/fromproto5/arguments_data.go
+++ b/internal/fromproto5/arguments_data.go
@@ -133,8 +133,6 @@ func ArgumentsData(ctx context.Context, arguments []*tfprotov5.DynamicValue, def
 	}
 
 	if definition.VariadicParameter != nil {
-		// Variadic parameters are represented as a framework Tuple where each element type of the tuple matches the variadic parameter type.
-		//
 		// MAINTAINER NOTE:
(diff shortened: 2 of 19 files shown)


.changes/unreleased/NOTES-20240216-090121.yaml
```diff
@@ -0,0 +1,8 @@
+kind: NOTES
+body: 'function: Changed the framework type for variadic parameters to `types.TupleType`,
+  where each element is the same element type. Provider-defined functions using a
+  `types.List` for retrieving variadic argument data will need to update their code
+  to use `types.Tuple`.'
```
austinvalle, line 5: Open to suggestions on wording here. I chose a `NOTES` update because the provider-defined functions are still in tech-preview.

internal/reflect/slice.go
```diff
@@ -90,14 +90,71 @@ func reflectSlice(ctx context.Context, typ attr.Type, val tftypes.Value, target
...
+
+		// go over each of the values passed in, create a Go value of the right
+		// type for them, and add it to our new slice
+		for pos, value := range values {
+			// create a new Go value of the type that can go in the slice
+			targetValue := reflect.Zero(elemType)
+
+			// update our path so we can have nice errors
+			// utilizing `AtListIndex` because tuple paths are also based on the underlying `PathStepElementKeyInt`
+			valPath := path.AtListIndex(pos)
```
austinvalle, line 143: I have been using path.AtListIndex, since the underlying element key is the same for tuples, would it be confusing if I created a new path.AtTupleIndex equivalent for readability?

website/docs/plugin/framework/functions/implementation.mdx
```diff
@@ -207,7 +207,9 @@ func (f *ExampleFunction) Definition(ctx context.Context, req function.Definitio
         Parameters: []function.Parameter{
             function.BoolParameter{},
         },
-        VariadicParameter: function.StringParameter{},
+        VariadicParameter: function.StringParameter{
+			Name: "variadic_param",
+        },
```
austinvalle, line 212: This was something interesting I found, Terraform core validates that the variadic parameter name is different from any of the normal parameters.

Should we change the default `Name` for variadic parameter to avoid this conflict? Also should we add validation for this, maybe when we introduce the user-supplied validation?

```bash
│ Error while loading schemas for plugin components: Failed to obtain provider schema: Could not load the schema for provider registry.terraform.io/austinvalle/sandbox:
│ provider registry.terraform.io/austinvalle/sandbox function "dynamic_test" reuses name "param" for both parameter 0 and its variadic parameter..
```

internal/reflect/slice.go
bflad: I personally don't have a strong opinion either which way! Maybe we should create it so developers don't need to do the mental gymnastics of list indexing being equivalent to tuple indexing. 👍

website/docs/plugin/framework/functions/implementation.mdx
bflad: Oh, very interesting! I would not have necessarily expected that. Let's follow up with that team to see if there is a technical reason for that or if it is more for aesthetics/readability of signatures.

.changes/unreleased/NOTES-20240216-090121.yaml
bflad: My personal opinion is that it should either be in breaking changes or it can be in notes if it mentions the feature its still in technical preview. I would much rather get 👀 on the "scary" change (and hey maybe get more folks looking at it) rather than trying to bury the lede.

internal/reflect/slice.go
```diff
@@ -90,14 +90,71 @@ func reflectSlice(ctx context.Context, typ attr.Type, val tftypes.Value, target
...
+
+			diags.Append(diag.WithPath(path, DiagIntoIncompatibleType{
+				Val:        val,
+				TargetType: target.Type(),
+				Err:        fmt.Errorf("cannot reflect %s using type information provided by %T, tuple type contained no element types but received values", val.Type(), t),
+			}))
+			return target, diags
+		}
+
+		// Ensure that all tuple element types are the same by comparing each element type to the first
```
bflad, line 115: Does this need an exception for successfully working with `[]attr.Value` that should be able to accept multiple element types?

bflad (approved):
Looks good to me 🚀 This will let us iterate forward.

internal/reflect/slice.go
austinvalle: Cool, easy add then! Added `path.AtTupleIndex` in 8c5175a

internal/fromproto5/arguments_data.go
```diff
@@ -133,7 +133,19 @@ func ArgumentsData(ctx context.Context, arguments []*tfprotov5.DynamicValue, def
 	}
 
 	if definition.VariadicParameter != nil {
-		variadicValue, variadicValueDiags := basetypes.NewListValue(definition.VariadicParameter.GetType(), variadicValues)
+		// Variadic parameters are represented as a TupleType where each element is the same element type.
```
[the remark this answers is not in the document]
austinvalle: Good call, that is a much more accurate way to describe that. ⚡ 80c2c2b

bendbennett (approved):
LGTM 🚀

website/docs/plugin/framework/functions/implementation.mdx
austinvalle: I opened https://github.com/hashicorp/terraform-plugin-framework/issues/926 to consider addressing this 👍🏻

internal/reflect/slice.go
austinvalle: This sparked an offline conversation about "why exactly we need to utilize a Tuple instead of a List".

I have attempted to distill that conversation into a maintainer comment. @bendbennett + @bflad if you could take a look at 6dfc0320832b, let me know if that makes things more/less confusing 😆

bflad (approved):
Looks good to me 🚀

internal/reflect/slice.go
bendbennett: Thanks for adding the maintainer comment, the expanded context is very useful.

PR #925, opened 2024-02-20
function: Replace usage of diagnostics with function errors during execution of provider-defined functions

bendbennett:
This PR contains changes for replacing the returning of diagnostics with function errors during the execution of the `CallFunction` RPC and the associated `Run` method defined within provider-defined functions.

A new error type, `function.FuncError` has been introduced.

This will require that provider developers alter the implementation of any previously implemented provider-defined functions as the response type that is returned by the `Run` method has been altered.

Currently, the structure of the `RunResponse` type is:

```go
type RunResponse struct {
	Diagnostics diag.Diagnostics
	Result ResultData
}
```

The changes contained within this PR have restructured the `RunResponse` in the following way:

```go
type RunResponse struct {
	Errors *function.FuncError
	Result ResultData
}
```

This will require that provider-defined function implementations that currently interact with diagnostics will need to be altered to interact with a function error. 

```go
func (f *ExampleFunction) Run(ctx context.Context, req function.RunRequest, resp *function.RunResponse) {
    var boolArg bool
    var stringArg string

    resp.Error = function.ConcatFuncErrors(resp.Error, req.Arguments.Get(ctx, &boolArg, &stringArg))
 
    // ... other logic ...
}
```

### Context

The motivation behind the changes in this PR arise from a desire to keep the data that flows between Terraform core and providers during the `CallFunction` RPC aligned with the native [cty interface](https://pkg.go.dev/github.com/zclconf/go-cty/cty/function#ImplFunc) that is used by Terraform core. There is an [associated PR](https://github.com/hashicorp/terraform/pull/34603) on Terraform core for making equivalent changes on the core side of the protocol.


go.mod
```diff
@@ -4,7 +4,7 @@ go 1.20
 
 require (
 	github.com/google/go-cmp v0.6.0
-	github.com/hashicorp/terraform-plugin-go v0.21.0
+	github.com/hashicorp/terraform-plugin-go v0.21.1-0.20240212100249-164d329d5982
```
bendbennett, line 7: This will need replacing with the released version of `terraform-plugin-go`.

fwerror/function_error.go
```diff
@@ -0,0 +1,27 @@
...
+package fwerror
+
+// FunctionError is an interface for errors generated during the execution
+// of provider-defined functions.
+//
+// See the NewFunctionError constructor for a generic implementation.
+//
+// To add argument position information to an existing function error,
+// see the WithFunctionArgument function.
+type FunctionError interface {
```
bflad, line 13: Given how bespoke these will be compared to other provider development functionality, it may be good to consider:

* Consolidating this into the existing `function` package as it just happens to be a very awkward situation these error types are being introduced, so (famous last words) other error types would not be created in preference of diagnostics, or if they were, they would hopefully be unrelated to functions.
* Whether `FunctionError` (`function.Error` if above happens) should be simplified to a `struct` type instead of an interface. I'm not sure whether we want the extensibility/complexity that is offered by the `diag` package here. It might be interesting to peek in source available provider code whether developers are leveraging the extensibility of diagnostics via those interface(s) or just implementing the existing types/methods. Implementing Go 1.13+ error wrapping capabilities might be good enough, where diagnostics do have separate capabilities beyond pure error handling. 😄

function/run.go
```diff
@@ -19,9 +17,9 @@ type RunRequest struct {
 // RunResponse represents a response to a RunRequest. An instance of this
 // response struct is supplied as an argument to the Function type Run method.
 type RunResponse struct {
-	// Diagnostics report errors or warnings related to defining the function.
-	// An empty slice indicates success, with no warnings or errors generated.
-	Diagnostics diag.Diagnostics
+	// Errors contains errors related to running the function.
+	// A nil error indicates success, with no errors generated.
+	Errors fwerror.FunctionErrors
```
bflad, line 22: Would it be safer for the framework to expose this abstraction more inline with the protocol (e.g. a single function error) and then potentially offer opinionated helper(s) that may change over time for joining multiple errors into one? Is `errors.Join()` or our own version of that acceptable enough? We have tried to avoid the framework from forcing opinionated behaviors on provider developer responses where possible in the past.

website/docs/plugin/framework/functions/errors.mdx
```diff
@@ -0,0 +1,251 @@
...
+## How Errors Affect State
+
+**Returning function errors does not stop the state from being updated**.
+Terraform will still persist the returned state even when function errors
+are returned with it. This is to allow Terraform to persist the values that have
+already been modified when a resource modification requires multiple API
+requests or an API request fails after an earlier one succeeded.
+
+When returning function errors, we recommend resetting the state in the
+response to the prior state available in the configuration.
```
bflad, line 68: Can we double check this? I believe that typically function calls and their errors happen will so early in the operation graph that this would not necessarily be true (except maybe if there are unknown values being passed as arguments).

I'm wondering if its better to just omit this whole section, since "state" is not a function concept and this was mostly relating to managed resources, or whether we need to get into those function error handling implementation details.

website/docs/plugin/framework/functions/errors.mdx
```diff
@@ -0,0 +1,251 @@
...
+    // ... other logic ...
+}
+```
+
+In this example, you will note that we opted to check `resp.Errors`
+instead of `funcErrs`. Technically checking either is correct, however, checking
+the response function errors can help ensure that any response will include the
+expected function errors.
+
+### Creating Diagnostics
```
bflad, line 132:
```suggestion
### Creating Errors
```

website/docs/plugin/framework/functions/errors.mdx
```diff
@@ -0,0 +1,251 @@
...
+
+When working with logic outside the framework, such as interacting with the
+vendor or `net/http` library to make the actual calls to manage infrastructure
```
bflad, line 135: Let's avoid `net/http` here since we want to discourage those types of functions. 👍

bflad (commented):
Some upfront thoughts! Please reach out with questions/concerns.

fwerror/function_error.go
bendbennett: I've gone ahead and refactored `fwerror.FunctionErrors` to `function.FuncError`. 

These changes have removed the `fwerror.FunctionError`, and `fwerror.FunctionErrorWithFunctionArgument` interfaces, as well as removing the `fwerror.FunctionErrors` type. In their place is a single `function.FuncError` struct type. 

A `function.ConcatFuncErrors` helper function has been created which concatenates the `function.FuncError` text fields of two or more instances of `function.FuncError`. I thought this might represent the simplest implementation that would allow handling of multiple instances of `function.FuncError`, given that `errors.Join` returns the `error` interface, requiring that we either alter the return type from `function.FuncError` to `error` and use type assertion where necessary, or implement our own version of `Join`, specifically for `function.FuncError`. In the latter case, unless we re-introduce an interface, I believe we'll need to add a `errs []*FuncError` field to `function.FuncError` struct type, which seems a little odd to me.

Interested to hear your thoughts on the `function.FuncError` implementation, and the associated helper functions.

The usages of `github.com/hashicorp/terraform-plugin-framework/diag` that I found all appeared to be using the pre-defined diagnostics rather than new implementations of the available interfaces:

- https://github.com/DataDog/terraform-provider-datadog/blob/46998f7edae69f6ecd01d522fb7c7fb92e7f401c/datadog/internal/utils/utils.go
- https://github.com/elastic/terraform-provider-elasticstack/blob/4d6e2cbc8d6c09eb3c926ed0be864a89017134e7/internal/clients/config/kibana.go
- https://github.com/lxc/terraform-provider-incus/blob/7dba95ee15d906db2489b96ef18343141c0c4a15/internal/common/incus_import.go
- https://github.com/linode/terraform-provider-linode/blob/b4fb82bcd8e20947be488777aef22092315d85f3/linode/firewall/framework_models.go
-  https://github.com/dnsimple/terraform-provider-dnsimple/blob/bc71acb3b1d1ced3521814dd52632f411d828b73/internal/framework/resources/zone_resource.go

function/run.go
bendbennett: As mentioned above, we now have a single `function.FuncError` struct type, with `function.ConcatFuncErrors` and `function.FuncErrorFromDiags` helper functions. The `function.ConcatFuncErrors` is our current version of `errors.Join()`.

website/docs/plugin/framework/functions/errors.mdx
bendbennett: If it's ok, I'm going to hold off making any further changes to the documentation until we converge on the implementation of `FunctionError`/`FuncError`. But as an aside, I checked this, and you're quite right, returning a diagnostic will result in no alteration to the state file. Returning a function error currently has no effect, presumably because it is not yet being handled within core, and requires https://github.com/hashicorp/terraform/pull/34603 to be merged. Although, I believe I could build Terraform with this branch and test.

go.mod
```diff
@@ -1,10 +1,10 @@
 module github.com/hashicorp/terraform-plugin-framework
 
-go 1.20
+go 1.21
 
 require (
 	github.com/google/go-cmp v0.6.0
-	github.com/hashicorp/terraform-plugin-go v0.21.0
+	github.com/hashicorp/terraform-plugin-go v0.22.0
```
bendbennett, line 7: Running `go get github.com/hashicorp/terraform-plugin-go` results in the addition of the following entry to `go.mod`:

```
toolchain go1.21.6
```

Which would result in an analogous [CI failure](https://github.com/hashicorp/terraform-plugin-mux/actions/runs/8017456549/job/21901361119#step:4:7) to that seen on _terraform-plugin-mux_

Given that our CI runs both Go `1.20` and `1.21`, I removed this directive, but perhaps this should be handled differently?

function/run.go
```diff
@@ -19,9 +15,9 @@ type RunRequest struct {
 // RunResponse represents a response to a RunRequest. An instance of this
 // response struct is supplied as an argument to the Function type Run method.
 type RunResponse struct {
-	// Diagnostics report errors or warnings related to defining the function.
-	// An empty slice indicates success, with no warnings or errors generated.
-	Diagnostics diag.Diagnostics
+	// Error contains errors related to running the function.
+	// A nil error indicates success, with no errors generated.
+	Error *FuncError
```
austinvalle, line 20: Maybe a shout here that `ConcatFuncErrors` can be utilized to simplify smushing together multiple errors? 🔨

go.mod
austinvalle: Oh interesting. It sounds like what you've done is basically the solution: https://go.dev/doc/toolchain

> The same requirement applies in reverse when downgrading: if the go.mod starts at go 1.22.1 and toolchain go1.24rc1, then go get toolchain@go1.22.9 will update only the toolchain line, but go get toolchain@go1.21.3 will downgrade the go line to go 1.21.3 as well. The effect will be to leave just go 1.21.3 with no toolchain line.
>
> The special form toolchain@none means to remove any toolchain line, as in go get toolchain@none or go get go@1.25.0 toolchain@none.

Sounds like we could also potentially just run a `go get toolchain@none` before we run our CI, but you've essentially done that..... Will a `go mod tidy` bring it back?

fwerror/function_error.go
austinvalle: I read this after I left [my comment](https://github.com/hashicorp/terraform-plugin-framework/pull/925#discussion_r1503342432) lol. TLDR: I like `concat`

website/docs/plugin/framework/functions/errors.mdx
bendbennett: I've updated the docs to remove this section, as suggested.

website/docs/plugin/framework/functions/errors.mdx
bendbennett: I've removed the reference to `net/http`.

go.mod
bendbennett: Running `go mod tidy` will not re-instate the toolchain directive.

function/arguments_data.go
```diff
@@ -49,30 +48,28 @@ func (d ArgumentsData) Equal(o ArgumentsData) bool {
...
 	if len(d.values) == 0 {
-		diags.AddError(
-			"Invalid Argument Data Usage",
-			"When attempting to fetch argument data during the function call, the provider code incorrectly attempted to read argument data. "+
-				"This is always an issue in the provider code and should be reported to the provider developers.\n\n"+
-				"Function does not have argument data.",
-		)
-
-		return diags
+		text := "Invalid Argument Data Usage: When attempting to fetch argument data during the function call, the provider code incorrectly attempted to read argument data. " +
```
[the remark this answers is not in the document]
bendbennett: I'm happy to go with consensus on this one. Right now `FuncErrorFromDiags` is formatting as follows:

```go
NewFuncError(fmt.Sprintf("%s: %s", d.Summary(), d.Detail()))
```

So, I'll leave the formatting as it is for right now, but if we decide to update to use `\n\n` between the "summary" and "detail" I'll update it in all locations.

diag/diagnostic.go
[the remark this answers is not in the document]
bendbennett: I've added a note into `doc.go` for both `diag` and `function` packages highlighting the need to use `FuncError` to notify practitioners of issues arising during execution of provider-defined functions.

function/arguments_data.go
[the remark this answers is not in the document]
bendbennett: Updated.

function/func_error.go
```diff
@@ -0,0 +1,126 @@
...
+		}
+	}
+
+	return nil
+}
+
+// FuncErrorFromDiags iterates over the given diagnostics and returns a new function error
+// with the text from all error diagnostics concatenated together. If any of the error diagnostics
+// have a function argument, the first one encountered will be used.
+func FuncErrorFromDiags(ctx context.Context, diags diag.Diagnostics) *FuncError {
```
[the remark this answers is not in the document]
bendbennett: I've expaned the Go docs for `FuncErrorFromDiags` to highlight how warning diagnostics will be processed. I've also added a [section to the website docs](https://github.com/hashicorp/terraform-plugin-framework/pull/925/commits/605ae1568ae410e1aa41a393be5023d8e8b94ce4#diff-d6ec4af155af892eabf61f24a4c62769264838d926afa2baf7a94727625f1803R132).

function/func_error.go
```diff
@@ -0,0 +1,126 @@
...
+
+// Equal returns true if the other function error is wholly equivalent.
+func (fe *FuncError) Equal(other *FuncError) bool {
+	var funcErr *FuncError
+
+	ok := errors.As(other, &funcErr)
+
+	if !ok {
+		return false
+	}
```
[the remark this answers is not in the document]
bendbennett: Oooopsie. Removed.

function/run.go
bendbennett: Have added a comment.

function/arguments_data.go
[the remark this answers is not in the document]
austinvalle: I'm fine leaving it as-is. Feels like avoiding formatting with `/n/n` is probably best 👍🏻

website/docs/plugin/framework/functions/errors.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+```go
+func (f *ExampleFunction) Run(ctx context.Context, req function.RunRequest, resp *function.RunResponse) {
+```
+
+This is the most common form for `FuncError`: a single error whose text
+is the concatenated error text from one or more errors. This approach allows
+your provider to inform practitioners about all relevant errors at the same
+time, allowing practitioners to fix their configuration or environment more
+quickly. You should only concatenate a `FunctError` and never replace or
```
austinvalle, line 36:
```suggestion
quickly. You should only concatenate a `FuncError` and never replace or
```

function/func_error.go
```diff
@@ -0,0 +1,118 @@
...
+func NewArgumentFuncError(functionArgument int64, text string) *FuncError {
+	return &FuncError{
+		Text:             text,
+		FunctionArgument: &functionArgument,
+	}
+}
+
+// FuncError is an error type specifically for function errors.
+type FuncError struct {
+	Text             string
```
austinvalle, line 34: Apologies, I had commented about the variable name of `text` while forgetting the underlying struct field was named `Text` 😆

website/docs/plugin/framework/functions/errors.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+`Text` is a practitioner-oriented description of the problem. This should
+contain sufficient detail to provide both general and more specific information
+regarding the issue. For example "Error executing function: foo can only contain
+letters, numbers, and digits."
+
+### FunctionArgument
+
+`FunctionArgument` identifies the specific function argument position that caused the
+error. Only errors that pertain to a function argument will include this information.
```
austinvalle, line 56: nit: Maybe add a note that it is an `int64` type and that the argument position is zero-based

website/docs/plugin/framework/functions/errors.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+To craft the message of a function error, it is recommended to use sufficient
+detail to convey both the cause of the error and as much contextual,
+troubleshooting, and next action information as possible. These details can
+use newlines for easier readability where necessary.
+
+#### NewFuncError
+
+When creating function errors that affect an entire data source, provider, or
+resource, and where a `function.FunctionError` is already available such as within
```
austinvalle, line 92: > _When creating function errors that affect an entire data source, provider, or
resource_

I didn't quite understand what this sentence is referring to, can you expand on this?

website/docs/plugin/framework/functions/errors.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+	val, err := // operation that may return an error
+
+	if err != nil {
+        resp.Error = ConcatFuncErrors(resp.Error, function.NewFuncError("Error performing operation: " + err.Error()))
+        return
+    }
+
+    // ... further logic ...
+}
```
austinvalle, line 111: nit: Go formatting 🙂 (does our website auto-fix the formatting? I wouldn't be surprised if it did)

Same formatting is a little off for some of the other examples (at least in GitHub)

austinvalle (approved):
Great work! :shipit: 🚀

function/func_error.go
bendbennett: No worries. Have left it as `errMsg`.

website/docs/plugin/framework/functions/errors.mdx
bendbennett: Updated to include your suggestion.

website/docs/plugin/framework/functions/errors.mdx
bendbennett: I've stripped that out.

website/docs/plugin/framework/functions/errors.mdx
bendbennett: Ah, the old tabs/spaces fun and games. Fixed throughout the docs.

function/func_error.go
bflad: Nit: I know its trivial in this case, but can we please ensure all exported types and fields have Go documentation? For example, the website documentation for these fields seems great.

function/result_data.go
```diff
@@ -22,7 +21,7 @@ type ResultData struct {
 }
 
 // Equal returns true if the value is equivalent.
-func (d ResultData) Equal(o ResultData) bool {
+func (d *ResultData) Equal(o ResultData) bool {
```
bflad, line 24: Nit: Was there a reason for this change? It now means that this method needs to potentially account for `d == nil`

function/result_data.go
```diff
@@ -32,24 +31,22 @@ func (d ResultData) Equal(o ResultData) bool {
...
 
 	d.value = reflectValue
 
-	return diags
+	return nil
 }
 
 // Value returns the saved value.
-func (d ResultData) Value() attr.Value {
+func (d *ResultData) Value() attr.Value {
```
bflad, line 49: Nit: Same here.

bflad (approved):
Looks good to me.

website/docs/plugin/framework/functions/errors.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+```console
+$ terraform plan
+# ... other plan output ...
+╷
+│ Error: Message
+│ 
+│   on example.tf line #:
+│    #: source configuration line
+╵
```
bflad, line 22: FYI the error diagnostic for a regular function error appears to be:

```
        Error: Error in function call
        
          on example.tf line #:
          #: source configuration line
        
        Call to function "{FUNCTION}" failed: {TEXT}.
```

And for an argument error:

```
        Error: Invalid function argument
        
          on example.tf line #:
          #: source configuration line
        
        Invalid value for "{POSITION}" parameter: {TEXT}.
```

function/result_data.go
bendbennett: The change was just to make all methods on the `ResultData` struct on a pointer receiver. Have reverted.

function/result_data.go
bendbennett: The change was just to make all methods on the `ResultData` struct on a pointer receiver.

website/docs/plugin/framework/functions/errors.mdx
bendbennett: Thanks. Have updated the docs with an example of both.

Issue #590, opened 2022-12-16
Consider Surfacing Framework Type System Data for Config/Plan/State


bflad:
### Module version

```
v1.0.0
```

### Use-cases

Provider developers should not need to drop down to the `github.com/hashicorp/terraform-plugin-go/tftypes` type system for advanced use cases that require working with schema-based data directly. In particular, `tfsdk.Config`, `tfsdk.Plan`, and `tfsdk.State` type `Raw` fields are currently `tftypes.Value`. Instead, providers should be able to work with the framework's type system for consistency, such as `types.Object`. If the `tftypes` data handling was still necessary by providers for whatever reason, they could still use the type implementations of `ToTerraformValue()` and `FromTerraformValue()` to roundtrip back and forth.

This effort would also help clean up _a lot_ of framework internal logic that goes back and forth between the two type systems.

### Proposal

_This will require a large amount of design effort to complete successfully and this proposal is not currently exhaustive. It is being written in the context of closing out the generic #172._

Likely, some `tftypes` functionality such as walking and transforming values may be desirable in the framework type system or at least available internally within the framework.

Then, one of the following approaches could be taken to expose the data using the framework's type system:

#### Existing Type Approach

Add a new exported `types.Object` retrieval method to the existing `tfsdk` types. Deprecate the `Raw` field.

This introduces the least changes for providers and the framework, but does not fix the framework's technical debt imposed by the data storage being an exported `tftypes.Value`. Without a setter method, this also means that providers must use `(types.Object).ToTerraformValue()` to save the data back, which could be awkward.

#### New Type Approach

New schema-based data types could be introduced in the `datasource`, `provider`, and `resource` packages based off the framework type system which deprecate the existing `tfsdk` types. This enables fully executing the vision of each concept being self-contained within its own package in the framework. Various requests/responses would need the `tfsdk` based fields to be deprecated in preference of new fields using the new types.

This approach would also allow a more tailored design that no longer exposes any implementation details as errantly mutable fields, but more design would be needed on that front.

### References

- #172
- #366

bflad:
[...]

PR #933, opened 2024-02-28
resource/schema: Ensure invalid attribute default value errors are raised

### LINKED ISSUE #930: Incorrect element type of default list or map value in schema causes all default computed values of schema to be silently set to `null`
### Module version

```
github.com/hashicorp/terraform-plugin-framework v1.5.0
```

```go
type ExampleResourceModel struct {
	ConfigurableAttribute types.String `tfsdk:"configurable_attribute"`
	Defaulted             types.String `tfsdk:"defaulted"`
	Id                    types.String `tfsdk:"id"`
	Foo                   []FooEntry   `tfsdk:"foo"`
}

type FooEntry struct {
}

func (r *ExampleResource) Schema(ctx context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
	resp.Schema = schema.Schema{
		// This description is used by the documentation generator and the language server.
		MarkdownDescription: "Example resource",

			"configurable_attribute": schema.StringAttribute{
				MarkdownDescription: "Example configurable attribute",
				Optional:            true,
			},
			"defaulted": schema.StringAttribute{
				MarkdownDescription: "Example configurable attribute with default value",
				Optional:            true,
				Computed:            true,
				Default:             stringdefault.StaticString("example value when not configured"),
			},
			"id": schema.StringAttribute{
				Computed:            true,
				MarkdownDescription: "Example identifier",
				},
			},
			"foo": schema.ListNestedAttribute{
				MarkdownDescription: "test",
				Optional:            true,
					Attributes: map[string]schema.Attribute{},
				},
				Computed: true,
				Default:  listdefault.StaticValue(types.ListValueMust(types.Int64Type, []attr.Value{})),
			},
		},
	}
}

```

```hcl
terraform {
  required_providers {
    scaffolding = {
      source = "edu/scaffolding"
    }
  }
}

provider "scaffolding" {}

resource "scaffolding_example" "example" {
  configurable_attribute = "some-value"
}

output "test" {
  value = scaffolding_example.example
}
```

### Expected Behavior
`output.test` should have `defaulted` set to `"example value when not configured"`

### Actual Behavior

`output.test.defaulted` is `null`

```
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  + create

  # scaffolding_example.example will be created
  + resource "scaffolding_example" "example" {
      + configurable_attribute = "some-value"
      + id                     = (known after apply)
    }

Plan: 1 to add, 0 to change, 0 to destroy.

Changes to Outputs:
  + test = {
      + configurable_attribute = "some-value"
      + defaulted              = null
      + foo                    = null
      + id                     = (known after apply)
    }
```

When commenting out `foo` in the schema and the model, the correct values appear:

```
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  + create

  # scaffolding_example.example will be created
  + resource "scaffolding_example" "example" {
      + configurable_attribute = "some-value"
      + defaulted              = "example value when not configured"
      + id                     = (known after apply)
    }

Plan: 1 to add, 0 to change, 0 to destroy.

Changes to Outputs:
  + test = {
      + configurable_attribute = "some-value"
      + defaulted              = "example value when not configured"
      + id                     = (known after apply)
    }
```

I am fairly new to Terraform plugins and golang, so I could be holding it wrong. But I think this case should at least raise an error. I spent many hours trying to figure out why my `Optional`+`Computed` properties were always `null` on plan/apply.

### Steps to Reproduce
1. Apply above patch to https://github.com/hashicorp/terraform-provider-scaffolding-framework
2. `go build`
3. `cd examples/provider`
4. `tf plan`

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/590
Closes #930

Previously the logic handling attribute `Default` values would silently ignore any type errors, which would lead to confusing planning data behaviors. This updates the logic to raise those error properly and adds covering unit testing.

These error messages are using the underlying `tftypes` type system errors which is currently a pragmatic compromise throughout various parts of the framework logic that bridges between both type systems to save additional type assertion logic and potential bugs relating to those conversions. In the future if the internal `tftypes` handling and exported fields are replaced with the framework type system types, this logic would instead return error messaging based on the framework type system errors.

This also will enhance the schema validation logic to check any `Default` response value and compare its type to the schema, which will raise framework type system errors during the `GetProviderSchema` RPC, or during schema unit testing if provider developers have implemented that additional testing.

### DIFF
--- a/internal/fwschemadata/data_default.go
+++ b/internal/fwschemadata/data_default.go
@@ -21,6 +21,7 @@ import (
 // when configRaw contains a null value at the same path.
 func (d *Data) TransformDefaults(ctx context.Context, configRaw tftypes.Value) diag.Diagnostics {
 	var diags diag.Diagnostics
+	var err error
 
 	configData := Data{
 		Description:    DataDescriptionConfiguration,
@@ -28,8 +29,7 @@ func (d *Data) TransformDefaults(ctx context.Context, configRaw tftypes.Value) d
 		TerraformValue: configRaw,
 	}
 
-	// Errors are handled as richer diag.Diagnostics instead.
-	d.TerraformValue, _ = tftypes.Transform(d.TerraformValue, func(tfTypePath *tftypes.AttributePath, tfTypeValue tftypes.Value) (tftypes.Value, error) {
+	d.TerraformValue, err = tftypes.Transform(d.TerraformValue, func(tfTypePath *tftypes.AttributePath, tfTypeValue tftypes.Value) (tftypes.Value, error) {
 		fwPath, fwPathDiags := fromtftypes.AttributePath(ctx, tfTypePath, d.Schema)
 
 		diags.Append(fwPathDiags...)
@@ -302,6 +302,16 @@ func (d *Data) TransformDefaults(ctx context.Context, configRaw tftypes.Value) d
 
 		return tfTypeValue, nil
 	})
+
+	// Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/930
+	if err != nil {
+		diags.Append(diag.NewErrorDiagnostic(
+			"Error Handling Schema Defaults",
+			"An unexpected error occurred while handling schema default values. "+
+				"Please report the following to the provider developer:\n\n"+
+				"Error: "+err.Error(),
+		))
+	}
 
 	return diags
 }
--- a/resource/schema/list_attribute.go
+++ b/resource/schema/list_attribute.go
@@ -269,6 +269,10 @@ func (a ListAttribute) ValidateImplementation(ctx context.Context, req fwschema.
 
 		resp.Diagnostics.Append(defaultResp.Diagnostics...)
 
+		if defaultResp.Diagnostics.HasError() {
+			return
+		}
+
 		if a.ElementType != nil && !a.ElementType.Equal(defaultResp.PlanValue.ElementType(ctx)) {
 			resp.Diagnostics.Append(fwschema.AttributeDefaultElementTypeMismatchDiag(req.Path, a.ElementType, defaultResp.PlanValue.ElementType(ctx)))
 		}
--- a/resource/schema/list_attribute.go
+++ b/resource/schema/list_attribute.go
@@ -251,7 +251,26 @@ func (a ListAttribute) ValidateImplementation(ctx context.Context, req fwschema.
 		resp.Diagnostics.Append(fwschema.AttributeMissingElementTypeDiag(req.Path))
 	}
 
-	if !a.IsComputed() && a.ListDefaultValue() != nil {
-		resp.Diagnostics.Append(nonComputedAttributeWithDefaultDiag(req.Path))
+	if a.ListDefaultValue() != nil {
+		if !a.IsComputed() {
+			resp.Diagnostics.Append(nonComputedAttributeWithDefaultDiag(req.Path))
+		}
+
+		// Validate Default implementation. This is safe unless the framework
+		// ever allows more dynamic Default implementations at which the
+		// implementation would be required to be validated at runtime.
+		// Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/930
+		defaultReq := defaults.ListRequest{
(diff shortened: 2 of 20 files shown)


internal/fwschemadata/data_default.go
```diff
@@ -21,15 +21,15 @@ import (
...
 
 	configData := Data{
 		Description:    DataDescriptionConfiguration,
 		Schema:         d.Schema,
 		TerraformValue: configRaw,
 	}
 
-	// Errors are handled as richer diag.Diagnostics instead.
-	d.TerraformValue, _ = tftypes.Transform(d.TerraformValue, func(tfTypePath *tftypes.AttributePath, tfTypeValue tftypes.Value) (tftypes.Value, error) {
+	d.TerraformValue, err = tftypes.Transform(d.TerraformValue, func(tfTypePath *tftypes.AttributePath, tfTypeValue tftypes.Value) (tftypes.Value, error) {
```
austinvalle, line 32: This is awesome ❤️ . I don't want to stretch this PR too far, but do we want to just add similar error handling now to other places we follow a similar pattern with `tftypes.Transform`. Quick search shows:
- `(*Data).NullifyCollectionBlocks`
- `(*Data).ReifyNullCollectionBlocks`

resource/schema/list_attribute.go
```diff
@@ -251,7 +251,26 @@ func (a ListAttribute) ValidateImplementation(ctx context.Context, req fwschema.
...
-		resp.Diagnostics.Append(nonComputedAttributeWithDefaultDiag(req.Path))
+	if a.ListDefaultValue() != nil {
+		if !a.IsComputed() {
+			resp.Diagnostics.Append(nonComputedAttributeWithDefaultDiag(req.Path))
+		}
+
+		// Validate Default implementation. This is safe unless the framework
+		// ever allows more dynamic Default implementations at which the
+		// implementation would be required to be validated at runtime.
+		// Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/930
```
austinvalle, line 262: Interesting 👀

resource/schema/list_attribute.go
```diff
@@ -251,7 +251,26 @@ func (a ListAttribute) ValidateImplementation(ctx context.Context, req fwschema.
...
+		}
+		defaultResp := &defaults.ListResponse{}
+
+		a.ListDefaultValue().DefaultList(ctx, defaultReq, defaultResp)
+
+		resp.Diagnostics.Append(defaultResp.Diagnostics...)
+
+		if a.ElementType != nil && !a.ElementType.Equal(defaultResp.PlanValue.ElementType(ctx)) {
+			resp.Diagnostics.Append(fwschema.AttributeDefaultElementTypeMismatchDiag(req.Path, a.ElementType, defaultResp.PlanValue.ElementType(ctx)))
+		}
```
austinvalle, line 274: Should we still check the equality of the element types if there are diagnostics retrieving the default?

> Same question applies to the other type comparisons in collections/nested attributes

austinvalle (approved):
LGTM 🚀 :shipit: - just a couple questions

bendbennett (commented):
LGTM 🚀

internal/fwschemadata/data_default.go
bflad: Created https://github.com/hashicorp/terraform-plugin-framework/issues/938 for tracking.

resource/schema/list_attribute.go
bflad: I was debating this, but I think you are right. Even though `Default` value handling should be static/offline today, it would probably be annoying to see both diagnostics if, for example, you return an error diagnostic early without setting a default value with a fully realized/valid type. I'll hopefully have some time later today to update this.

bflad: Does this look okay? https://github.com/hashicorp/terraform-plugin-framework/pull/933/commits/6056a2883a9c829a0e3b235da2cb07b70f6990a6

PR #949, opened 2024-03-11
`attr.Value` → `attr.Type`

skyzyx:
These should be types, according to the Golang LSP in my IDE.


bendbennett (approved):
@skyzyx thank you for this fix.

PR #951, opened 2024-03-13
Add clarification for nil check of ProviderData in data source and resource Configure() method

bendbennett:
Reference: https://github.com/hashicorp/tutorials/issues/1944

Updating documentation to clarify that a `nil` check of `ProviderData` should always be performed in the `Configure()` method of data sources and resources.


bflad (approved):
Excellent update 🚀

PR #956, opened 2024-03-15
docs: Explicitly call out object parameter attributes are required

bflad:
Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/955

This attempts to document the existing behavior of Terraform and the framework with relation to object parameters requiring all object attributes to be explicitly configured by adding an early paragraph about the behavior. The other updates on the page are to continue the overall parameter page consistency of configuration examples being used in the implementation examples later in the pages, so here the `attr3` and `null` value handling is shown in all the examples.

In the future, there may be enhancements which could enable "optional" object attributes at which time this new documentation can be updated in relation to that potential new functionality.


austinvalle (approved):
Great addition! 🚀

PR #957, opened 2024-03-15
Correct resource State Move documentation

ewbankkit:
The [_Resource State Move_ documentation](https://developer.hashicorp.com/terraform/plugin/framework/resources/state-move#implementation) states that a resource should implement a `StateMove` method, satisfying the `resource.ResourceWithStateMove` interface.
The interface is actually named `resource.ResourceWithMoveState`, and the method `MoveState`.
This PR corrects the documentation.


bflad (approved):
Thank you for this fix, @ewbankkit 🚀

Issue #893, opened 2023-12-18
function: Support Type-Based Validation with Argument Position Data


bflad:
### Module version

```
v1.5.0 (not yet released)
```

### Use-cases

Provider developers can create reusable [custom types](https://developer.hashicorp.com/terraform/plugin/framework/handling-data/types/custom) either in their provider codebases or as a shared Go module, such as [terraform-plugin-framework-jsontypes](https://github.com/hashicorp/terraform-plugin-framework-jsontypes). One nice benefit of a custom type is the ability to support type-based validation, meaning that any usage of that custom type in a schema automatically will perform the validation logic. This saves developers time and repetitive lines of code, such as declaring validators on each attribute definition.

When the type-based validation support was introduced, it was done in a manner that was using the lower-level type system (`tftypes.Value`) and very schema-specific (`path.Path`). This was documented as sub-optimal in https://github.com/hashicorp/terraform-plugin-framework/issues/589 when the framework v1.0.0 was being released, but was not handled prior due to time.

While the usage of `tftypes.Value` would be okay for the upcoming provider-defined function support, the `path.Path` usage is invalid since function argument information in diagnostics is very different. The custom type `Validate` method is called, however the `path.Path` passed is always errant since there is no schema-based path information for function arguments.

### Proposal

Once #589 is sorted, add the updated type-based validation bits to `internal/fromproto5.ArgumentsData()` and `internal/fromproto6.ArgumentsData()` where it appropriately sets the function argument position for the validation request.

### References

- https://github.com/hashicorp/terraform-plugin-framework/issues/589

Issue #589, opened 2022-12-16
attr/xattr: Consider Deprecating tftypes.Value Handling In Preference of attr.Value Handling


bflad:
### Module version

```
v1.0.0
```

### Use-cases

Provider developers should not need to drop down to the `github.com/hashicorp/terraform-plugin-go/tftypes` type system when creating custom types via `attr.Value` with validation. In particular:

`attr/xattr.TypeWithValidate`

- `Validate(context.Context, tftypes.Value, path.Path) diag.Diagnostics`

### Proposal

Similar to the framework type system specific `types/basetypes` package interfaces, such as `StringTypable`/`StringValuable`, introduce validation interfaces that are based on the framework types. Potentially a new `types/validation` package. The `types/basetypes` package _cannot_ be used as `diag.Diagnostics` has a dependency on `basetypes.SetValue`.

The existing `attr/xattr.TypeWithValidate` interface and its `Validate` method would be deprecated, instead encouraging the usage of the new interfaces. The `attr/xattr` package cannot be removed until the next major version due to compatibility promises.

Proposed `types/validation` package:

```go
package validation

type Bool interface {
  basetypes.BoolValuable

  ValidateBool(context.Context, BoolRequest, *BoolResponse)
}

type Float64 interface {
  basetypes.Float64Valuable

  ValidateFloat64(context.Context, Float64Request, *Float64Response)
}

type Int64 interface {
  basetypes.Int64Valuable

  ValidateInt64(context.Context, Int64Request, *Int64Response)
}

type List interface {
  basetypes.ListValuable

  ValidateList(context.Context, ListRequest, *ListResponse)
}

type Map interface {
  basetypes.MapValuable

  ValidateMap(context.Context, MapRequest, *MapResponse)
}

type Number interface {
  basetypes.NumberValuable

  ValidateNumber(context.Context, NumberRequest, *NumberResponse)
}

type Object interface {
  basetypes.ObjectValuable

  ValidateObject(context.Context, ObjectRequest, *ObjectResponse)
}

type Set interface {
  basetypes.SetValuable

  ValidateSet(context.Context, SetRequest, *SetResponse)
}

type String interface {
  basetypes.StringValuable

  ValidateString(context.Context, StringRequest, *StringResponse)
}
```

Note this proposal switches implementation to a request/response pattern similar to other framework handling so that future changes to the requests/responses can be done without updates to the interfaces themselves. Those types can be fairly straightforward for now to match the existing handling:

```go
[...]

PR #968, opened 2024-03-22
Add type-based provider-defined function parameter validation

bendbennett:
Closes: #589 
Closes: #893 

### Background
Provider-defined functions can accept parameters, or arguments as input. There is an opportunity to provide validation of such parameters in an analogous manner to the validation of values supplied in configuration for attributes, by implementing parameter-based and type-based validation for provider-defined function parameters. 

This PR is concerned with the addition of type-based validation, which enables provider developers using custom value types to implement parameter validation.

### Validation Package
This PR adds type-based provider-defined function parameter validation through the introduction of a `validation` package with a new `ValidateableParameter` interface.

```go
type ValidateableParameter interface {
	ValidateParameter(context.Context, ValidateParameterRequest, *ValidateParameterResponse)
}

type ValidateParameterRequest struct {
	Position int64
}

type ValidateParameterResponse struct {
	Error *function.FuncError
}
```

### Deprecation of `xattr.TypeWithValidate`
This PR also introduces a new `ValidateableAttribute` interface, and deprecates the `xattr.TypeWithValidate` interface. Provider developers who are using custom value types that will be used in the context of both attribute validation and parameter validation would need to implement both interfaces on the custom value type, but could share the validation logic to reduce duplication.

```go
type ValidateableAttribute interface {
	ValidateAttribute(context.Context, ValidateAttributeRequest, *ValidateAttributeResponse)
}

type ValidateAttributeRequest struct {
	Path path.Path
}

type ValidateAttributeResponse struct {
	Diagnostics diag.Diagnostics
}
```


attr/xattr/attribute.go
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+// TODO: Just putting here for the moment to get rid of import cycle.
```
bendbennett, line 4: As per the TODO. Interested in hearing opinions on where this interface would best be located.

attr/xattr/attribute.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+)
+
+// ValidateableAttribute defines an interface for validating an attribute value.
+type ValidateableAttribute interface {
```
bendbennett, line 14: Interface is currently located in `xattr` package to remove import cycles present when it is in `validation` package:

```shell
package github.com/hashicorp/terraform-plugin-framework/datasource
        imports github.com/hashicorp/terraform-plugin-framework/datasource/schema
        imports github.com/hashicorp/terraform-plugin-framework/internal/fwschema
        imports github.com/hashicorp/terraform-plugin-framework/resource/schema/defaults
        imports github.com/hashicorp/terraform-plugin-framework/types
        imports github.com/hashicorp/terraform-plugin-framework/types/basetypes
        imports github.com/hashicorp/terraform-plugin-framework/internal/reflect
        imports github.com/hashicorp/terraform-plugin-framework/types/validation
        imports github.com/hashicorp/terraform-plugin-framework/function
        imports github.com/hashicorp/terraform-plugin-framework/internal/reflect: import cycle not allowed
```

internal/fwfunction/xfwfunction/doc.go
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+// Package xfwfunction contains additional functions for use with provider defined functions.
+// This package is separate from the fwfunction package to prevent import cycles.
```
bendbennett, line 5: This package is solely to remove the import cycle that arises from having the `Parameter` function in the `fwfunction` packge:

```shell
package github.com/hashicorp/terraform-plugin-framework/function
        imports github.com/hashicorp/terraform-plugin-framework/internal/fwfunction
        imports github.com/hashicorp/terraform-plugin-framework/function: import cycle not allowed
```

.changes/unreleased/FEATURES-20240404-154553.yaml
```diff
@@ -0,0 +1,8 @@
+kind: FEATURES
+body: 'types/validation: Added `BoolValuableWithValidateableAttribute`, `Float64ValuableWithValidateableAttribute`,
```
bendbennett, line 2: The utility of the value-type interfaces is debatable as the interface that they embed, `ValidateableAttribute`, can be implemented directly by a custom value type. However, the value-type interfaces do streamline defining the interfaces that a custom value type should implement. For example

Using `xattr.ValidateableAttribute`:

```go
var _ basetypes.StringValuable = CustomStringValue{}
var _ xattr.ValidateableAttribute = CustomStringValue{}
```

Using `xattr. StringValuableWithValidateableAttribute`:
```go
var _ StringValuableWithValidateableAttribute = CustomStringValue
```

.changes/unreleased/FEATURES-20240404-155023.yaml
```diff
@@ -0,0 +1,8 @@
+kind: FEATURES
+body: 'types/validation: Added `BoolValuableWithValidateableParameter`, `Float64ValuableWithValidateableParameter`,
```
bendbennett, line 2: The utility of the value-type interfaces is debatable as the interface that they embed, `ValidateableParameter`, can be implemented directly by a custom value type. However, the value-type interfaces do streamline defining the interfaces that a custom value type should implement. For example

Using `validation.ValidateableParameter`:

```go
var _ basetypes.StringValuable = CustomStringValue{}
var _ validation.ValidateableParameter = CustomStringValue{}
```

Using `validation. StringValuableWithValidateableParameter`:
```go
var _ validation.StringValuableWithValidateableParameter = CustomStringValue
```

attr/xattr/attribute.go
austinvalle: An unfortunate consequence of the `reflect` package still using `tftypes` 🙁, not much we can realistically do here so I think leaving it in `xattr` makes sense.

internal/fwfunction/xfwfunction/doc.go
austinvalle: Off the top of my head, i can't think of a better option 😞

.changes/unreleased/FEATURES-20240404-154553.yaml
austinvalle: Currently, I'm leaning towards omitting the `{TYPE}ValuableWithValidatableAttribute` and `{TYPE}ValuableWithValidatableParameter` interfaces.

In the case of implementing all 3 interfaces you'd potentially have something like:
```go
var (
   _ validation.StringValuableWithValidateableAttribute = CustomStringValue
   _ validation.StringValuableWithValidateableParameter = CustomStringValue
)
```
vs.
```go
var (
   _ basetypes.StringValuable = CustomStringValue
   _ xattr.ValidateableAttribute = CustomStringValue
   _ function.ValidateableParameter = CustomStringValue
)
```

I think the second is a little more readable and explicit

website/docs/plugin/framework/functions/parameters/index.mdx
```diff
@@ -94,3 +94,7 @@ Parameter names are used in runtime errors to highlight which parameter is causi
 │
 │ Invalid value for "bool_param" parameter: a bool is required.
 ```
+
+## Parameter Validation
```
SBGoods, line 98: Should the parameter-based validation documentation be here instead since value and type validations are already covered in `website/docs/plugin/framework/handling-data/types/custom.mdx`? 

It's a little awkward since we don't have any specific page for function-specific validations to serve as an index but there is not enough specific info for functions that isn't already covered in `website/docs/plugin/framework/validation.mdx` or `website/docs/plugin/framework/handling-data/types/custom.mdx`

attr/xattr/attribute.go
SBGoods: I agree with leaving it in the `xattr` package.

.changes/unreleased/FEATURES-20240404-154553.yaml
SBGoods: I think that `{TYPE}ValuableWithValidatableAttribute` and `{TYPE}ValuableWithValidatableParameter` interfaces are easier to discover with the clear naming, but since existing custom value types already implement `basetypes.{Type}Valuable`, I think it might be confusing to remove that interface just to obfuscate it behind another interface.

austinvalle (commented):
Looks great! I left some general thoughts around package structure and some of the questions you asked.

internal/fwfunction/xfwfunction/doc.go
bendbennett: I like your thinking! Decided to just inline the logic from the `Parameter()` function within the `ArgumentsData()` function. Test coverage for `ArgumentsData()` already checks for error conditions (i.e., out-of-bounds errors), and various permutations of having one or multiple arguments.

.changes/unreleased/FEATURES-20240404-154553.yaml
bendbennett: Thanks for the feedback. I've removed the `<Type>ValuableWithValidateable<Attribute|Parameter>` interfaces.

website/docs/plugin/framework/functions/parameters/index.mdx
bendbennett: Fair point. I've added some documentation to cover implementation of `validation.ValidateableParameter` interface and linked through to custom types.

internal/reflect/interfaces.go
```diff
@@ -74,18 +75,38 @@ func FromUnknownable(ctx context.Context, typ attr.Type, val Unknownable, path p
...
 
-		if typeWithValidate, ok := typ.(xattr.TypeWithValidate); ok {
-			diags.Append(typeWithValidate.Validate(ctx, tfVal, path)...)
+		res, err := typ.ValueFromTerraform(ctx, tfVal)
+		if err != nil {
+			return nil, append(diags, valueFromTerraformErrorDiag(err, path))
+		}
+
+		switch t := res.(type) {
+		case xattr.ValidateableAttribute:
```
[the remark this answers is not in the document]
bendbennett: > For example, in the case of a custom type only being used on a function parameter, it might not implement this ValidateableAttribute interface. When the original ArgumentData is being built from tfplugin-go, it would call the ValidateableParameter interface, but when calling (ArgumentData).Get or (ResultData).Set it would be using the ValidateableAttribute interface (via the reflection) and therefore not get called.

If a custom type that is only being used as a function parameter implements `ValidateableParameter` but does not implement `ValidateableAttribute`, the call to `ValidateParameter()` will occur during the `CallFunction` RPC when `fromproto5/6.ArgumentsData()` is called. The call to `fromproto5/6.ArgumentsData()` occurs, prior to calling `(ArgumentData).Get()` or `(ResultData).Set()`, the latter of which are typically called within the `Run()` method of the provider-defined function. Does this answer your concern?

> I wonder how big of a refactor it would be to completely remove validation from the reflect package and make it the responsibility of the caller to validate the reflection value is valid. So (ArgumentData).Get, (ResultData).Set, (Data).Get, (Data).Set, etc. would need to perform this validation.

I chatted with Brian about this out-of-band, as I was also wondering about the embedded validation within reflection. I think this is worthy of further investigation but I believe would be best suited as a follow-up issue/discussion.

> Or perhaps we should be attempting to add the parameter validation into the reflect logic, although we'll likely run into import cycles there because of FuncError, so maybe if we don't do anything here it's just a documentation problem?

In terms of the documentation, is the concern that it may not be clear under which circumstances to implement `ValidateableParameter` vs `ValidateableAttribute`? There's some updates to the docs in this PR to try and make this clear, but let me know if you think this should be expanded.

types/basetypes/float64_type.go
```diff
@@ -9,16 +9,18 @@ import (
...
 	"github.com/hashicorp/terraform-plugin-framework/attr/xattr"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/path"
-	"github.com/hashicorp/terraform-plugin-go/tftypes"
 )
 
 // Float64Typable extends attr.Type for float64 types.
 // Implement this interface to create a custom Float64Type type.
 type Float64Typable interface {
+	//nolint:staticcheck // xattr.TypeWithValidate is deprecated, but we still need to support it.
```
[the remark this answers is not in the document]
bendbennett: I was wondering about this. For float64 and int64 I was thinking that this may not be necessary as the calls to  `<Float64|Int64>Type.ValueFromTerraform` ensure that the `tftypes.Value` supplied can be represented as 64-bit floating point, and integer values, respectively. The constructors on the corresponding value types only accept 64-bit floating point, and integer values too (e.g., `NewFloat64Value`). So it doesn't seem that there is a way to generate a float64 or int64 value type that would have a value that couldn't be represented as a 64-bit floating point, or integer value, respectively.

I'll create a follow-up issue for list, map, set. Let me know if you think float64 and int64 should be included and we can discuss this further.

types/basetypes/list_value_test.go
```diff
@@ -161,9 +162,9 @@ func TestNewListValueFrom(t *testing.T) {
 			expectedDiags: diag.Diagnostics{
 				diag.NewAttributeErrorDiagnostic(
 					path.Empty(),
-					"List Type Validation Error",
-					"An unexpected error was encountered trying to validate an attribute value. This is always an error in the provider. Please report the following to the provider developer:\n\n"+
-						"expected List value, received tftypes.Value with value: tftypes.String<\"oops\">",
+					"Value Conversion Error",
+					"An unexpected error was encountered trying to convert the Terraform value. This is always an error in the provider. Please report the following to the provider developer:\n\n"+
+						"can't use tftypes.String<\"oops\"> as value of List with ElementType basetypes.StringType, can only use tftypes.String values",
```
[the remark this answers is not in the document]
bendbennett: Seems that this particular error message is being returned from the call to `(ListType).ValueFromTerraform`:

```go
	if !in.Type().Equal(l.TerraformType(ctx)) {
		return nil, fmt.Errorf("can't use %s as value of List with ElementType %T, can only use %s values", in.String(), l.ElementType(), l.ElementType().TerraformType(ctx).String())
	}
```

We could update these error messages, but it might be worth getting input from the rest of the team. Happy to open an issue for us to consider this further.

types/validation/doc.go
```diff
@@ -0,0 +1,5 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+// Package validation contains additional interfaces for base types.
+package validation
```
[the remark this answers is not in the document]
bendbennett: Sounds reasonable to me. Have moved `ValidatableParameter` to `function` package.

.changes/unreleased/NOTES-20240404-155606.yaml
```diff
@@ -0,0 +1,6 @@
+kind: NOTES
+body: 'attr/xattr: The `TypeWithValidate` interface has been deprecated in preference
+  of the `ValidateableAttribute` interface'
```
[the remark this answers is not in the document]
bendbennett: Have added a note to that effect.

attr/xattr/type.go
```diff
@@ -6,14 +6,17 @@ package xattr
...
 	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/path"
-	"github.com/hashicorp/terraform-plugin-go/tftypes"
 )
 
 // TypeWithValidate extends the attr.Type interface to include a Validate
 // method, used to bundle consistent validation logic with the Type.
+//
+// Deprecated: Use the ValidateableAttribute interface instead.
```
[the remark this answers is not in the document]
bendbennett: Have expanded this comment.

austinvalle (approved):
Great work Ben! 🚀 🎸

internal/reflect/interfaces.go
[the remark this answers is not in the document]
bendbennett: This is an interesting case! If we want to explicitly validate values that are being passed to `Result.Set` it seems that maybe neither `ValidateableParameter` nor `ValidateableAttribute` are appropriate interfaces, as the value is neither a parameter nor an attribute.

One option might be to consider a `Validateable` interface that operates solely on the value. Perhaps something along the following lines:

```go
type Validateable interface {
	Validate(context.Context, ValidateRequest, *ValidateResponse)
}

type ValidateRequest struct{}

type ValidateResponse struct {
	Error *FuncError
}
```

A custom type that was to be used for validating a provider-defined function parameter, _and_ the result from the function could be defined as follows:

```go
func (v CustomStringValue) ValidateParameter(ctx context.Context, req function.ValidateParameterRequest, resp *function.ValidateParameterResponse) {
	if v.IsNull() || v.IsUnknown() {
		return
	}

	if !v.isValid(v.ValueString()) {
		resp.Error = function.NewArgumentFuncError(
			req.Position,
			fmt.Sprintf("Invalid String Value: value %q length does not equal %d", v.ValueString(), 10))
	}
}

func (v CustomStringValue) Validate(ctx context.Context, req function.ValidateRequest, resp *function.ValidateResponse) {
	if v.IsNull() || v.IsUnknown() {
		return
	}

	if !v.isValid(v.ValueString()) {
		resp.Error = function.NewFuncError(
			fmt.Sprintf("Invalid String Value: value %q length does not equal %d", v.ValueString(), 10))
	}
}

func (v CustomStringValue) isValid(in string) bool {
	return len(in) == 10
}
```

The function definition could then be specified as:

```go
func (f ExampleFunction) Definition(ctx context.Context, req function.DefinitionRequest, resp *function.DefinitionResponse) {
			function.StringParameter{
				CustomType: CustomStringType{},
				Name:       "arg0",
			},
		},
		Return: function.StringReturn{
			CustomType: CustomStringType{},
		},
	}
}
```

The validation of the result value could then be handled by modifying the `ResultData.Set` method:

```go
func (d *ResultData) Set(ctx context.Context, value any) *FuncError {
	reflectValue, reflectDiags := fwreflect.FromValue(ctx, d.value.Type(ctx), value, path.Empty())

	funcErr := FuncErrorFromDiags(ctx, reflectDiags)

	if funcErr != nil {
		return funcErr
	}

	if v, ok := reflectValue.(Validateable); ok {
		resp := ValidateResponse{}

		v.Validate(ctx, ValidateRequest{}, &resp)

		if resp.Error != nil {
			return resp.Error
		}
	}

	d.value = reflectValue

	return nil
}
```

Thoughts?

internal/fromproto5/arguments_data.go
```diff
@@ -155,18 +209,49 @@ func ArgumentsData(ctx context.Context, arguments []*tfprotov5.DynamicValue, def
...
+
+	if position >= len(d.Parameters) {
+		return nil, function.NewArgumentFuncError(
+			pos,
+			"Invalid Parameter Position for Definition: "+
+				"When determining the parameter for the given argument position, an invalid value was given. "+
+				"This is always an issue in the provider code and should be reported to the provider developers.\n\n"+
+				fmt.Sprintf("Max argument position: %d\n", len(d.Parameters)-1)+
+				fmt.Sprintf("Given position: %d", position),
+		)
```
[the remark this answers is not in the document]
bendbennett: D'oh! Thanks. Removed

internal/fromproto5/arguments_data.go
```diff
@@ -54,73 +50,131 @@ func ArgumentsData(ctx context.Context, arguments []*tfprotov5.DynamicValue, def
...
+			)
+		case position >= len(definition.Parameters):
+			return function.NewArgumentsData(nil), function.NewArgumentFuncError(
+				pos,
+				"Invalid Parameter Position for Definition: "+
+					"When determining the parameter for the given argument position, an invalid value was given. "+
+					"This is always an issue in the provider code and should be reported to the provider developers.\n\n"+
+					fmt.Sprintf("Max argument position: %d\n", len(definition.Parameters)-1)+
+					fmt.Sprintf("Given position: %d", position),
+			)
```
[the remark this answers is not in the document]
bendbennett: Good call. Looks like both of these cases are handled by the previous inspection of arguments and definition parameters:

```go
	if definition.VariadicParameter == nil && len(arguments) != len(definition.Parameters) {
		/* ... */
	}

	if len(arguments) < len(definition.Parameters) {
		/* ... */
	}
```

internal/reflect/interfaces.go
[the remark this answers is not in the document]
austinvalle: In the case of something like `Validateable`, I'd assume it would end up just returning a plain `error` 🤔. I'm not 100% sure adding a new interface just for validation in reflection is worth it, but I think the general problem has been shown here.

I.e. the `ValidateableAttribute` interface is being used for more than just attributes. Perhaps something that we can't avoid? Something we don't necessarily need to avoid? 🤷🏻 

Would be interested to hear the rest of the teams thoughts on this, although I don't think it's really a showstopper

SBGoods (approved):
Great job, Ben!

internal/reflect/interfaces.go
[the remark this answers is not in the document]
bendbennett: Thanks for the feedback. Ok, so I've amended the Go doc for `ValidateableAttribute` as follows:

```go
// ValidateableAttribute defines an interface for validating an attribute value.
// The ValidateAttribute method is called implicitly by the framework when value
// types from Terraform are converted into framework types.
type ValidateableAttribute interface {
```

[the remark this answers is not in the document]
austinvalle: LGTM 👍🏻

Issue #894, opened 2023-12-18
function: Consider Parameter-Based Validators


bflad:
### Module version

```
v1.5.0 (not yet released)
```

### Use-cases

Provider developers implementing a schema definition for a data source, resource, or provider can define [attribute-based validators](https://developer.hashicorp.com/terraform/plugin/framework/validation#attribute-validation) to define validation rules that should be ran against configuration values. It might be nice to offer the same for provider-defined function definitions, since each parameter should theoretically be able to do something similar. The main "challenge" here is that the attribute-based validator interfaces and implementations are named relatively towards schemas, e.g. in their package import naming.

### Attempted Solutions

Manually implement value validation in provider-defined function logic (`Run` method), checking the value, adding any diagnostics, and returning early if necessary.

### Proposal

_Open to suggestions_

Double check whether the existing validator interfaces are acceptable to also serve as interfaces for parameter-based validation.

Add `Validators` fields to `function.Parameter` implementations. Implement parameter-based validation in `internal/fromproto5.ArgumentsData()` and `internal/fromproto6.ArgumentsData()` by looping through the `Validators`.

### References

- #893 (similar-ish; dealing with validation as well)

PR #971, opened 2024-03-29
Add parameter-based provider-defined function validation

SBGoods:
Closes: #894 

### Background
Provider-defined functions can accept parameters, or arguments as input. There is an opportunity to provide validation of such parameters in an analogous manner to the validation of values supplied in configuration for attributes, by implementing parameter-based and type-based validation for provider-defined function parameters.

This PR is concerned with the addition of parameter-based validation, which enables provider developers to define custom validators on framework-provided type parameters or on custom parameters.

### `<Type>Validator` Interfaces
This PR adds various `<Type>Validator` interfaces for custom validators to implement for parameter-based validation
```go
type BoolValidator interface {
	Validate(context.Context, BoolRequest, *BoolResponse)
}

type BoolRequest struct {
	ArgumentPosition int64
	Value types.Bool
}

type BoolResponse struct {
	Error *FuncError
}
```

### `ParameterWith<Type>Validators` Interfaces
This PR adds various `ParameterWith<Type>Validators` interfaces for custom parameter types to implement to enable parameter-based validation
```go
type ParameterWithBoolValidators interface {
	Parameter

	BoolValidators() []BoolValidator
}
```


internal/fromproto6/arguments_data.go
```diff
@@ -159,6 +160,289 @@ func ArgumentsData(ctx context.Context, arguments []*tfprotov6.DynamicValue, def
 			}
 		}
 
+		switch parameterWithValidators := parameter.(type) {
+		case function.ParameterWithBoolValidators:
+			for _, functionValidator := range parameterWithValidators.BoolValidators() {
+				boolVal, ok := attrValue.(types.Bool)
```
bendbennett, line 166: Should we be using an assertion for `attrValue.(basetypes.BoolValuable)` here to accomodate the possibility that the validators have been declared on a custom type? If so, similar question for further `attrValue` type assertions in this `switch` block. 

If the `Valuable` interfaces are to be used we'll need to obtain the `BoolValue` to pass into the `BoolRequest`:

```go
bVal, diags := boolVal.ToBoolValue(ctx)
...
```

internal/fromproto6/arguments_data.go
```diff
@@ -159,6 +160,289 @@ func ArgumentsData(ctx context.Context, arguments []*tfprotov6.DynamicValue, def
...
+					ArgumentPosition: pos,
+					Value:            boolVal,
+				}
+				resp := &function.BoolResponse{}
+				functionValidator.Validate(ctx, req, resp)
+				if resp.Error != nil {
+					funcError = function.ConcatFuncErrors(funcError, function.NewArgumentFuncError(
+						pos,
+						resp.Error.Error(),
+					))
```
bendbennett, line 189: Should we have a `continue` statement here so that we only append to the `argumentValues` slice when no error is raised during validation? Similar question for each `case` statement within this `switch`.

function/bool_parameter.go
```diff
@@ -70,6 +71,14 @@ type BoolParameter struct {
 	// alphabetical character and followed by alphanumeric or underscore
 	// characters.
 	Name string
+
+	// Validators is a list of bool validators that should be applied to the
+	// parameter.
+	Validators []BoolValidator
+}
+
```
bendbennett, line 79:
```suggestion
// BoolValidators returns the Validators field value.
```

Or something along those lines for the Go docs.

function/bool_parameter_validator.go
```diff
@@ -0,0 +1,33 @@
...
+package function
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+// BoolValidator is a function validator for types.Bool parameters.
+type BoolValidator interface {
```
bendbennett, line 13: The file name, `bool_parameter_validator.go` makes the intent clear, but I'm wondering whether more explicit naming might be warranted given that the `<Type>Validator`, and request and response types are all in the `function` package. Perhaps `BoolParameterValidator`, `BoolParameterValidatorRequest`, `BoolParameterValidatorResponse`?

bendbennett (commented):
Looks great!
Have left a couple of comments.
Do you think the website docs should be modified to indicate that parameter-based validation is available?

internal/fromproto5/arguments_data.go
```diff
@@ -159,6 +160,289 @@ func ArgumentsData(ctx context.Context, arguments []*tfprotov5.DynamicValue, def
 			}
 		}
 
+		switch parameterWithValidators := parameter.(type) {
+		case function.ParameterWithBoolValidators:
+			for _, functionValidator := range parameterWithValidators.BoolValidators() {
+				boolVal, ok := attrValue.(types.Bool)
```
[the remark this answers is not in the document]
SBGoods: Good catch! I've added the assertions and corresponding tests for custom types.

bendbennett (approved):
LGTM 🚀 
Nice job!

function/bool_parameter_validator.go
```diff
@@ -0,0 +1,33 @@
...
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+// BoolParameterValidator is a function validator for types.Bool parameters.
+type BoolParameterValidator interface {
+
+	// Validate performs the validation.
+	Validate(context.Context, BoolParameterValidatorRequest, *BoolParameterValidatorResponse)
```
bflad, line 16: Discussed more in the RFC, but I think it is important to use more explicit method names inside these interfaces:

* Adding the type name, so a parameter validator can support multiple types (refer to https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework-validators/internal/schemavalidator which apply to any type)
* Adding the what-it-applies-to, so existing attribute validators can also implement parameter validations (refer to validators such as https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework-validators@v0.12.0/stringvalidator#LengthAtLeast which is something I would anticipate that provider developers would want to use in parameters now). This prevents developers from needing to implement and consumers from needing to discover two things that essentially do the same thing. They'd still get a compilation error if they tried to use a validator that didn't implement the interface. (We need to slightly adjust terraform-plugin-framework-validators Go module implementations to return the underlying implementation type in their exported functions, not the interface type, but hopefully you get the idea).

Using either `ValidateParameterBool` or `ValidateBoolParameter` is reasonable to me.

internal/fromproto5/arguments_data_test.go
```diff
@@ -717,3 +720,2111 @@ func TestArgumentsData(t *testing.T) {
...
+
+func createObjectValue(attributeTypes map[string]attr.Type, attributes map[string]attr.Value) basetypes.ObjectValue {
+	object, _ := basetypes.NewObjectValue(attributeTypes, attributes)
+	return object
+}
+
+func createSetValue(elementType attr.Type, elements []attr.Value) basetypes.SetValue {
+	list, _ := basetypes.NewSetValue(elementType, elements)
+	return list
+}
```
austinvalle, line 2825: Can we use the `New{Type}ValueMust` equivalents in-place of these? (similar comment for the other `fromproto6` test)

website/docs/plugin/framework/validation.mdx
```diff
@@ -225,6 +225,213 @@ func Int64IsGreaterThan(expressions ...path.Expression) validator.Int64 {
...
+
+        return
+    }
+}
+
+func (v computeInstanceIdentifierValue) isValid(in string) bool {
+    return strings.HasPrefix(in, "instance-")
+}
+```
+
```
austinvalle, line 434: I think this section got duplicated, perhaps during a merge 🙂 
```suggestion
```

internal/fromproto6/arguments_data_test.go
austinvalle, line 1: Great test coverage here! 🥳

austinvalle (approved):
LGTM, just two minor comments :shipit:

internal/fromproto5/arguments_data_test.go
SBGoods: Thanks for pointing this out, I didn't know about the `New{Type}ValueMust` functions.

Issue #597, opened 2022-12-20
types: Consider Method for Checking Whether Collection Values are Fully Known


bflad:
### Module version

```
v1.0.1
```

### Use-cases

Configuration validation logic generally needs to skip over any potentially unknown values. The `IsUnknown() bool` method checks whether the value itself is unknown, but in the case of collection types such as `List`, `Map`, `Object`, and `Set` there may be underlying attributes/elements which are unknown. For validators such as the terraform-plugin-framework-validators [`listvalidator.UniqueValues()`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework-validators@v0.9.0/listvalidator#UniqueValues), it tries to detect duplicate values in the list based on value type equality. If there are unknown values from differing sources, it may unexpectedly trip up the equality checking since there is no differentiation in the type system between unknown values. For example:

```terraform
resource "examplecloud_thing" "example" {
  count = 2
}

resource "examplecloud_widget" "example" {
  thing_ids = [for id in examplecloud_thing[*].id: id] # differing unknown strings
}
```

### Proposal

Upstream in terraform-plugin-go, the [`tftypes.Value.IsFullyKnown() bool` method](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-go@v0.14.2/tftypes#Value.IsFullyKnown) provides additional recursive checks about whether the value contains unknown values. This module could provide something similar for all the framework types or at least the collection types. The implementation may need to lean on using the `*Valuable` interfaces to handle custom types.

PR #980, opened 2024-04-16
Extend `attr.Value` interface to support `IsFullyNullableKnown()`

magodo:
This new method is similar to the `Value.IsFullyKnown()` that is available in the `github.com/hashicorp/terraform-plugin-go/tftypes`.

The difference here is that in `tftypes`, each value can only has two states: a concrete value (including `nil`) or "unknown". While in the fw, each value can has three states: null, unknown and known. This is why the method name is chose so (as I can't figure out another better name, as `IsFullyNotKnown` or `IsPartiallyUnknown` are ambiguous than the current one, IMO).

The reason for introducing this method is to allow provider developers to check the state of an aggregate value during the `ModifyPlan`, where the code might stop processing that property if its value contains any unknwon value. Currently, the developer has two solutions:

- Convert the `ToTerraformValue()` to convert the `attr.Value` to `tftypes.Value`, then call its `IsFullyKnown()`. This works fine (and it is also used in the FW itself somewhere), while it is a bit over kill to do the conversion where the intent is only to check the whole (un)known-ness.
- Self implement the `IsFullyKnown()` for the `attr.Value`, similar to:

    ```go
    func IsFullyKnown(val attr.Value) bool {
            if val == nil {
                    return true
            }
            if val.IsUnknown() {
                    return false
            }
            switch v := val.(type) {
            case types.Dynamic:
                    return IsFullyKnown(v.UnderlyingValue())
            case types.List:
                    for _, e := range v.Elements() {
                            if !IsFullyKnown(e) {
                                    return false
                            }
                    }
                    return true
            case types.Set:
                    for _, e := range v.Elements() {
                            if !IsFullyKnown(e) {
                                    return false
                            }
                    }
                    return true
            case types.Tuple:
                    for _, e := range v.Elements() {
                            if !IsFullyKnown(e) {
                                    return false
                            }
                    }
                    return true
            case types.Map:
                    for _, e := range v.Elements() {
                            if !IsFullyKnown(e) {
                                    return false
                            }
                    }
                    return true
            case types.Object:
                    for _, e := range v.Attributes() {
                            if !IsFullyKnown(e) {
                                    return false
                            }
                    }
                    return true
            default:
                    return true
            }
    }
    ```

This PR tries to put this common logic to the FW so that more developers can save the run/develop time effort for the same purpose. I chose to extend the `attr.Value` interface, instead of introducing a helper method in the `attr` package, as a random choice. If the latter looks better, then I can rework this PR.

### DIFF
--- a/attr/value.go
+++ b/attr/value.go
@@ -57,8 +57,15 @@ type Value interface {
 	// IsNull returns true if the Value is not set, or is explicitly set to null.
 	IsNull() bool
 
-	// IsUnknown returns true if the value is not yet known.
+	// IsUnKnown returns true if the value is not yet known.
+	// If the value is an aggregate type, only the top level of the aggregate type
+	// is checked; elements and attributes are not checked.
 	IsUnknown() bool
+
+	// IsFullyNullableKnown returns true if the value is nullable known. If the value
+	// is an aggregate type, IsFullyNullableKnown only returns true if all elements
+	// and attributes are nullable known, as well.
+	IsFullyNullableKnown() bool
 
 	// String returns a summary representation of either the underlying Value,
 	// or UnknownValueString (`<unknown>`) when IsUnknown() returns true,
--- a/internal/testing/testtypes/bool.go
+++ b/internal/testing/testtypes/bool.go
@@ -116,6 +116,10 @@ func (b Bool) IsUnknown() bool {
 	return b.Bool.IsUnknown()
 }
 
+func (b Bool) IsFullyNullableKnown() bool {
+	return b.Bool.IsFullyNullableKnown()
+}
+
 func (b Bool) String() string {
 	return b.Bool.String()
 }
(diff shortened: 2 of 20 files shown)


bflad:
Related feature request issue: https://github.com/hashicorp/terraform-plugin-framework/issues/597

For any reviewers, please note that the current `attr.Value` interface change is a breaking change since it would force all existing custom type implementations to need to include the new method. I'm adding the GitHub label to call this out. @magodo I'm not providing a full review, but it might be good to update the proposed implementation to avoid the breaking change so it potentially can get into any release, rather than needing to wait for a future major version of the Go module. If you would like to discuss available options there, please reach out.

magodo:
@bflad The other possible option I can think of is the 2nd one I mentioned above, which won't handle the custom types. I don't know how can we avoid breaking change while still taking custom types into consideration at this moment..

PR #987, opened 2024-04-22
Fix `PlanModifiers` usage in code-snippet

kangasta:
The example seems to be missing `[]` in `PlanModifiers` field.


austinvalle (approved):
Thanks for the fix @kangasta! I see a couple other similar mistakes in other docs, will fix in a follow-up 👍🏻

PR #991, opened 2024-04-23
function: Missing underlying type validation and refactor parameter name validation

### LINKED ISSUE #965: function: Add validation to prevent collection parameters/return definitions from having missing element types
### Module version

```
v1.6.0
```

### Use-cases

#931 will introduce a new internal interface `fwfunction.ParameterWithValidateImplementation` and `fwfunction.ReturnWithValidateImplementation` that will enable us to implement logic for validating function parameter and return definitions.

We should introduce validation on all collection parameter and return definitions to ensure that element types are populated.

```go
        function.ListParameter{
            Name: "list",
            // Missing element type!
        },
    },
    Return: function.ListReturn{
        // Missing element type!
    },
}
```

### Proposal

Implementation could look similar to collection attributes, such as `ListAttribute`: https://github.com/hashicorp/terraform-plugin-framework/blob/f86c88c9700c4bd03743b32382b20694577d71a7/datasource/schema/list_attribute.go#L210-L212

We should add this to:
- `ListParameter`
- `ListReturn`
- `MapParameter`
- `MapReturn`
- `SetParameter`
- `SetReturn`

### References

- #931

### LINKED ISSUE #967: function: Refactor name validation to utilize new `internal/fwfunction` parameter validation interfaces
### Module version

```
v1.6.0
```

### Proposal

Following #964 and the `v1.7.0` release, we should refactor the parameter `Name` required validation to utilize the `internal/fwfunction` interfaces for validating parameters.

> Perhaps can be added alongside #965

bflad:
Closes #965
Closes #967

The framework will now raise implementation errors if a function parameter or return definition is missing underlying type information (e.g. collection element type or object attribute type).

### DIFF
--- a/internal/fwtype/missing_underlying_type_validation_test.go
+++ b/internal/fwtype/missing_underlying_type_validation_test.go
@@ -1,0 +1,1611 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package fwtype_test
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/internal/fwtype"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+func TestContainsMissingUnderlyingType(t *testing.T) {
+	t.Parallel()
+
+	testCases := map[string]struct {
+		attrTyp  attr.Type
+		expected bool
+	}{
+		"nil": {
+			attrTyp:  nil,
+			expected: true,
+		},
+		"bool": {
+			attrTyp:  types.BoolType,
+			expected: false,
+		},
+		"dynamic": {
+			attrTyp:  types.DynamicType,
+			expected: false,
+		},
+		"float64": {
+			attrTyp:  types.Float64Type,
+			expected: false,
+		},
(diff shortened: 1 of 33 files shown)


internal/fwtype/missing_underlying_type_validation_test.go
```diff
@@ -0,0 +1,1611 @@
...
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/internal/fwtype"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+func TestContainsMissingUnderlyingType(t *testing.T) {
```
austinvalle, line 15: nit: Do we want to add a test here with a custom type in a collection/object/etc?

austinvalle (approved):
Great PR! Just one nit 🚀

internal/fwtype/missing_underlying_type_validation_test.go
bflad: Is https://github.com/hashicorp/terraform-plugin-framework/pull/991/commits/584f83db2e23e9ec145d9377391836f9dbad583f what you are looking for?

austinvalle: Yeah that looks great!

PR #992, opened 2024-04-25
function: Mention that return documentation is expected in definition documentation

bflad:
Followup from an out-of-band conversation. The website documentation for returns already mentions this, but it was missing from the Go package documentation.


bendbennett (approved):
LGTM 🚀

PR #999, opened 2024-05-07
all: Add support for Manual Deferred Actions

SBGoods:
Ref: https://github.com/hashicorp/terraform-plugin-go/pull/403

This PR implements manual deferred action support for ReadResource, ReadDataSource, ImportResourceState, and PlanResourceChange RPCs. It adds new `resource.DeferredResponse` and `datasource.DeferredResponse` structs so that providers can indicate a deferral to the Terraform client.

Deferred actions is an experimental feature introduced in [Terraform v1.9.0-alpha20240404](https://github.com/hashicorp/terraform/releases/tag/v1.9.0-alpha20240404) and is only available in experimental Terraform builds.

A follow up PR will implement automatic deferred action support for `PROVIDER_CONFIG_UNKNOWN` during the ConfigureProvider RPC.


.changes/unreleased/FEATURES-20240508-105105.yaml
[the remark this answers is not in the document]
austinvalle: Preserving the original comment, but we're also chatting about this in the SDKv2 PR: https://github.com/hashicorp/terraform-plugin-sdk/pull/1335#discussion_r1595449264

Whatever we decide to do, we should make sure the messaging is the same between SDKv2 and Framework 👍🏻

internal/fwserver/server_importresourcestate.go
```diff
@@ -128,6 +143,10 @@ func (s *Server) ImportResourceState(ctx context.Context, req *ImportResourceSta
 		private.Provider = importResp.Private
 	}
 
+	if importResp.DeferredResponse != nil {
+		resp.Deferred = importResp.DeferredResponse
+	}
```
austinvalle, line 148: super nit: I believe since this is a pointer we could just always set the value (if it's nil, will just continue to be nil)

internal/toproto5/importresourcestate.go
```diff
@@ -33,5 +34,11 @@ func ImportResourceStateResponse(ctx context.Context, fw *fwserver.ImportResourc
 		proto5.ImportedResources = append(proto5.ImportedResources, proto5ImportedResource)
 	}
 
+	if fw.Deferred != nil {
+		proto5.Deferred = &tfprotov5.Deferred{
+			Reason: tfprotov5.DeferredReason(fw.Deferred.Reason),
+		}
+	}
```
austinvalle, line 41: nit: We can probably break this into it's own file/function for `Deferred` in `toproto{5/6}`. Makes it easier to add more fields later 👍🏻 

Similar to `terraform-plugin-go`: https://github.com/hashicorp/terraform-plugin-go/blob/main/tfprotov5/internal/toproto/deferred.go

resource/import_state.go
```diff
@@ -44,6 +57,13 @@ type ImportStateResponse struct {
 	// This field is not pre-populated as there is no pre-existing private state
 	// data during the resource's Import operation.
 	Private *privatestate.ProviderData
+
+	// DeferredResponse indicates that Terraform should defer
+	// importing this resource.
+	//
+	// This field can only be set if
+	// `(resource.ImportStateRequest.ImportStateClientCapabilities).DeferralAllowed` is true.
```
austinvalle, line 65:
```suggestion
	// `(resource.ImportStateRequest).ClientCapabilities.DeferralAllowed` is true.
```

resource/modify_plan.go
```diff
@@ -65,4 +78,11 @@ type ModifyPlanResponse struct {
 	// indicates a successful plan modification with no warnings or errors
 	// generated.
 	Diagnostics diag.Diagnostics
+
+	// DeferredResponse indicates that Terraform should defer
+	// importing this resource.
+	//
+	// This field can only be set if
+	// `(resource.ModifyPlanRequest.ModifyPlanClientCapabilities).DeferralAllowed` is true.
```
austinvalle, line 86:
```suggestion
	// `(resource.ModifyPlanRequest).ClientCapabilities.DeferralAllowed` is true.
```

resource/read.go
```diff
@@ -50,4 +63,11 @@ type ReadResponse struct {
 	// resource. An empty slice indicates a successful operation with no
 	// warnings or errors generated.
 	Diagnostics diag.Diagnostics
+
+	// DeferredResponse indicates that Terraform should defer
+	// importing this resource.
+	//
+	// This field can only be set if
+	// `(resource.ReadRequest.ReadClientCapabilities).DeferralAllowed` is true.
```
austinvalle, line 71:
```suggestion
	// `(resource.ReadRequest).ClientCapabilities.DeferralAllowed` is true.
```

austinvalle (commented):
Changes look great! I have a couple questions about default values + an on-going discussion about experimental wording.

We'll need to figure out internally how we want to release this work (wait for automatic deferral work?)

.changes/unreleased/FEATURES-20240508-105105.yaml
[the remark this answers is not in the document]
austinvalle: And the chatting has commenced 😆 , what do you think about adding a similar changeset to this: https://github.com/hashicorp/terraform-plugin-sdk/pull/1335/commits/207c950e468b7ff8eeae648afaa35f88cee7c5b9

datasource/read.go
```diff
@@ -22,6 +31,10 @@ type ReadRequest struct {
 
 	// ProviderMeta is metadata from the provider_meta block of the module.
 	ProviderMeta tfsdk.Config
+
+	// ClientCapabilities defines optionally supported protocol features for the
+	// ReadDataSource RPC, such as forward-compatible Terraform behavior changes.
+	ClientCapabilities *ReadClientCapabilities
```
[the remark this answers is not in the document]
bflad: (I'm okay with simplifying the _request_ side of this for consumers if we want -- we had discussed out of band the implications of pointers for the response side and a preference for explicitness, but for this particular situation, there's not a semantic difference whether or not the client capabilities were actually announced since all fields should be disabled by default in the protocol when they are added.)

datasource/read.go
```diff
@@ -37,4 +50,11 @@ type ReadResponse struct {
 	// source. An empty slice indicates a successful operation with no
 	// warnings or errors generated.
 	Diagnostics diag.Diagnostics
+
+	// DeferredResponse indicates that Terraform should defer
+	// importing this resource.
+	//
+	// This field can only be set if
+	// `(datasource.ReadRequest.ReadClientCapabilities).DeferralAllowed` is true.
+	DeferredResponse *DeferredResponse
```
bflad, line 59: Bikeshed: Do we want to name the field `Deferred` (acknowledging that the `Response` naming is already present in the containing type name) or both the name and type `Deferred` here (matching protocol/terraform-plugin-go)? 

Same for resource package.

internal/fromproto6/readdatasource_test.go
```diff
@@ -46,6 +47,8 @@ func TestReadDataSourceRequest(t *testing.T) {
 		},
 	}
 
+	testClientCapabilities := tfprotov6.ReadDataSourceClientCapabilities{DeferralAllowed: true}
```
bflad, line 50: Nit: For single usage, maybe easier for future travelers to not add the variable indirection

internal/fwserver/server_readdatasource.go
```diff
@@ -75,17 +77,30 @@ func (s *Server) ReadDataSource(ctx context.Context, req *ReadDataSourceRequest,
 		readReq.ProviderMeta = *req.ProviderMeta
 	}
 
+	readReq.ClientCapabilities = req.ClientCapabilities
```
bflad, line 80: Nit: `ClientCapabilities` could be set in the `readReq` instantiation directly since its not conditional -- that should be the typical pattern in these request/response mapping functions. 👍 

Similar elsewhere.

bflad (commented):
Drive-by minor comments since @austinvalle is already reviewing -- please let me know if you have any questions. 😄

datasource/read.go
austinvalle: Ah, good call. I think it'd be fine to drop the `Response` on both the type and field name since it's not really adding any additional context.

I will update the SDKv2 PR to also follow this convention 👍🏻

datasource/read.go
```diff
@@ -37,4 +53,14 @@ type ReadResponse struct {
 	// source. An empty slice indicates a successful operation with no
 	// warnings or errors generated.
 	Diagnostics diag.Diagnostics
+
+	// Deferred indicates that Terraform should defer
+	// reading this data source.
```
bflad, line 58: Nit: I'm not sure if we may want to clarify this against the existing deferred behavior that Terraform will automatically do if you do have a data source that references an unknown value, e.g.

```terraform
data "examplecloud_thing" "example" {
  example = # some unknown value reference
}
```

Which causes the Terraform plan to show a read on apply action for the data source. Maybe one way to do this is to talk about deferral until the next Terraform apply operation?

```suggestion
	// Deferred indicates that Terraform should defer reading this
	// data source until a followup apply operation.
```

internal/fwserver/server_importresourcestate_test.go
```diff
@@ -299,6 +303,42 @@ func TestServerImportResourceState(t *testing.T) {
 				},
 			},
 		},
+		"response-importedresources-deferral": {
```
bflad, line 306: Nit: It would be good to have unit testing cases here and for the other RPCs for request client capabilities -- both the zero-value and when enabled 👍

internal/toproto5/importresourcestate.go
bflad: It still seems like this feedback might be valid, even if that means having two separate `DataSourceDeferred()` and `ResourceDeferred()`, e.g.

```go
func ResourceDeferred(fw *resource.Deferred) *tfprotov5.Deferred {
  if fw == nil {
    return nil
  }

  return &tfprotov5.Deferred{
    Reason: tfprotov5.DeferredReason(fw.Deferred.Reason)
  }
}
```

As Austin mentions, this continues the pattern of having shared logic for the terraform-plugin-go type conversions.

resource/modify_plan.go
```diff
@@ -65,4 +81,14 @@ type ModifyPlanResponse struct {
 	// indicates a successful plan modification with no warnings or errors
 	// generated.
 	Diagnostics diag.Diagnostics
+
+	// Deferred indicates that Terraform should defer
+	// importing this resource.
```
bflad, line 86: Similar suggestion here:

```suggestion
	// Deferred indicates that Terraform should defer importing this
	// resource until a followup apply operation.
```

resource/read.go
```diff
@@ -50,4 +66,14 @@ type ReadResponse struct {
 	// resource. An empty slice indicates a successful operation with no
 	// warnings or errors generated.
 	Diagnostics diag.Diagnostics
+
+	// Deferred indicates that Terraform should defer
+	// importing this resource.
```
bflad, line 71: Read 😄 and similar suggestion:

```suggestion
	// Deferred indicates that Terraform should defer refreshing this
	// resource until a followup plan operation.
```

bflad (commented):
Overall looks good to me, nice work. I'll let @austinvalle give the final 👍

internal/fromproto5/client_capabilities.go
```diff
@@ -0,0 +1,59 @@
...
+	if in == nil {
+		return nil
+	}
+
+	resp := &resource.ImportStateClientCapabilities{
+		DeferralAllowed: in.DeferralAllowed,
+	}
+
+	return resp
+}
```
austinvalle, line 59: This feels like a good place to apply the defaulting, especially because we can be explicit about it here, rather than implicitly letting Go do the defaulting. For example:

```go
func ReadDataSourceClientCapabilities(in *tfprotov5.ReadDataSourceClientCapabilities) datasource.ReadClientCapabilities {
	if in == nil {
		// Client did not indicate any supported capabilities
		return datasource.ReadClientCapabilities{
				DeferralAllowed: false,
		}
	}

	return datasource.ReadClientCapabilities{
		DeferralAllowed: in.DeferralAllowed,
	}
}
```

Then the rest of the Framework logic and structs don't need to use pointers or worry about doing nil checks. (similar thoughts for `fromproto6`)

internal/fwserver/server_importresourcestate.go
```diff
@@ -35,13 +35,16 @@ type ImportResourceStateRequest struct {
 	// TypeName is the resource type name, which is necessary for populating
 	// the ImportedResource TypeName of the ImportResourceStateResponse.
 	TypeName string
+
+	ClientCapabilities *resource.ImportStateClientCapabilities
```
austinvalle, line 39: Similar to feedback above, if we do the defaulting earlier in `fromproto`, this can just be:
```go
ClientCapabilities resource.ImportStateClientCapabilities
```

internal/fwserver/server_importresourcestate_test.go
```diff
@@ -95,6 +99,67 @@ func TestServerImportResourceState(t *testing.T) {
...
+			expectedResponse: &fwserver.ImportResourceStateResponse{
+				ImportedResources: []fwserver.ImportedResource{
+					{
+						State:    *testState,
+						TypeName: "test_resource",
+						Private:  testEmptyPrivate,
+					},
+				},
+			},
+		},
```
austinvalle, line 162: Similar to the feedback above, if we move the defaulting up, we can test this `unset` scenario in `fromproto` (similar thought for other `fwserver` tests for this)

austinvalle (commented):
I have one final thought about the defaulting and perhaps moving it up to `fromproto`

internal/fromproto5/client_capabilities.go
SBGoods: That's a good point, I was just thinking about this since I moved the defaulting to `ConfigureProviderClientCapabilities()` in the automatic provider deferred action implementation since `server_configureprovider.go` doesn't have an equivalent `*fwserver.ConfigureProviderRequest` to return a pointer to.

internal/toproto6/deferred.go
```diff
@@ -0,0 +1,29 @@
...
+}
+
+func ResourceDeferred(fw *resource.Deferred) *tfprotov6.Deferred {
+	if fw == nil {
+		return nil
+	}
+	return &tfprotov6.Deferred{
+		Reason: tfprotov6.DeferredReason(fw.Reason),
+	}
+}
```
austinvalle, line 29: I don't think these are hooked up to the relevant functions in:
- `internal/toproto6/importresourcestate.go`
- `internal/toproto6/planresourcechange.go`
- `internal/toproto6/readdatasource.go`
- `internal/toproto6/readresource.go`

Also, TIL that `golangci-lint`'s `unused` check doesn't apply for exported members in an `internal` package 😆 - https://github.com/dominikh/go-tools/issues/1406

PR #1002, opened 2024-05-16
all: Add automatic deferred action support for unknown provider configuration

SBGoods:
Ref: https://github.com/hashicorp/terraform-plugin-go/pull/403

This PR implements deferred action support for `PROVIDER_CONFIG_UNKNOWN` during the `ConfigureProvider` RPC. This functionality will allow a provider to defer all resources and data sources associated with the framework provider during the `ReadResource`, `ReadDataSource`, `ImportResourceState`, and `PlanResourceChange` RPCs.

Deferred actions is an experimental feature introduced in [`Terraform v1.9.0-alpha20240404`](https://github.com/hashicorp/terraform/releases/tag/v1.9.0-alpha20240404) and is only available in experimental Terraform builds.


internal/fwserver/server_readdatasource.go
```diff
@@ -38,6 +38,13 @@ func (s *Server) ReadDataSource(ctx context.Context, req *ReadDataSourceRequest,
 		return
 	}
 
+	//TODO: add logging for replacing deferred reason
+	if s.deferred != nil {
+		resp.Deferred = &datasource.Deferred{
+			Reason: resp.Deferred.Reason,
+		}
+	}
```
austinvalle, line 46: See comment in `ReadResource`, same idea applies.

SDKv2 logic: https://github.com/hashicorp/terraform-plugin-sdk/blob/e90e8a2497cb1bced1d5614e65c904453244d18c/helper/schema/grpc_provider.go#L1374-L1398

Your logic for constructing an unknown value will be a little different since it's not using `cty`, but the idea is the same, we'll set `State` to an unknown value for the entire data source object.

internal/fwserver/server_importresourcestate.go
```diff
@@ -53,6 +53,13 @@ func (s *Server) ImportResourceState(ctx context.Context, req *ImportResourceSta
 		return
 	}
 
+	//TODO: add logging for replacing deferred reason
+	if s.deferred != nil {
+		resp.Deferred = &resource.Deferred{
+			Reason: resp.Deferred.Reason,
+		}
+	}
```
austinvalle, line 61: See comment in `ReadResource`, same idea applies.

SDKv2 logic: hhttps://github.com/hashicorp/terraform-plugin-sdk/blob/e90e8a2497cb1bced1d5614e65c904453244d18c/helper/schema/grpc_provider.go#L1223-L1263

Your logic for constructing an unknown value will be also be different here since it's not using `cty`.. It also will not need to do the error checking for type name. But the goal is to return a single unknown object state for `ImportedResources`

internal/fwserver/server_configureprovider.go
```diff
@@ -22,6 +22,18 @@ func (s *Server) ConfigureProvider(ctx context.Context, req *provider.ConfigureR
...
+			resp.Diagnostics.AddError("Invalid Deferred Provider Response",
+				"Provider configured a deferred response for all resources and data sources but the Terraform request "+
+					"did not indicate support for deferred actions. This is an issue with the provider and should be reported to the provider developers.")
+		} else {
+			logging.FrameworkDebug(ctx, "Provider has configured a deferred response, "+
+				"all associated resources and data sources will automatically return a deferred response.")
+		}
+	}
+
+	s.deferred = resp.Deferred
```
austinvalle, line 36: nit: We only need to set this in the `else` statement above

internal/fwserver/server_configureprovider.go
```diff
@@ -22,6 +22,18 @@ func (s *Server) ConfigureProvider(ctx context.Context, req *provider.ConfigureR
 
 	logging.FrameworkTrace(ctx, "Called provider defined Provider Configure")
 
+	if resp.Deferred != nil {
+		if !req.ClientCapabilities.DeferralAllowed {
+			resp.Diagnostics.AddError("Invalid Deferred Provider Response",
+				"Provider configured a deferred response for all resources and data sources but the Terraform request "+
+					"did not indicate support for deferred actions. This is an issue with the provider and should be reported to the provider developers.")
```
austinvalle, line 29: super nit: We could probably immediately return after this, since the return isn't right after this statement

internal/fwserver/server_planresourcechange.go
```diff
@@ -215,6 +216,15 @@ func (s *Server) PlanResourceChange(ctx context.Context, req *PlanResourceChange
...
 	}
 
+	// Skip plan modification for automatic deferrals
+	// unless ProviderDeferredBehavior.EnablePlanModification is true
+	if s.deferred != nil && req.ResourceBehavior.ProviderDeferred.EnablePlanModification == false {
+		logging.FrameworkDebug(ctx, "Provider has deferred response configured, automatically returning deferred response.")
+		resp.Deferred = &resource.Deferred{
+			Reason: resource.DeferredReason(s.deferred.Reason),
+		}
+	}
```
austinvalle, line 226: Same thoughts about returning early from the `ReadResource` comment.

This is the right idea, we'll just need to run the framework-defined logic only here and then immediately return. There's two ways to approach it:
1. How you did it here, where we would need to duplicate the "marking all computed nil as unknown" logic call up here, then immediately return. If `EnablePlanModification == true`, then it'll be set by your second `if s.deferred != nil` below. This matches how [SDKv2 does it](https://github.com/hashicorp/terraform-plugin-sdk/blob/e90e8a2497cb1bced1d5614e65c904453244d18c/helper/schema/grpc_provider.go#L777-L792)
2. Add if statements to all provider-defined logic calls, which could get cumbersome and we'd need to remember to keep propogating that if we ever add more provider-defined logic to plan modification.

I'm currently leaning towards option 1

Brian did mention that we could potentially run the schema-based plan modification, which may make option 2 look more appealing.

austinvalle (commented):
Left some comments, good work so far!

internal/fwserver/server_importresourcestate_test.go
```diff
@@ -280,7 +289,53 @@ func TestServerImportResourceState(t *testing.T) {
...
+						if req.ID != "test-id" {
+							resp.Diagnostics.AddError("unexpected req.ID value: %s", req.ID)
+						}
+
+						resp.Deferred = &resource.Deferred{
+							Reason: resource.DeferredReasonAbsentPrereq,
+						}
+
+						resource.ImportStatePassthroughID(ctx, path.Root("id"), req, resp)
+
```
austinvalle, line 321: This logic shouldn't be executed I believe. Maybe we instead just return a diagnostic in this logic to ensure it fails the test if called?

https://github.com/hashicorp/terraform-plugin-sdk/blob/4e590aed07ae4b94ea765b07a35ae3c14fc4284b/helper/schema/grpc_provider_test.go#L5091-L5095

internal/fwserver/server_readdatasource_test.go
```diff
@@ -236,7 +245,47 @@ func TestServerReadDataSource(t *testing.T) {
...
+							TestRequired types.String `tfsdk:"test_required"`
+						}
+
+						resp.Diagnostics.Append(req.Config.Get(ctx, &config)...)
+
+						resp.Deferred = &datasource.Deferred{Reason: datasource.DeferredReasonAbsentPrereq}
+
+						if config.TestRequired.ValueString() != "test-config-value" {
+							resp.Diagnostics.AddError("unexpected req.Config value: %s", config.TestRequired.ValueString())
+						}
```
austinvalle, line 278: Similar though to the import test, let's return a diagnostic here so the test fails if this is called

internal/fwserver/server_planresourcechange_test.go
```diff
@@ -3007,7 +3009,130 @@ func TestServerPlanResourceChange(t *testing.T) {
...
+							resp.Deferred = &resource.Deferred{Reason: resource.DeferredReasonAbsentPrereq}
+						}
+
+						var data testSchemaData
+
+						resp.Diagnostics.Append(req.Plan.Get(ctx, &data)...)
+
+						data.TestComputed = types.StringValue("test-plannedstate-value")
+
+						resp.Diagnostics.Append(resp.Plan.Set(ctx, &data)...)
```
austinvalle, line 3119: Similar though to the import test, let's return a diagnostic here so the test fails if this is called

internal/fwserver/server_readresource_test.go
```diff
@@ -339,7 +341,46 @@ func TestServerReadResource(t *testing.T) {
...
+							TestRequired types.String `tfsdk:"test_required"`
+						}
+
+						resp.Diagnostics.Append(req.State.Get(ctx, &data)...)
+
+						resp.Deferred = &resource.Deferred{Reason: resource.DeferredReasonAbsentPrereq}
+
+						if data.TestRequired.ValueString() != "test-currentstate-value" {
+							resp.Diagnostics.AddError("unexpected req.State value: %s", data.TestRequired.ValueString())
+						}
```
austinvalle, line 373: Similar though to the import test, let's return a diagnostic here so the test fails if this is called

resource/metadata.go
```diff
@@ -21,4 +21,29 @@ type MetadataResponse struct {
...
+	// ResourceBehavior is used to control framework-specific logic when
+	// interacting with this resource.
+	ResourceBehavior ResourceBehavior
+}
+
+// ResourceBehavior controls framework-specific logic when interacting
+// with a resource.
+type ResourceBehavior struct {
+	// ProviderDeferred enables provider-defined logic to be executed
+	// in the case of an automatic deferred response from provider configure.
```
austinvalle, line 34:
```suggestion
	// ProviderDeferred enables provider-defined logic to be executed
	// in the case of an automatic deferred response from provider configure.
	//
	// NOTE: This functionality is related to deferred action support, which is currently experimental and is subject
	// to change or break without warning. It is not protected by version compatibility guarantees.
```

We can also add the note here

internal/fwserver/server_planresourcechange.go
```diff
@@ -250,6 +251,16 @@ func (s *Server) PlanResourceChange(ctx context.Context, req *PlanResourceChange
...
 
+	// Skip resource-level ModifyPlan for automatic deferrals
+	// unless ProviderDeferredBehavior.EnablePlanModification is true
+	if s.deferred != nil && !req.ResourceBehavior.ProviderDeferred.EnablePlanModification {
+		logging.FrameworkDebug(ctx, "Provider has deferred response configured, automatically returning deferred response.")
+		resp.Deferred = &resource.Deferred{
+			Reason: resource.DeferredReason(s.deferred.Reason),
+		}
+		return
+	}
```
austinvalle, line 262: I'm interested what the potential impact might be with introducing the automatic deferral after all of these have run:
- Resource `Configure` method
- Provider-defined schema default plan modifiers
- Null computed attributes marked as unknown
- Provider-defined schema plan modifiers

I think this should be okay since we don't explicitly pass API client information here, but some of our documentation refers to plan modification as "online", like the blurb about validation: https://developer.hashicorp.com/terraform/plugin/framework/resources/validate-configuration

If we do go with this route, I think we should update the naming/documentation of `ResourceBehavior.EnablePlanModification` to be more explicit about "what" plan modification logic is being enabled/disabled. Might need to bikeshed a name like `EnableModifyPlan`, `EnableResourcePlanModification`, etc 🤔 

Any thoughts on this @bflad ?

austinvalle (commented):
Looking good! Some general notes and a comment about plan modification where we might want to adjust some of our docs/naming

internal/fwserver/server_planresourcechange.go
bflad: That is a fair point. Another option is doing similar as the RPCs and returning without doing _anything_ and would technically be the safest choice. I had hoped that since most of the list up there shouldn't be doing anything with using APIs, etc. we could improve the plan output for future rounds, but that could be changed later as an enhancement, rather than the reverse of taking functionality away being a breaking change. 

I guess I'll concede we should be safer here and return super early before anything runs. We can always change that decision later. I do worry about the null to unknown logic not running although maybe core will already do something like that in the plan output based on the configuration (something also worth checking with ReadDataSource).

bflad (commented):
Great work on this, @SBGoods 👍 A few notes, but overall pretty close. Will let @austinvalle give the final ✅

internal/fwserver/server_planresourcechange.go
```diff
@@ -288,7 +303,20 @@ func (s *Server) PlanResourceChange(ctx context.Context, req *PlanResourceChange
...
+			if modifyPlanResp.Deferred != nil {
+				logging.FrameworkDebug(ctx, fmt.Sprintf("Provider deferred response reason: %s replaced resource deferred response reason: %s",
+					s.deferred.Reason.String(), modifyPlanResp.Deferred.Reason.String()))
+			}
+			resp.Deferred = &resource.Deferred{
+				Reason: resource.DeferredReason(s.deferred.Reason),
+			}
+		} else {
+			resp.Deferred = modifyPlanResp.Deferred
+		}
```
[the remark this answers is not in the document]
austinvalle: number 2 sounds reasonable to me 👍🏻

internal/fwserver/server_readdatasource.go
```diff
@@ -38,6 +40,23 @@ func (s *Server) ReadDataSource(ctx context.Context, req *ReadDataSourceRequest,
...
+		logging.FrameworkDebug(ctx, "Provider has deferred response configured, automatically returning deferred response.",
+			map[string]interface{}{
+				logging.KeyDeferredReason: s.deferred.Reason.String(),
+			},
+		)
+		// Send an unknown value for the data source
+		resp.State = &tfsdk.State{
+			Raw:    tftypes.NewValue(req.DataSourceSchema.Type().TerraformType(ctx), tftypes.UnknownValue),
+			Schema: req.DataSourceSchema,
+		}
```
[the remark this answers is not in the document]
austinvalle: When I chatted with core about this, they weren't concerned about having configured values being in-place, so we went with the "full unknown object" for ease of implementation on our end.

If that changes in the future we should also come back and update SDKv2 with whatever logic is needed: https://github.com/hashicorp/terraform-plugin-sdk/blob/5fee97c1b78c352a1195194a12a9ad514508e3c6/helper/schema/grpc_provider.go#L1383-L1397

internal/fwserver/server_planresourcechange.go
austinvalle: For safe routes, would it be a bad idea to just duplicate the "marked computed nils as unknown" logic after checking for a deferral at the top of `PlanResourceChange`?

```go
modifiedPlan, err := tftypes.Transform(resp.PlannedState.Raw, MarkComputedNilsAsUnknown(ctx, req.Config.Raw, req.ResourceSchema))
```

I'm also cool with just coming back later and enhancing it once we have more info, using `ProposedNewState` like SDKv2: https://github.com/hashicorp/terraform-plugin-sdk/blob/5fee97c1b78c352a1195194a12a9ad514508e3c6/helper/schema/grpc_provider.go#L777-L792

bflad: Probably fine but I guess it might be good to evaluate the "current" behavior if the provider side just reflects that data without adding the extra logic as maybe its not needed. That evaluation does not need to happen immediately though, which is nice.

.changes/unreleased/FEATURES-20240520-180458.yaml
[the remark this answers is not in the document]
SBGoods: Since we are planning on releasing this with the manual deferred action work, this should be covered by: https://github.com/hashicorp/terraform-plugin-framework/blob/main/.changes/unreleased/NOTES-20240510-143136.yaml

internal/fwserver/server_planresourcechange.go
SBGoods: I've added a flag in [31f33f8](https://github.com/hashicorp/terraform-plugin-framework/pull/1002/commits/31f33f83f969b60cf83a3194165ce66965fa8491) to essentially skip everything except "Null computed attributes marked as unknown" logic. Unfortunately, there's a lot of prerequisites to calling `MarkComputedNilsAsUnknown()` that I felt was bloating the deferral check block, but if the flag logic is too complicated, I can go back to duplicating the logic in the deferral check.

austinvalle: My only reservation in adding a flag to all the other pieces of plan logic is just future maintenance/changes to `PlanResourceChange`, i.e. it's easier to describe just the operations that automatic deferred plans needs to run, rather than describe all the operations (and future operations) that shouldn't run.

If there are a ton of operations and such that need to run for the computed marking, maybe we should just return early with `ProposedNewState` (like [SDKv2](https://github.com/hashicorp/terraform-plugin-sdk/blob/5fee97c1b78c352a1195194a12a9ad514508e3c6/helper/schema/grpc_provider.go#L777-L792) is doing) for now and treat it as an enhancement once deferred actions/stacks is more in-use.

Plans with deferred actions aren't held to the consistency rules by Terraform between rounds, so this is mostly a UX problem

bflad (approved):
Just need to update the PlanResourceChange unit testing to cover deferred reason overrides, otherwise nice work and looks good to me 🚀

Issue #801, opened 2023-07-13
Add Int32 type


pgier:
### Module version

```
github.com/hashicorp/terraform-plugin-framework v1.3.2
```

### Use-cases
Any time you want to use a default int in golang.

### Attempted Solutions
Currently using the schema.Int64Attribute and related types.  I have to do type conversions in various places.

### Proposal
Add an `Int` type similar to Int64 which just returns a default golang `int`.

bflad:
Hi @pgier 👋 Thank you for raising this.

One thing that has come up occasionally before is specifically supporting a 32-bit integer type, but this is the first direct request for a Go-specific `int` type. One of the challenges with the Go `int` type is that it is defined by the CPU architecture running the code (or lower, such as running a 32-bit binary on 64-bit CPU), so supported integers may be 64-bit or 32-bit depending on where the provider was running. This can ultimately leading to practitioner confusion when there is an unexpected integer overflow when the API may happen to return integers larger than 32 bits.

Can you describe where/why you're running into these type conversions? Are you working with an API SDK that deals exclusively with the loosely-defined `int`/`*int` types? For example, would providing a way to create Int64Attribute value from a Go `int` help?

pgier:
I guess an int32 type would be fine.  My specific cases that I've run into are:
1. Writing code against an existing client library which uses in32 types.
2. Using things like `strconv.Atoi` where it would be nice to just pass a plain int type

For (1), I'm now using `ObjectValue.As` to do the type conversions and that seems to be working fine.
For (2), I should probably just use `strconv.FormatInt` or something else that accepts an int64.

If there is already a request for an int32 type, I think you can close this as a duplicate.

bflad:
> Writing code against an existing client library which uses in32 types.

Just to ensure we are on the same page, are those client library types the Go built-in `int32` type or just `int`? When we have thought about introducing something like `types.Int32`/`schema.Int32Attribute`, it would explicitly be the `int32` type to prevent any confusion about the value storage.

> If there is already a request for an int32 type, I think you can close this as a duplicate.

I don't believe there is a public one in this issue tracker at the moment, so if you're okay with that we can re-title this.

pgier:
> Just to ensure we are on the same page, are those client library types the Go built-in int32 type or just int?

This particular library is using specifically int32 types.

> if you're okay with that we can re-title this

Yes, that's fine with me

PR #1010, opened 2024-06-21
all: Add int32 type, attribute, and function support

SBGoods:
Closes: #801 

This PR introduces Int32 type support to `terraform-plugin-framework` with the addition of:
- Int32 type and value implementations: `types.Int32Type` and `types.Int32`
- Custom type interfaces:`basetypes.Int32Typable`, `basetypes.Int32Valuable` and `basetypes.Int32ValuableWithSemanticEquals`
- Attribute support: `schema.Int32Attribute` for resource, data source, and provider schemas
- Function support: `function.Int32Parameter` and `function.Int32Return`
- Int32 plan modifier interfaces and implementations (requires replace, use state if unknown, etc.)
- Int32 default interfaces and static implementation
- Int32 validation interface: `validator.Int32`


types/basetypes/int32_type_test.go
```diff
@@ -0,0 +1,76 @@
...
+			expectation: NewInt32Unknown(),
+		},
+		"null": {
+			input:       tftypes.NewValue(tftypes.Number, nil),
+			expectation: NewInt32Null(),
+		},
+		"wrongType": {
+			input:       tftypes.NewValue(tftypes.String, "oops"),
+			expectedErr: "can't unmarshal tftypes.String into *big.Float, expected *big.Float",
+		},
```
austinvalle, line 39: Since we have some logic around the bounds/validation of the value, we should add some additional tests for:
- `math.Int32Min` success
- `math.Int32Max` success
- `math.Int32Min-1` failure
- `math.Int32Max+1` failure
- Maybe a test that something like `tftypes.NewValue(tftypes.Number, 123.1)` returns an error

austinvalle (approved):
LGTM, left some suggestions in the PR 

I see references to float32 attributes that don't exist but it seems likely they'll go in the same release of `terraform-plugin-framework v1.10.0`

PR #1014, opened 2024-06-26
all: Add float32 type, attribute, and function support

SBGoods:
Relates: #1010 
This PR introduces Float32 type support to terraform-plugin-framework with the addition of:

- Float32 type and value implementations: `types.Float32Type` and `types.Float32`
- Custom type interfaces: `basetypes.Float32Typable`, `basetypes.Float32Valuable` and `basetypes.Float32ValuableWithSemanticEquals`
- Attribute support: `schema.Float32Attribute` for resource, data source, and provider schemas
- Function support: `function.Float32Parameter` and `function.Float32Return`
- Float32 plan modifier interfaces and implementations (requires replace, use state if unknown, etc.)
- Float32 default interfaces and static implementation
- Float32 validation interface: `validator.Float32`


types/basetypes/float32_type.go
```diff
@@ -0,0 +1,110 @@
...
+	err := in.As(&bigF)
+
+	if err != nil {
+		return nil, err
+	}
+
+	f, accuracy := bigF.Float32()
+	logging.FrameworkDebug(ctx, fmt.Sprintf("Float32Type ValueFromTerraform Float32 value %f, accuracy %d", f, accuracy))
+	f64, f64accuracy := bigF.Float64()
+	logging.FrameworkDebug(ctx, fmt.Sprintf("Float32Type ValueFromTerraform Float64 value %f, accuracy %d", f64, f64accuracy))
```
austinvalle, line 85: Is there a specific scenario you're looking to capture with these debug logs? I'm not necessarily against it, but this code could use a comment if there's a situation where this information would be useful.

Do we care only when the accuracy is changed? Maybe that could trigger the log rather than always logging

austinvalle (approved):
Great work! Everything lgtm, just had two notes to consider but nothing blocking 🚀

PR #1015, opened 2024-07-01
reflect: Use `float64` string representation for `float32` reflection equivalency

SBGoods:
Relates: #1014 

This fixes a bug encountered when using `function.Float32Parameter` in a provider defined function, similar to issue #914:
```go
=== RUN   TestFloat32Function_known
    float32_function_test.go:20: Step 1/1 error: Error running pre-apply plan: exit status 1
        
        Error: Error in function call
        
          on terraform_plugin_test.tf line 13, in output "test":
          13: 					value = provider::framework::float32(1.23)
            ├────────────────
            │ while calling provider::framework::float32(float32_param)
        
        Call to function "provider::framework::float32" failed: Value Conversion
        Error: An unexpected error was encountered trying to convert to number. This
        is always an error in the provider. Please report the following to the
        provider developer:
        
        cannot store 1.23 in float32.
--- FAIL: TestFloat32Function_known (0.36s)
```

The current reflection logic for `float32` compares the reflection target's string representation of the `float32` value with the string representation of that same value in a new `big.Float` object and throws an error if the representations differ. Creating a new `big.Float` object requires a conversion from `float32` to `float64` resulting in precision loss and differing string representations. 

This PR switches the string comparisons to `float64` and adds overflow and underflow checks for `float32`.

### DIFF
--- a/internal/reflect/number.go
+++ b/internal/reflect/number.go
@@ -132,9 +132,9 @@ func Number(ctx context.Context, typ attr.Type, val tftypes.Value, target reflec
 			return reflect.ValueOf(uintResult), diags
 		}
 	case reflect.Float32:
-		floatResult, _ := result.Float32()
+		float64Result, _ := result.Float64()
 
-		bf := big.NewFloat(float64(floatResult))
+		bf := big.NewFloat(float64Result)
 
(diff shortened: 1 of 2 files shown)


austinvalle (approved):
lgtm, nice test additions 🚀

PR #1016, opened 2024-07-04
Remove extra unaccepted comma in the example code

nasa9084:
The comma in the example code is unexpected in the context and causes a parsing error


austinvalle (approved):
Thanks for the doc fix @nasa9084 🚀

PR #1027, opened 2024-08-14
docs: Miscellaneous doc fixes in attribute + unit tests

austinvalle:
Just some cleaning up I found while working on something else.

------------------

FYI, for the unit tests being uncommented, they were originally failing because collection types with missing element types are considered invalid, and thus not equal: https://github.com/hashicorp/terraform-plugin-framework/blob/91edd79716f11d776a12427a8d49a6a15e0566b9/types/basetypes/list_type.go#L112-L115

Since all these custom type tests were embedding the underlying collection type, the element type was `nil`, thus failing the `cmp.Diff` in all the tests with a confusing diff 😆 
```bash
--- FAIL: TestMapNestedAttributeGetType (0.00s)
    --- FAIL: TestMapNestedAttributeGetType/custom-type (0.00s)
        /Users/austin.valle/code/terraform-plugin-framework/resource/schema/map_nested_attribute_test.go:411: unexpected difference:   testtypes.MapType(
            - 	{},
            + 	{},
              )
```

The int64/int32/float64/float32/object unit tests were never failing, but likely just commented out on accident 👍🏻

### DIFF
--- a/.changes/unreleased/NOTES-20240906-095258.yaml
+++ b/.changes/unreleased/NOTES-20240906-095258.yaml
@@ -1,0 +1,7 @@
+kind: NOTES
+body: 'all: This Go module has been updated to Go 1.22 per the [Go support policy](https://go.dev/doc/devel/release#policy).
+  It is recommended to review the [Go 1.22 release notes](https://go.dev/doc/go1.22)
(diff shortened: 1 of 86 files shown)


austinvalle:
Also contains the linter fixes for go 1.23 😆

SBGoods (approved):

PR #1064, opened 2024-12-06
internal/fwschemadata: Rewrite `SetValue` semantic equality logic to ignore order 

### LINKED ISSUE #1061: Semantic Equal on Sets expects identical order of elements
### Module version

```
% go list -m github.com/hashicorp/terraform-plugin-framework
github.com/hashicorp/terraform-plugin-framework v1.13.0
```

Schema with SetNestedAttribute with a custom type `CaseInsensitiveStringType` that implements `StringSemanticEquals`.
```go
	resp.Schema = schema.Schema{
			"id": schema.StringAttribute{
				Computed:    true,
				},
			},
			"set_attribute": schema.SetNestedAttribute{
				Optional:    true,
						"resource_id": schema.StringAttribute{
							Required:    true,
						},
						"value": schema.StringAttribute{
							CustomType:  internaltypes.CaseInsensitiveStringType{},
							Description: "(case-insensitive) custom type",
							Required:    true,
						},
					},
				},
			},
		},
	}
```

```hcl
resource "..." "..." {
  set_attribute = [
    {
      resource_id = "id1"
      value   = "Value"
    },
    {
      resource_id = "id2"
      value   = "Value2"
    }
  ]
}
```

### Expected Behavior
Sets should not care about the order of the elements contained within. This is how the `SetValue.Equals` function behaves here:

https://github.com/hashicorp/terraform-plugin-framework/blob/426086150717f06336f79f69cb0c46e3fd136bc1/types/basetypes/set_value.go#L282-L286

This property should hold for both Equals and Semantic Equals

### Actual Behavior
Semantic Equals compares elements based on their index in the elements array and will result in plan changes if the order of elements and case of `value` change during `Read` on the Resource:

https://github.com/hashicorp/terraform-plugin-framework/blob/426086150717f06336f79f69cb0c46e3fd136bc1/internal/fwschemadata/value_semantic_equality_set.go#L131-L166

This results in a plan output like:

```
  ~ resource "..." "..." {
        id                 = "..."
      ~ set_attribute = [
          - {
              - resource_id = "id1" -> null
              - value   = "value" -> null
            },
          - {
              - resource_id = "id2" -> null
              - value   = "value2" -> null
            },
          + {
              + resource_id = "id1"
              + value   = "Value"
            },
          + {
              + resource_id = "id2"
              + value   = "Value2"
            },
        ]
    }
```

austinvalle:
Closes #1061

This PR fixes the set semantic equality logic, which prior to this change, was written equivalent to list semantic equality.

Sets are unordered, so to correctly compare the semantic equality of two sets, we must compare each element in the set against all other elements to determine if there is a semantically equivalent element in the set. Once we find a semantically equal element, we remove it from the slice of candidates and continue.

----------

As an example, consider a set of case insensitive strings.

With the previous logic, the following sets would be considered semantically equal:
```bash
["value-a", "value-b", "value-c"]
["VALUE-A", "VALUE-B", "VALUE-C"]
```

However the following set would be incorrectly considered semantically not equal, due to the difference in order:
```bash
["value-a", "value-b", "value-c"]
["VALUE-B", "VALUE-C", "VALUE-A"]
```

This change will result in both of these examples being considered semantically equal.

### Corner tests

I also wrote some acceptance tests to make it a little easier to understand the impact (using IPv6 address strings as an example)

https://github.com/hashicorp/terraform-provider-corner/pull/297

### Unit test failures before the fix
https://github.com/hashicorp/terraform-plugin-framework/actions/runs/12832297072/job/35784696031?pr=1064
```bash
--- FAIL: TestValueSemanticEqualitySet (0.00s)
    --- FAIL: TestValueSemanticEqualitySet/SetValue-SetValue-StringValuableWithSemanticEquals-true-diff-order (0.00s)
        value_semantic_equality_set_test.go:1021: unexpected difference:   &fwschemadata.ValueSemanticEqualityResponse{
            - 	NewValue:    s`[["KEEP-LOWERCASE-012","KEEP-LOWERCASE-789"],["KEEP-LOWERCASE-456","KEEP-LOWERCASE-123"]]`,
            + 	NewValue:    s`[["keep-lowercase-123","keep-lowercase-456"],["keep-lowercase-789","keep-lowercase-012"]]`,
              	Diagnostics: nil,
              }
    --- FAIL: TestValueSemanticEqualitySet/SetValue-StringValuableWithSemanticEquals-true-diff-order (0.00s)
        value_semantic_equality_set_test.go:10[21](https://github.com/hashicorp/terraform-plugin-framework/actions/runs/12832297072/job/35784696031?pr=1064#step:5:22): unexpected difference:   &fwschemadata.ValueSemanticEqualityResponse{
            - 	NewValue:    s`["KEEP-LOWERCASE-456","KEEP-LOWERCASE-123"]`,
            + 	NewValue:    s`["keep-lowercase-123","keep-lowercase-456"]`,
              	Diagnostics: nil,
              }
```

### DIFF
--- a/internal/fwschemadata/value_semantic_equality_set.go
+++ b/internal/fwschemadata/value_semantic_equality_set.go
@@ -145,10 +145,10 @@ func ValueSemanticEqualitySetElements(ctx context.Context, req ValueSemanticEqua
 		}
 
 		// Loop through all prior value elements and see if there are any semantically equal elements
-		for pIdx, priorValue := range priorValueElements {
+		for pIdx, priorValueElement := range priorValueElements {
 			elementReq := ValueSemanticEqualityRequest{
 				Path:             req.Path.AtSetValue(proposedNewValueElement),
-				PriorValue:       priorValue,
+				PriorValue:       priorValueElement,
 				ProposedNewValue: proposedNewValueElement,
 			}
 			elementResp := &ValueSemanticEqualityResponse{
--- a/internal/fwschemadata/value_semantic_equality_set.go
+++ b/internal/fwschemadata/value_semantic_equality_set.go
@@ -128,10 +128,6 @@ func ValueSemanticEqualitySetElements(ctx context.Context, req ValueSemanticEqua
 	// Short circuit flag
 	updatedElements := false
 
-	// The underlying loop will mutate priorValueElements to avoid keeping
-	// duplicate semantically equal elements. Need the original length to avoid a panic
-	originalPriorElementsLength := len(priorValueElements)
-
 	// Loop through proposed elements by delegating to the recursive semantic
 	// equality logic. This ensures that recursion will catch a further
 	// underlying element type has its semantic equality logic checked, even if
@@ -139,10 +135,6 @@ func ValueSemanticEqualitySetElements(ctx context.Context, req ValueSemanticEqua
 	for idx, proposedNewValueElement := range proposedNewValueElements {
 		// Ensure new value always contains all of proposed new value
 		newValueElements[idx] = proposedNewValueElement
-
-		if idx >= originalPriorElementsLength {
-			continue
-		}
 
 		// Loop through all prior value elements and see if there are any semantically equal elements
 		for pIdx, priorValueElement := range priorValueElements {
--- a/internal/fwschemadata/value_semantic_equality_set.go
+++ b/internal/fwschemadata/value_semantic_equality_set.go
@@ -128,6 +128,10 @@ func ValueSemanticEqualitySetElements(ctx context.Context, req ValueSemanticEqua
 	// Short circuit flag
 	updatedElements := false
 
+	// The underlying loop will mutate priorValueElements to avoid keeping
+	// duplicate semantically equal elements. Need the original length to avoid panicks
+	originalPriorElementsLength := len(priorValueElements)
+
 	// Loop through proposed elements by delegating to the recursive semantic
 	// equality logic. This ensures that recursion will catch a further
 	// underlying element type has its semantic equality logic checked, even if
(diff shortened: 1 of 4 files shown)


austinvalle:
Since we have write-only attributes landing soon, this bug fix will go in after

bbasata:
I read line-by-line through `internal/fwschemadata/value_semantic_equality_set.go` a couple times until it tangibly clicked with me what semantic equality is about -- that was my first challenge. "Why does equality modify values?" Oh, ok! Now I want to call it "prior state prevails when the difference is cosmetic" 😃

With that, I think I have unlocked the power-up needed to review the PR 😃

bbasata:
> Once we find a semantically equal element, we remove it from the slice of candidates and continue.

Non-blocking for merge, simply a consideration: is this for optimization or correctness? I'm guessing optimization -- I think the behavior is still correct without doing this.

Related: A set might contain the same value more than once. Or multiple semantically-equal values. And I imagine it's a Hard (and unnecessary) Problem to make any promise about the behavior in that case. If the behavior mattered in that scenario, we'd likely advise folks to use a `List`.

internal/fwschemadata/value_semantic_equality_set.go
```diff
@@ -136,33 +140,44 @@ func ValueSemanticEqualitySetElements(ctx context.Context, req ValueSemanticEqua
 		// Ensure new value always contains all of proposed new value
 		newValueElements[idx] = proposedNewValueElement
 
-		if idx >= len(priorValueElements) {
+		if idx >= originalPriorElementsLength {
```
bbasata, line 143: Correctness: since we're handling the sets as unordered, I think this is not needed. A match for a prior set element could appear anywhere in the proposed set.

bbasata (approved):
:shipit:

internal/fwschemadata/value_semantic_equality_set.go
```diff
@@ -136,33 +140,44 @@ func ValueSemanticEqualitySetElements(ctx context.Context, req ValueSemanticEqua
...
+				continue
+			}
+
+			// Prior state was kept, meaning that we found a semantically equal element
+			updatedElements = true
 
-		updatedElements = true
-		newValueElements[idx] = elementResp.NewValue
+			// Remove the semantically equal element from the slice of candidates
+			priorValueElements = append(priorValueElements[:pIdx], priorValueElements[pIdx+1:]...)
```
austinvalle, line 175: @bbasata Adding as a comment to make a thread 😄 

> Non-blocking for merge, simply a consideration: is this for optimization or correctness? I'm guessing optimization -- I think the behavior is still correct without doing this.
> 
> Related: A set might contain the same value more than once. Or multiple semantically-equal values. And I imagine it's a Hard (and unnecessary) Problem to make any promise about the behavior in that case. If the behavior mattered in that scenario, we'd likely advise folks to use a `List`.

I believe it is necessary for correctness as well, but maybe I'm overthinking it 🤔 

Consider these values, where the set is of case-insensitive strings:
```bash
prior: ["value-a", "value-b", "value-D", "value-c"]
new: ["VALUE-A", "VALUE-B", "VALUE-b", "VALUE-C"]
```

When looping through the new values, if we don't remove candidates when they match we'll get these iterations:
1. ["value-a", "VALUE-B", "VALUE-b", "VALUE-C"] // new[0] matched with prior[0]
2. ["value-a", "value-b", "VALUE-b", "VALUE-C"] // new[1] matched with prior[1]
3. ["value-a", "value-b", "value-b", "VALUE-C"] // new[2] matched with prior[1]
4. ["value-a", "value-b", "value-b", "value-c"] // new[3] matched with prior[3] (an invalid set now)

What we really want to end with is: `["value-a", "value-b", "value-D", "value-c"]`

Does that make sense?

austinvalle: Re-reading your comment, I believe that was the exact scenario you were poking at 😄

austinvalle: Ah yes, you're right! Then I guess all that leaves it is as optimization. Do you think it's fair to keep the logic or should I remove it as premature optimization?

bbasata: I have no strong opinion on it, so :shipit: 😃

It seemed interesting, so I wanted to poke at it a bit & see what it was telling us.

austinvalle: Okay cool, I'm going to leave it as-is for now, and we can adjust the logic later if we prefer to simplify it

bbasata (approved):
:shipit: 

I see the CI / lint Action wants something from us -- I'm happy to stamp this again if/when needed.

austinvalle:
Will create a new PR for lint deprecations since it's gonna be noisy and unrelated to these changes 👍🏻

PR #1090, opened 2025-02-11
Improve dynamic attribute internal validation handling

SBGoods:
Follow up PR to #1044 

Mainly adds logic to handle underlying null values for dynamic attributes in the internal attribute validation. It also addresses minor comments in the write-only attribute PR.


internal/fwserver/server_upgraderesourcestate.go
```diff
@@ -264,16 +264,5 @@ func (s *Server) UpgradeResourceState(ctx context.Context, req *UpgradeResourceS
 	}
 	upgradeResourceStateResponse.State.Raw = modifiedState
 
-	// If the write-only nullification results in a null state, then this is a provider error
```
SBGoods, line 267: Addresses: https://github.com/hashicorp/terraform-plugin-framework/pull/1044#discussion_r1946615764

internal/fwserver/write_only_nullification.go
```diff
@@ -58,14 +57,6 @@ func NullifyWriteOnlyAttributes(ctx context.Context, resourceSchema fwschema.Sch
 		// Value type from new state to create null with
 		newValueType := attribute.GetType().TerraformType(ctx)
 
-		// If the attribute is dynamic set the new value type to DynamicPseudoType
```
SBGoods, line 61: Addresses: https://github.com/hashicorp/terraform-plugin-framework/pull/1044#discussion_r1946625602

We've verified with Core that erasing the underlying type is fine for write-only value nullification, so this conditional is no longer needed.

internal/fwserver/write_only_nullification_test.go
```diff
@@ -39,9 +39,6 @@ func TestNullifyWriteOnlyAttributes(t *testing.T) {
 			"dynamic-nil": schema.DynamicAttribute{
 				Optional: true,
 			},
-			"dynamic-underlying-string-nil-computed": schema.DynamicAttribute{
```
SBGoods, line 42: Addresses: https://github.com/hashicorp/terraform-plugin-framework/pull/1044#discussion_r1946634361

website/docs/plugin/framework/resources/write-only-arguments.mdx
```diff
@@ -109,4 +109,4 @@ Since write-only arguments have no prior values, user intent or value changes ca
 - Pair write-only arguments with a configuration attribute (required or optional) to “trigger” the use of the write-only argument
   - For example, a `password_wo` write-only argument can be paired with a configured `password_wo_version` attribute. When the `password_wo_version` is modified, the provider will send the `password_wo` value to the API.
 - Use a keepers attribute (which is used in the [Random Provider](https://registry.terraform.io/providers/hashicorp/random/latest/docs#resource-keepers)) that will take in arbitrary key-pair values. Whenever there is a change to the `keepers` attribute, the provider will use the write-only argument value.
-- Use the resource's [private state] to store secure hashes of write-only argument values, the provider will then use the hash to determine if a write-only argument value has changed in later Terraform runs.
```
SBGoods, line 112: Addresses: https://github.com/hashicorp/terraform-plugin-framework/pull/1044#discussion_r1946671898

internal/fwserver/attribute_validation.go
```diff
@@ -101,47 +101,97 @@ func AttributeValidate(ctx context.Context, a fwschema.Attribute, req ValidateAt
...
-		)
-	}
+	// Dynamic values need to perform more logic to check the config value for null/unknown-ness
+	dynamicValuable, ok := attributeConfig.(basetypes.DynamicValuable)
+	if ok {
+		dynamicConfigVal, diags := dynamicValuable.ToDynamicValue(ctx)
+		resp.Diagnostics.Append(diags...)
+		if diags.HasError() {
+			return
+		}
```
austinvalle, line 111: Perhaps something to consider, now that we're doing a bunch more of these null checks on `attributeConfig` that need to consider `types.Dynamic`, is storing the determination of if the config value is "null" to a variable.

```go
configHasNullValue := attributeConfig.IsNull()
// If the value is dynamic, we need to ensure the underlying value is null
if dynamicValuable, isDynamic := attributeConfig.(basetypes.DynamicValuable); !configHasNullValue && isDynamic {
	dynamicConfigVal, diags := dynamicValuable.ToDynamicValue(ctx)
	if diags.HasError() {
		return
	}
	if dynamicConfigVal.IsUnderlyingValueNull() {
		configHasNullValue = true
	}
}
```

Then you could use `configHasNullValue` in place of `attributeConfig.IsNull()` in this `AttributeValidate` function. You could also do the same thing for unknown values and remove that dynamic checking logic completely from the deprecation logic

internal/fwserver/attribute_validation.go
```diff
@@ -101,47 +101,97 @@ func AttributeValidate(ctx context.Context, a fwschema.Attribute, req ValidateAt
...
-			fmt.Sprintf("Must set a configuration value for the %s attribute as the provider has marked it as required.\n\n", req.AttributePath.String())+
-				"Refer to the provider documentation or contact the provider developers for additional information about configurable attributes that are required.",
-		)
-	}
+		// Terraform CLI does not automatically perform certain configuration
+		// checks yet. If it eventually does, this logic should remain at least
+		// until Terraform CLI versions 0.12 through the release containing the
+		// checks are considered end-of-life.
+		// Reference: https://github.com/hashicorp/terraform/issues/30669
+		if a.IsComputed() && !a.IsOptional() && !dynamicConfigVal.IsUnderlyingValueNull() {
```
austinvalle, line 118: I believe what you're likely observing with the corner provider errors are that you also need to check the container (`attributeConfig.IsNull()`) for null/not nullness as well for dynamics. If you refactor to putting this into a variable we can avoid duplicating the logic as well 👍🏻

austinvalle (commented):
Left some thoughts on the dynamic attribute validation, everything else looks good!

Also thank you for leaving a bread crumb in the comments, that was very helpful 🙏🏻

internal/fwserver/attribute_validation.go
austinvalle: I definitely missed adding any unit tests for these scenarios when I implemented dynamic types, which makes sense because the logic was broken 😆

austinvalle (approved):
Great work! 🚀

PR #1095, opened 2025-02-18
resource/schema: Add validation to prevent write-only attributes in sets and set based data

SBGoods:
Write-only attributes when used with ephemeral data cannot be used within any set and set-based data due to limitations in Terraform Core. This PR adds validation to block write-only attributes from being used with sets, set nested attributes, and set nested blocks.


resource/schema/write_only_nested_attribute_validation_test.go
```diff
@@ -7,14 +7,13 @@ import (
...
 
 	"github.com/hashicorp/terraform-plugin-framework/internal/fwschema"
-	"github.com/hashicorp/terraform-plugin-framework/provider/metaschema"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
 )
 
 func TestContainsAllWriteOnlyChildAttributes(t *testing.T) {
 	t.Parallel()
 	tests := map[string]struct {
-		nestedAttr metaschema.NestedAttribute
```
austinvalle, line 17: 😆 nice catch!

austinvalle (approved):
LGTM 🚀

Issue #1096, opened 2025-02-19
Write-only dynamic attribute causes plan diff everytime


magodo:
### Module version

```
ecd80f67daed
```

### Relevant provider source code

```go
...
```

### Terraform Configuration Files

```hcl
terraform {
  required_providers {
    restful = {
      source = "magodo/restful"
    }
  }
}

provider "restful" {
  base_url = "http://local:3000"
}

variable "foo" {
   type = string
   ephemeral = true
   default = null
}

resource "restful_resource" "test" {
  path = "/resources"
  read_path = "/resources/$(body.id)"
  body = {
      foo = "bar"
  }
  ephemeral_body = {
    bar = var.foo
  }
}
```

### Debug Output

After the initial successful apply, any further plan will cause diff:

```shell
Terraform used the selected providers to generate the following execution plan. Resource actions are indicated
with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  # restful_resource.test will be updated in-place
  ~ resource "restful_resource" "test" {
        id             = "/resources/2"
      ~ output         = {
          - foo = "bar"
          - id  = 2
        } -> (known after apply)
        # (4 unchanged attributes hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```

### Expected Behavior

If I understand the [document](https://github.com/hashicorp/terraform-plugin-framework/blob/main/website/docs/plugin/framework/resources/write-only-arguments.mdx) correctly:

> Write-only argument values cannot produce a Terraform plan difference.
>
> This is because the prior state value for a write-only argument will always be null and the planned/final state value will also be null, therefore, it cannot produce a diff on its own.

Even if the ephemeral input variable got changed, the plan shall also show no diff as long as other non write-only attributes are not changed.

### Actual Behavior

After the initial successful apply, any further plan will cause diff.

### Steps to Reproduce

Use the branch https://github.com/magodo/terraform-provider-restful/tree/ephemeral_body to build and use the provider locally.

### References

Note that the `output` attribute is Dynamic type. I've also tried to generate the plan output, which is like below:

```json
{
  "format_version": "1.2",
  "terraform_version": "1.12.0-dev",
  "variables": {
    "foo": {
      "value": null
    }
  },
  "planned_values": {
    "root_module": {
      "resources": [
        {
          "address": "restful_resource.test",
          "mode": "managed",
          "type": "restful_resource",
          "name": "test",
[...]

PR #1097, opened 2025-02-19
internal/fwserver: Move write-only nullification to earlier in PlanResourceChange

SBGoods:
Fixes: https://github.com/hashicorp/terraform-plugin-framework/issues/1096

During PlanResourceChange, computed attributes that are `null` in the configuration are marked as `unknown`. This logic only runs when there is a non-null planned state and the planned state is not equal to the prior state. If a write-only attribute is set in config, then the prior state (which is always `null`) and planned state (which contains the config value) will always differ causing `null` computed values to always be marked as `unknown` instead of using the prior value. This causes perpetual diffs after resource creation if a write-only attribute is set in config and the resource has non-null computed attribute values in state. 

The fix moves the write-only value nullification for the planned state to happen before this check.

### DIFF
--- a/.changes/unreleased/BUG FIXES-20250219-152752.yaml
+++ b/.changes/unreleased/BUG FIXES-20250219-152752.yaml
@@ -1,0 +1,5 @@
+kind: BUG FIXES
+body: 'internal/fwserver: fixed bug where write-only attributes set in configuration would cause perpetual diffs for computed attributes.'
+time: 2025-02-19T15:27:52.52508-05:00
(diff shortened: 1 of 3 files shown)


austinvalle (approved):
LGTM, would like to see this change run against the new corner provider tests before we release the next version 🚀

magodo:
Thanks @SBGoods for the instant fix! @austinvalle I can confirm this solved my issue #1096. I'll continue developing my provider and report new issues if any.

PR #1106, opened 2025-03-03
Migrate docs to web-unified-docs

im2nguyen:
We are migrating the `terraform-plugin-*` documentation to a unified HashiCorp product documentation repository. As a result of this migration, the unified repo (`web-unified-docs`) will be the source of truth.

This PR adds a noticed to all existing MDX files in the original location so folks know where to contribute. We plan on deprecating and removing the files in `/website`.

### DIFF
--- a/website/README.md
+++ b/website/README.md
@@ -1,5 +1,8 @@
 # Terraform Documentation
 
+> [!IMPORTANT]  
+> **Documentation Update:** Product documentation previously located in `/website` has moved to the `hashicorp/web-unified-docs` repository, where all product documentation is now centralized. Please make contributions directly to `web-unified-docs`, since changes to `/website` in this repository will not appear on [developer.hashicorp.com](https://developer.hashicorp.com)
+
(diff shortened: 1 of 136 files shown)


RubenSandwich:
@im2nguyen just a reminder that we won't be merging this in during the trail period.

im2nguyen:
@RubenSandwich just confirmed with you on Slack and with Tony Vetter that we are going to merge this as part of the terraform-plugin-framework migration tomorrow AM

deniseyu (approved):

LeahMarieBush:
Do we want this as a comment in the mdx instead of as actual content? Just in case we need to roll back UDR for whatever reason

im2nguyen:
> Do we want this as a comment in the mdx instead of as actual content? Just in case we need to roll back UDR for whatever reason

It doesn't show in the preview if we do it as a comment in the MDX. If we need to rollback, we can do ctrl+f and replace all

LeahMarieBush (approved):

PR #1107, opened 2025-03-06
Resource Identity: Initial implementation + Schemas

austinvalle:
> Since Framework has so many files to update, I felt like this was a good breaking point for a first PR.

This PR implements the `tfprotov{5/6}.ProviderServerWithResourceIdentity` interface, for each RPC:
- `GetResourceIdentitySchemas`: This PR contains about 95% of the implementation with the external packages/internal attribute updates and tests. The missing piece is validation on list attribute (see below).
- `UpgradeResourceIdentity`: This RPC will be implemented in a separate PR
- All of the RPCs that are just adding the identity data to the request/response will be implemented in a separate PR, but they are all additional data, so ignoring them is fine for existing providers.

------------------------------

## Other related work for resource identity (coming soon)
- Adding identity data to request/response of all existing RPCs
- Validation logic for list attributes
- Import Resource RPC updates
- Upgrade Resource Identity implementation
- Changelogs
- Website documentation (separate repo)


internal/toproto6/getresourceidentityschemas_test.go
```diff
@@ -0,0 +1,367 @@
...
+						IdentityAttributes: []*tfprotov6.ResourceIdentitySchemaAttribute{
+							{
+								Name:              "test_attribute",
+								RequiredForImport: true,
+								Type:              tftypes.Bool,
+							},
+						},
+					},
+				},
+			},
```
ansgarm, line 150: This test looks identical to the one above – or am I lost in the code?
image

ansgarm: Seems like it's intentional though 🤔

internal/toproto6/identity_schema_attribute.go
```diff
@@ -0,0 +1,43 @@
...
+
+	identitySchemaAttribute := &tfprotov6.ResourceIdentitySchemaAttribute{
+		Name:              name,
+		RequiredForImport: a.IsRequiredForImport(),
+		OptionalForImport: a.IsOptionalForImport(),
+		Type:              a.GetType().TerraformType(ctx),
+
+		// Unlike other schema attributes, identity attributes only have a single description field which
+		// is assumed to be markdown. Both a.GetDescription() and a.GetMarkdownDescription() will return
+		// the same string, so we just chose one here.
```
ansgarm, line 38: ❤️

ansgarm (dismissed):
Nice work 👏 This is a lot of gruntwork, respect for going through it so meticulously!
I tried my best to find copy paste errors but you let me down here 😆💯

SBGoods (dismissed):

internal/toproto6/getresourceidentityschemas_test.go
austinvalle: You're correct they are definitely the same, and likely not both needed. The intention with these tests is to just check the mapping, but here it's impossible to define an attribute without the type, so they end up the same.

(honestly there is a ton of duplicated tests in FW, we'll probably clean this up in a year or two once they start annoying us lol)

PR #1112, opened 2025-03-14
Resource Identity: Add identity data to RPCs needed to store/read from state

austinvalle:
This PR allows a framework provider to store/read identity data for a managed resource (along with [`Terraform v1.12.0-alpha20250312`](https://github.com/hashicorp/terraform/releases/tag/v1.12.0-alpha20250312)). The RPCs implemented for this are `ReadResource` (`Read`), `PlanResourceChange`, and `ApplyResourceChange` (split in framework to `Create`/`Update`/`Delete`).

I left some intentional TODOs for pieces of work I wasn't 100% sure of that should be resolved before we release stable `v1.15.0`.

--------

Generally, the flow of data through the packages (using protocol v5 as an example, it's the same for v6):
- `proto5server` (implements the RPC in plugin-go) => `fromproto5` (maps request from plugin-go types to fwserver types)
- `proto5server` => `fwserver` (this is where any framework-specific logic is implemented for the RPC)
- `fwserver` => `resource` (the actual provider code for the resource, like aws or azurerm)
- `proto5server` => `toproto5` (maps response from fwserver types to plugin-go types)

--------

The other RPCs that are not needed to successfully test the happy path of `plan`/`apply` workflow, these will be implemented in follow-up PRs:
- `ImportResourceState`
- `MoveResourceState`
- `UpgradeResourceIdentity`


austinvalle:
Using the core alpha, storing identity via Read or Create looks like the follow:
image

No validation yet on the immutability of identity, but more to come there!

internal/fwserver/server_deleteresource.go
```diff
@@ -96,6 +98,17 @@ func (s *Server) DeleteResource(ctx context.Context, req *DeleteResourceRequest,
 		resp.Private = req.PlannedPrivate
 	}
 
+	// If the resource supports identity pre-populate a null value.
+	// TODO:ResourceIdentity: This should probably be prior identity, but we don't currently have that in the protocol.
```
ansgarm, line 102: should we / did you already mention this to core?

internal/fwserver/server_deleteresource_test.go
```diff
@@ -334,6 +353,114 @@ func TestServerDeleteResource(t *testing.T) {
...
+				},
+				IdentitySchema: testIdentitySchema,
+				ResourceSchema: testSchema,
+				Resource: &testprovider.Resource{
+					DeleteMethod: func(ctx context.Context, req resource.DeleteRequest, resp *resource.DeleteResponse) {
+						if resp.Identity == nil || !resp.Identity.Raw.IsNull() {
+							resp.Diagnostics.AddError(
+								"Unexpected resp.Identity",
+								"expected resp.Identity to be a null object of the schema type.",
+							)
```
ansgarm, line 376: Related to your comment above – I could see an ask of provider developers to be able to use identity to delete resources. At least once providers don't have to support the absence of identity support (i.e. older Terraform versions) anymore 🤔

ansgarm (approved):

SBGoods (approved):

rainkwan:
LGTM as well 👍🏽

bbasata (approved):

internal/fwserver/server_deleteresource_test.go
austinvalle: Yeah, I think that data should be populated regardless. Core still hasn't given us a strict definition of when the data will be what values, but it should be something at least 😆

PR #1123, opened 2025-03-27
Add resource identity to Move RPC

rainkwan:
Draft

### DIFF
--- a/internal/fwserver/server_moveresourcestate.go
+++ b/internal/fwserver/server_moveresourcestate.go
@@ -65,14 +65,23 @@ type MoveResourceStateRequest struct {
 	// TargetTypeName is the type name of the target resource as given by
 	// Terraform across the protocol.
 	TargetTypeName string
+
+	// SourceIdentity is the identity of the source resource.
+	//
+	// Only the underlying JSON field is populated.
+	SourceIdentity *tfprotov6.RawState
+
+	// SourceIdentitySchemaVersion is the version of the source resource state.
+	SourceIdentitySchemaVersion int64
 }
 
 // MoveResourceStateResponse is the framework server response for the
 // MoveResourceState RPC.
 type MoveResourceStateResponse struct {
-	Diagnostics   diag.Diagnostics
-	TargetPrivate *privatestate.Data
-	TargetState   *tfsdk.State
+	Diagnostics    diag.Diagnostics
+	TargetPrivate  *privatestate.Data
+	TargetState    *tfsdk.State
+	TargetIdentity *tfsdk.ResourceIdentity
 }
 
 // MoveResourceState implements the framework server MoveResourceState RPC.
@@ -125,17 +134,24 @@ func (s *Server) MoveResourceState(ctx context.Context, req *MoveResourceStateRe
 
 	for _, resourceStateMover := range resourceStateMovers {
 		moveStateReq := resource.MoveStateRequest{
-			SourcePrivate:         sourcePrivate,
-			SourceProviderAddress: req.SourceProviderAddress,
-			SourceRawState:        req.SourceRawState,
-			SourceSchemaVersion:   req.SourceSchemaVersion,
-			SourceTypeName:        req.SourceTypeName,
+			SourcePrivate:               sourcePrivate,
+			SourceProviderAddress:       req.SourceProviderAddress,
+			SourceRawState:              req.SourceRawState,
+			SourceSchemaVersion:         req.SourceSchemaVersion,
+			SourceTypeName:              req.SourceTypeName,
+			SourceIdentity:              req.SourceIdentity,
+			SourceIdentitySchemaVersion: req.SourceIdentitySchemaVersion,
 		}
 		moveStateResp := resource.MoveStateResponse{
 			TargetPrivate: privatestate.EmptyProviderData(ctx),
 			TargetState: tfsdk.State{
 				Schema: req.TargetResourceSchema,
 				Raw:    tftypes.NewValue(req.TargetResourceSchema.Type().TerraformType(ctx), nil),
+			},
+			// TODO: See if we need to change this to identity
+			TargetIdentity: &tfsdk.ResourceIdentity{
+				Raw:    tftypes.NewValue(req.TargetResourceSchema.Type().TerraformType(ctx), nil),
+				Schema: req.TargetResourceSchema,
 			},
 		}
 
@@ -237,6 +253,8 @@ func (s *Server) MoveResourceState(ctx context.Context, req *MoveResourceStateRe
 			"Source Provider Address: "+req.SourceProviderAddress+"\n"+
 			"Source Resource Type: "+req.SourceTypeName+"\n"+
 			"Source Resource Schema Version: "+strconv.FormatInt(req.SourceSchemaVersion, 10)+"\n"+
-			"Target Resource Type: "+req.TargetTypeName,
+			"Target Resource Type: "+req.TargetTypeName, //+"\n"+
+		// "Source Identity Schema Version: "+strconv.FormatInt(req.SourceSchemaVersion, 10),
+
 	)
 }
--- a/resource/move_state.go
+++ b/resource/move_state.go
@@ -75,6 +75,14 @@ type MoveStateRequest struct {
 	// other request fields, to determine whether the request should be handled
 	// by this particular implementation.
 	SourceTypeName string
+
+	// SourceIdentity is the identity of the source resource.
+	//
+	// Only the underlying JSON field is populated.
+	SourceIdentity *tfprotov6.RawState
+
+	// SourceIdentitySchemaVersion is the version of the source resource state.
+	SourceIdentitySchemaVersion int64
 }
 
 // MoveStateResponse represents a response to a MoveStateRequest.
@@ -107,4 +115,7 @@ type MoveStateResponse struct {
 	// operation. This field is not pre-populated as it is up to implementations
 	// whether the source private data is relevant for the target resource.
 	TargetPrivate *privatestate.ProviderData
+
+	// TargetIdentity is the identity of the target resource.
+	TargetIdentity *tfsdk.ResourceIdentity
 }
--- a/internal/fromproto5/moveresourcestate.go
+++ b/internal/fromproto5/moveresourcestate.go
@@ -38,14 +38,15 @@ func MoveResourceStateRequest(ctx context.Context, proto5 *tfprotov5.MoveResourc
 	}
 
 	fw := &fwserver.MoveResourceStateRequest{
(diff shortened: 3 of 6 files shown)


internal/toproto5/resource_identity.go
```diff
@@ -14,7 +14,7 @@ import (
 
 // ResourceIdentity returns the *tfprotov5.ResourceIdentityData for a *tfsdk.ResourceIdentity.
 func ResourceIdentity(ctx context.Context, fw *tfsdk.ResourceIdentity) (*tfprotov5.ResourceIdentityData, diag.Diagnostics) {
-	if fw == nil {
+	if fw == nil || fw.Schema == nil {
```
austinvalle, line 17: Is this something you ran into specifically? I would expect this to never happen 🤔

resource/move_state.go
```diff
@@ -75,6 +75,14 @@ type MoveStateRequest struct {
 	// other request fields, to determine whether the request should be handled
 	// by this particular implementation.
 	SourceTypeName string
+
+	// SourceIdentity is the identity of the source resource.
+	//
+	// Only the underlying JSON field is populated.
+	SourceIdentity *tfprotov6.RawState
+
+	// SourceIdentitySchemaVersion is the version of the source resource state.
```
austinvalle, line 84:
```suggestion
	// SourceIdentitySchemaVersion is the version of the source resource identity.
```

austinvalle (commented):
Overall everything is looking good! I had a couple questions/nits

internal/fwserver/server_moveresourcestate.go
```diff
@@ -237,6 +271,8 @@ func (s *Server) MoveResourceState(ctx context.Context, req *MoveResourceStateRe
 			"Source Provider Address: "+req.SourceProviderAddress+"\n"+
 			"Source Resource Type: "+req.SourceTypeName+"\n"+
 			"Source Resource Schema Version: "+strconv.FormatInt(req.SourceSchemaVersion, 10)+"\n"+
-			"Target Resource Type: "+req.TargetTypeName,
+			"Target Resource Type: "+req.TargetTypeName, //+"\n"+
+		// "Source Identity Schema Version: "+strconv.FormatInt(req.SourceSchemaVersion, 10),
```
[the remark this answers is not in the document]
rainkwan: Yeah, I think I will omit returning a diagnostic here for now to avoid confusion and opt for the more specific functionality for that later if needed as we discussed

internal/fwserver/server_moveresourcestate_test.go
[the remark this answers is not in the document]
rainkwan: Added the test here 

https://github.com/hashicorp/terraform-plugin-framework/pull/1123/files#diff-599c61942d4492eedc1e197f8fd0cb49150b360ca7fc4b6939815ab637cc6eb8R930-R968 

But when calling `resp.Diagnostics.Append(resp.TargetIdentity.SetAttribute(ctx, path.Root("test_id"), "should-not-be-set")...)` it triggers a nil pointer panic in `SetAttribute` when calling `s.data()`

So I had to move the error checking there in order to prevent the nil pointer but not sure if that was the best way to handle it. It felt like leaving it as a nil pointer error was bad.

https://github.com/hashicorp/terraform-plugin-framework/pull/1123/files#diff-3c465cc799700a6886a2a6a7b400ba25ec775308952e85c0c155405ff8078b17R72-R81

tfsdk/resource_identity.go
```diff
@@ -70,6 +69,16 @@ func (s *ResourceIdentity) Set(ctx context.Context, val interface{}) diag.Diagno
...
 // Lists can only have the next element added according to the current length.
 func (s *ResourceIdentity) SetAttribute(ctx context.Context, path path.Path, val interface{}) diag.Diagnostics {
+	// If s is nil, then calling s.data triggers a nil pointer error so we return the error diag here
+	if s == nil {
+		return diag.Diagnostics{
+			diag.NewErrorDiagnostic(
+				"Unexpected Move State Response",
+				"An unexpected error was encountered when creating the move state response. Identity data was returned by the provider move state operation, but the resource does not indicate identity support.\n\n"+
+					"This is always a problem with the provider and should be reported to the provider developer."),
+		}
```
austinvalle, line 79: I think this is a good idea in combination with the other suggestions I left, especially since resource identity frequently uses pointers. My only notes would be to adjust the message and add similar errors to the other methods on `tfsdk.ResourceIdentity`! `Get`, `GetAttribute`, `PathMatches`, `Set` and `SetAttribute`

```suggestion
		return diag.Diagnostics{
			diag.NewErrorDiagnostic(
				"Missing Identity Definition",
				"An unexpected error was encountered when attempting to set a resource identity attribute. The resource does not indicate support via a resource identity schema.\n\n"+
					"This is always a problem with the provider and should be reported to the provider developer."),
		}
```

internal/fwserver/server_moveresourcestate.go
austinvalle, line 1: I still think we need to keep this logic from line 228
```go
if resp.TargetIdentity != nil && req.IdentitySchema == nil {
    resp.Diagnostics.AddError(
        "Unexpected Move State Response",
        "An unexpected error was encountered when creating the move state response. New identity data was returned by the provider move state operation, but the resource does not indicate identity support.\n\n"+
            "This is always a problem with the provider and should be reported to the provider developer.",
    )

    return
}
```

It's possible to just create a valid `tfsdk.ResourceIdentity`, which we can't send back to Terraform if the resource doesn't support identity ([example](https://github.com/hashicorp/terraform-plugin-framework/blob/b25d4c2c6651dab7b3903669219ec84ad723e1b4/internal/fwserver/server_readresource_test.go#L670-L674)):
```go
resp.TargetIdentity = &tfsdk.ResourceIdentity{
	Raw: tftypes.NewValue(testIdentitySchemaType, map[string]tftypes.Value{
		"test_id": tftypes.NewValue(tftypes.String, "new-id-123"),
	}),
	Schema: testIdentitySchema,
}
```

internal/fwserver/server_moveresourcestate_test.go
```diff
@@ -812,6 +824,148 @@ func TestServerMoveResourceState(t *testing.T) {
...
+					"id":                 "test-id-value",
+					"required_attribute": true,
+				}),
+				TargetResource: &testprovider.ResourceWithMoveState{
+					MoveStateMethod: func(ctx context.Context) []resource.StateMover {
+						return []resource.StateMover{
+							{
+								StateMover: func(_ context.Context, req resource.MoveStateRequest, resp *resource.MoveStateResponse) {
+									// Try to set TargetIdentity even though req.IdentitySchema is nil
+									resp.Diagnostics.Append(resp.TargetIdentity.SetAttribute(ctx, path.Root("test_id"), "should-not-be-set")...)
```
austinvalle, line 945: This is the scenario we're trying to prevent, if I do this it'll send back identity data to Terraform despite us not "announcing" an identity schema 🙂:

```suggestion
									// This resource doesn't indicate identity support (via a schema), so this should raise a diagnostic.
									resp.TargetIdentity = &tfsdk.ResourceIdentity{
										Raw: tftypes.NewValue(testIdentitySchema.Type().TerraformType(ctx), map[string]tftypes.Value{
											"test_id": tftypes.NewValue(tftypes.String, "should-not-be-set"),
										}),
										Schema: testIdentitySchema,
									}
```

internal/fwserver/server_moveresourcestate_test.go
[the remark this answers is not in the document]
austinvalle: Ah I see the confusion! We want `resp.TargetIdentity` to be `nil` because in this scenario/test, the resource doesn't actually support identity, so we don't want to populate that data with anything. Providers shouldn't be setting it at all.

I left some comments elsewhere to try and illustrate what I'm talking about since with this approach it's still possible for provider developers to return identity data when the resource doesn't support it.

internal/fwserver/server_moveresourcestate_test.go
rainkwan: This doesn't raise a diagnostic, should it?

austinvalle: Yeah it should be raising a diagnostic, it looks like the check being on line 228 was too early because `resp.TargetIdentity` is not set at that point (it's at `moveStateResp.TargetIdentity` when it comes back from your provider code that sets it).

Other RPCs just immediately set `resp.Identity` after the RPC, but if you want to move that piece of logic down lower we can also do that!

I'll make a comment suggestion 👍🏻

internal/fwserver/server_moveresourcestate.go
```diff
@@ -205,6 +226,16 @@ func (s *Server) MoveResourceState(ctx context.Context, req *MoveResourceStateRe
...
+		if resp.TargetIdentity != nil && req.IdentitySchema == nil {
+			resp.Diagnostics.AddError(
+				"Unexpected Move State Response",
+				"An unexpected error was encountered when creating the move state response. New identity data was returned by the provider move state operation, but the resource does not indicate identity support.\n\n"+
+					"This is always a problem with the provider and should be reported to the provider developer.",
+			)
+
+			return
+		}
+
```
austinvalle, line 238: `resp.TargetIdentity` hasn't been set yet in this code path, so it'll always be nil.

```suggestion
```

internal/fwserver/server_moveresourcestate.go
```diff
@@ -221,6 +252,9 @@ func (s *Server) MoveResourceState(ctx context.Context, req *MoveResourceStateRe
 		if !moveStateResp.TargetState.Raw.Equal(tftypes.NewValue(req.TargetResourceSchema.Type().TerraformType(ctx), nil)) {
 			resp.Diagnostics = moveStateResp.Diagnostics
 			resp.TargetState = &moveStateResp.TargetState
+			if moveStateResp.TargetIdentity != nil {
+				resp.TargetIdentity = moveStateResp.TargetIdentity
+			}
```
austinvalle, line 257: We can check the identity here! (you can also check it above by just using `moveStateResp.TargetIdentity`, doesn't matter which 🤷🏻 

```suggestion
			if moveStateResp.TargetIdentity != nil {
				resp.TargetIdentity = moveStateResp.TargetIdentity
			}

			if resp.TargetIdentity != nil && req.IdentitySchema == nil {
				resp.Diagnostics.AddError(
					"Unexpected Move State Response",
					"An unexpected error was encountered when creating the move state response. New identity data was returned by the provider move state operation, but the resource does not indicate identity support.\n\n"+
						"This is always a problem with the provider and should be reported to the provider developer.",
				)

				return
			}
```

austinvalle: Once you update this your test should show diagnostics 👍🏻

austinvalle (approved):
🎉

PR #1125, opened 2025-04-02
Implement validation logic for "identityschema.ListAttribute"

rainkwan:
Most of the “invalid” identity attribute types are simply not possible to create with Framework (without abusing custom types, which we can safely ignore imo) because we have a dedicated “identityschema” package: https://github.com/hashicorp/terraform-plugin-framework/tree/main/resource/identityschema

The lone validation we need to apply at runtime is checking the element type of the “identityschema.ListAttribute” since there is no restriction on what kind of “attr.Type” is provided. The only allowed element types are:

types.Bool

types.Int64, types.Int32, types.Float64, types.Float32, types.Number

types.String

The validation was done in the ValidateImplementation function which runs during GetResourceIdentitySchemas.

Added AttributeInvalidElementTypeDiag to give the error message and IsAllowedPrimitiveType to check to see if the element is an allowed type.

### DIFF
--- a/internal/fwschema/diagnostics.go
+++ b/internal/fwschema/diagnostics.go
@@ -64,3 +64,18 @@ func AttributeDefaultTypeMismatchDiag(attributePath path.Path, expectedType attr
 			"The default value must match the type of the schema.",
 	)
 }
+
+// AttributeInvalidElementTypeDiag returns an error diagnostic to provider
+// developers about using non-primitive types in their Attribute
+// implementation. This is not allowed.
+func AttributeInvalidElementTypeDiag(attributePath path.Path) diag.Diagnostic {
+	// The diagnostic path is intentionally omitted as it is invalid in this
+	// context. Diagnostic paths are intended to be mapped to actual data,
+	// while this path information must be synthesized.
+	return diag.NewErrorDiagnostic(
+		"Invalid Attribute Implementation",
+		"When validating the schema, an implementation issue was found. "+
+			"This is always an issue with the provider and should be reported to the provider developers.\n\n"+
+			fmt.Sprintf("%q contains an Attribute type that is not allowed for Resource Identity. ", attributePath),
+	)
+}
--- a/internal/fwschema/diagnostics.go
+++ b/internal/fwschema/diagnostics.go
@@ -68,7 +68,7 @@ func AttributeDefaultTypeMismatchDiag(attributePath path.Path, expectedType attr
 // AttributeInvalidElementTypeDiag returns an error diagnostic to provider
 // developers about using non-primitive types in their Attribute
 // implementation. This is not allowed.
-func AttributeInvalidElementTypeDiag(attributePath path.Path) diag.Diagnostic {
+func AttributeInvalidElementTypeDiag(attributePath path.Path, actualType attr.Type) diag.Diagnostic {
 	// The diagnostic path is intentionally omitted as it is invalid in this
 	// context. Diagnostic paths are intended to be mapped to actual data,
 	// while this path information must be synthesized.
@@ -76,6 +76,7 @@ func AttributeInvalidElementTypeDiag(attributePath path.Path) diag.Diagnostic {
 		"Invalid Attribute Implementation",
 		"When validating the schema, an implementation issue was found. "+
 			"This is always an issue with the provider and should be reported to the provider developers.\n\n"+
-			fmt.Sprintf("%q contains an Attribute type that is not allowed for Resource Identity. ", attributePath),
+			fmt.Sprintf("%q contains an Attribute of type %q that is not allowed for Lists in Resource Identity. ", attributePath, actualType)+
+			"Lists in Resource Identity may only have primitive element types such as Bool, Int, Float, Number and String.",
 	)
 }
--- a/internal/fwschema/diagnostics.go
+++ b/internal/fwschema/diagnostics.go
@@ -76,7 +76,7 @@ func AttributeInvalidElementTypeDiag(attributePath path.Path, actualType attr.Ty
 		"Invalid Attribute Implementation",
 		"When validating the schema, an implementation issue was found. "+
 			"This is always an issue with the provider and should be reported to the provider developers.\n\n"+
-			fmt.Sprintf("%q contains an Attribute of type %q that is not allowed for Lists in Resource Identity. ", attributePath, actualType)+
+			fmt.Sprintf("%q contains an element of type %q that is not allowed for Lists in Resource Identity. ", attributePath, actualType)+
 			"Lists in Resource Identity may only have primitive element types such as Bool, Int, Float, Number and String.",
 	)
 }
--- a/internal/fwtype/static_collection_validation.go
+++ b/internal/fwtype/static_collection_validation.go
@@ -143,13 +143,13 @@ func ReturnCollectionWithDynamicTypeDiag() diag.Diagnostic {
 // IsAllowedPrimitiveType checks if the given attr.Type is an allowed primitive type for resource identity.
 func IsAllowedPrimitiveType(typ attr.Type) bool {
 	switch typ.(type) {
-	case basetypes.BoolType,
-		basetypes.Float64Type,
-		basetypes.Float32Type,
-		basetypes.Int64Type,
-		basetypes.Int32Type,
-		basetypes.NumberType,
-		basetypes.StringType:
+	case basetypes.BoolTypable,
+		basetypes.Float64Typable,
+		basetypes.Float32Typable,
+		basetypes.Int64Typable,
+		basetypes.Int32Typable,
+		basetypes.NumberTypable,
+		basetypes.StringTypable:
 		return true
 	default:
 		return false
--- a/internal/fwtype/static_collection_validation.go
+++ b/internal/fwtype/static_collection_validation.go
@@ -5,7 +5,6 @@ package fwtype
 
 import (
 	"fmt"
-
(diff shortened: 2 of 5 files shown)


resource/identityschema/list_attribute.go
```diff
@@ -160,10 +161,13 @@ func (a ListAttribute) IsOptionalForImport() bool {
...
 // should never include false positives.
-func (a ListAttribute) ValidateImplementation(ctx context.Context, req fwschema.ValidateImplementationRequest, resp *fwschema.ValidateImplementationResponse) {
+func (a ListAttribute) ValidateImplementation(_ context.Context, req fwschema.ValidateImplementationRequest, resp *fwschema.ValidateImplementationResponse) {
 	if a.CustomType == nil && a.ElementType == nil {
 		resp.Diagnostics.Append(fwschema.AttributeMissingElementTypeDiag(req.Path))
+		return
 	}
 
-	// TODO:ResourceIdentity: Write validation + tests that ensure the element type only contains primitive elements (bool, string, number)
+	if !fwtype.IsAllowedPrimitiveType(a.ElementType) {
```
austinvalle, line 170: nit - We don't need to check this if it's a custom type
```suggestion
	if a.CustomType == nil && !fwtype.IsAllowedPrimitiveType(a.ElementType) {
```

internal/fwtype/static_collection_validation.go
```diff
@@ -140,3 +139,19 @@ func ReturnCollectionWithDynamicTypeDiag() diag.Diagnostic {
...
+func IsAllowedPrimitiveType(typ attr.Type) bool {
+	switch typ.(type) {
+	case basetypes.BoolType,
+		basetypes.Float64Type,
+		basetypes.Float32Type,
+		basetypes.Int64Type,
+		basetypes.Int32Type,
+		basetypes.NumberType,
+		basetypes.StringType:
+		return true
```
austinvalle, line 153: I don't think it's super common, but I'm thinking we should use the "typable" interfaces here instead (which will give you a superset of the functionality you have here.

```suggestion
	case basetypes.BoolTypable,
		basetypes.Float64Typable,
		basetypes.Float32Typable,
		basetypes.Int64Typable,
		basetypes.Int32Typable,
		basetypes.NumberTypable,
		basetypes.StringTypable:
		return true
```

The difference there is that if you set the element type to a custom type that is a primitive (like an `iptypes.IPv4Address`, then it'll still recognize it as a valid list of strings)

internal/fwschema/diagnostics.go
```diff
@@ -64,3 +64,18 @@ func AttributeDefaultTypeMismatchDiag(attributePath path.Path, expectedType attr
...
+func AttributeInvalidElementTypeDiag(attributePath path.Path) diag.Diagnostic {
+	// The diagnostic path is intentionally omitted as it is invalid in this
+	// context. Diagnostic paths are intended to be mapped to actual data,
+	// while this path information must be synthesized.
+	return diag.NewErrorDiagnostic(
+		"Invalid Attribute Implementation",
+		"When validating the schema, an implementation issue was found. "+
+			"This is always an issue with the provider and should be reported to the provider developers.\n\n"+
+			fmt.Sprintf("%q contains an Attribute type that is not allowed for Resource Identity. ", attributePath),
+	)
```
austinvalle, line 80: I think we should consider printing more information in this diagnostic and maybe be a little more specific about what's wrong (since the diagnostic is only used by one validation)

Some wording that mentions that the only valid `list` for resource identity is one that has a primitive element type (bool, int32, etc.), then maybe print out the `attr.Type` that we received?

austinvalle (commented):
Looking good 👍🏻 , I have some suggestions for the case switch + diagnostic messaging, but all the functionality looks correct!

resource/identityschema/list_attribute.go
bbasata: > We don't need to check this if it's a custom type

💭 Do we need another conditional here to validate a custom type against the allowed typables? Or is that somehow implicit?

austinvalle: Well `a.ElementType` will be `nil` if it's a custom type which is why I mentioned that specifically here.

As for whether we should validate a custom type in general, we could although I wasn't as concerned about abusing a custom type to avoid those restrictions.

If we do want to validate the custom type, we would probably want to use the `.TerraformType()`, since that's ultimately what we send back to Terraform.

bbasata: Got it! Makes sense -- it took me another read to see that `a.CustomType` here is the concrete type of the `ListAttribute`, not the type of the list elements.

internal/fwschema/diagnostics.go
```diff
@@ -64,3 +64,19 @@ func AttributeDefaultTypeMismatchDiag(attributePath path.Path, expectedType attr
...
+func AttributeInvalidElementTypeDiag(attributePath path.Path, actualType attr.Type) diag.Diagnostic {
+	// The diagnostic path is intentionally omitted as it is invalid in this
+	// context. Diagnostic paths are intended to be mapped to actual data,
+	// while this path information must be synthesized.
+	return diag.NewErrorDiagnostic(
+		"Invalid Attribute Implementation",
+		"When validating the schema, an implementation issue was found. "+
+			"This is always an issue with the provider and should be reported to the provider developers.\n\n"+
+			fmt.Sprintf("%q contains an Attribute of type %q that is not allowed for Lists in Resource Identity. ", attributePath, actualType)+
+			"Lists in Resource Identity may only have primitive element types such as Bool, Int, Float, Number and String.",
```
rainkwan, line 80: Updated the error message here

internal/fwschema/diagnostics.go
```diff
@@ -64,3 +64,19 @@ func AttributeDefaultTypeMismatchDiag(attributePath path.Path, expectedType attr
...
+// implementation. This is not allowed.
+func AttributeInvalidElementTypeDiag(attributePath path.Path, actualType attr.Type) diag.Diagnostic {
+	// The diagnostic path is intentionally omitted as it is invalid in this
+	// context. Diagnostic paths are intended to be mapped to actual data,
+	// while this path information must be synthesized.
+	return diag.NewErrorDiagnostic(
+		"Invalid Attribute Implementation",
+		"When validating the schema, an implementation issue was found. "+
+			"This is always an issue with the provider and should be reported to the provider developers.\n\n"+
+			fmt.Sprintf("%q contains an Attribute of type %q that is not allowed for Lists in Resource Identity. ", attributePath, actualType)+
```
austinvalle, line 79: A small but important distinction 🙂 
```suggestion
			fmt.Sprintf("%q contains an element of type %q that is not allowed for Lists in Resource Identity. ", attributePath, actualType)+
```

resource/identityschema/list_attribute_test.go
```diff
@@ -480,6 +480,135 @@ func TestListAttributeValidateImplementation(t *testing.T) {
...
+				Name: "test",
+				Path: path.Root("test"),
+			},
+			expected: &fwschema.ValidateImplementationResponse{
+				Diagnostics: diag.Diagnostics{
+					diag.NewErrorDiagnostic(
+						"Invalid Attribute Implementation",
+						"When validating the schema, an implementation issue was found. "+
+							"This is always an issue with the provider and should be reported to the provider developers.\n\n"+
+							"\"test\" contains an Attribute of type \"basetypes.DynamicType\" that is not allowed for Lists in Resource Identity. "+
```
austinvalle, line 564:
```suggestion
							"\"test\" contains an element of type \"basetypes.DynamicType\" that is not allowed for Lists in Resource Identity. "+
```

resource/identityschema/list_attribute_test.go
```diff
@@ -480,6 +480,135 @@ func TestListAttributeValidateImplementation(t *testing.T) {
...
+				Name: "test",
+				Path: path.Root("test"),
+			},
+			expected: &fwschema.ValidateImplementationResponse{
+				Diagnostics: diag.Diagnostics{
+					diag.NewErrorDiagnostic(
+						"Invalid Attribute Implementation",
+						"When validating the schema, an implementation issue was found. "+
+							"This is always an issue with the provider and should be reported to the provider developers.\n\n"+
+							"\"test\" contains an Attribute of type \"types.ObjectType[]\" that is not allowed for Lists in Resource Identity. "+
```
austinvalle, line 585:
```suggestion
							"\"test\" contains an element of type \"types.ObjectType[]\" that is not allowed for Lists in Resource Identity. "+
```

resource/identityschema/list_attribute_test.go
```diff
@@ -480,6 +480,135 @@ func TestListAttributeValidateImplementation(t *testing.T) {
...
+				Name: "test",
+				Path: path.Root("test"),
+			},
+			expected: &fwschema.ValidateImplementationResponse{
+				Diagnostics: diag.Diagnostics{
+					diag.NewErrorDiagnostic(
+						"Invalid Attribute Implementation",
+						"When validating the schema, an implementation issue was found. "+
+							"This is always an issue with the provider and should be reported to the provider developers.\n\n"+
+							"\"test\" contains an Attribute of type \"types.MapType[!!! MISSING TYPE !!!]\" that is not allowed for Lists in Resource Identity. "+
```
austinvalle, line 606:
```suggestion
							"\"test\" contains an element of type \"types.MapType[!!! MISSING TYPE !!!]\" that is not allowed for Lists in Resource Identity. "+
```

austinvalle (approved):
Just had one small adjustment to the error message, great work on this! 🙌🏻

austinvalle (approved):

PR #1126, opened 2025-04-03
ResourceIdentity: Add support for import by identity and update pass-through implementations

austinvalle:
This PR adds support for plannable import by resource identity, which is first supported in Terraform [`v1.12.0-beta1`](https://github.com/hashicorp/terraform/releases/tag/v1.12.0-beta1).

-------

Today, the only supported plannable import data is the `id` field, which is just a string that provider typically just redirects to the [import stub state ](https://github.com/hashicorp/terraform/blob/main/docs/resource-instance-change-lifecycle.md#import-behavior):
```terraform
import {
  to = aws_route53_record.myrecord
  id = "Z4KAPRWWNC7JR_dev.example.com_NS"
}
```

Now, if the resource supports identity, the import block will allow the identity to be supplied in the `identity` attribute:
```terraform
import {
  to = aws_route53_record.myrecord
  identity = {
    zone_id = "Z4KAPRWWNC7JR"
    record_name = "dev.example.com"
    record_type = "NS"
  }
}
```

------

The identity is automatically passed through to the `Read` implementation, which typically does the heavy lifting for populating state of an imported resource, however the provider can choose to populate the identity with more information, set data in the import stub state, etc.

### DIFF
--- a/resource/import_state.go
+++ b/resource/import_state.go
@@ -34,7 +34,18 @@ type ImportStateRequest struct {
 	// as an Attribute. However, this identifier can also be treated as
 	// its own type of value and parsed during import. This value
 	// is not stored in the state unless the provider explicitly stores it.
+	//
+	// This ID field is supplied in the "terraform import" CLI command or in the import config block "id" attribute.
+	// Either ID or Identity must be supplied by the practitioner, depending on the method used to import the resource.
 	ID string
+
+	// Identity is the configuration data provided by the practitioner in the import config block "identity" attribute. This
+	// configuration data will conform to the identity schema defined by the managed resource. If the resource does not support identity,
+	// this value will not be set.
+	//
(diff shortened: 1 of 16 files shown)


resource/import_state.go
austinvalle, line 1: One thing not included in this PR is an implementation of an "identity => import stub state" pass-through. I'm still not 100% sure what provider developers will prefer in-terms of helper functions, but we can easily create a new one if needed.

.changes/unreleased/NOTES-20250403-121229.yaml
austinvalle, line 1: My plan was to release a framework `1.15.0-beta.1` with this PR (along with Rain's currently open #1125 and #1123 )

bbasata (approved):
:shipit: ✨

PR #1132, opened 2025-04-22
resource: Add identity to DeleteRequest

austinvalle:
**Context:** https://hashicorp.atlassian.net/browse/TFECO-9284

This PR introduces the `Identity` field to `DeleteRequest`. This was missed during initial implementation, despite always existing in the protocol/core. I left a comment explaining the possible confusion of using "planned" objects as a "prior" object.

### DIFF
--- a/internal/fwserver/server_applyresourcechange.go
+++ b/internal/fwserver/server_applyresourcechange.go
@@ -76,6 +76,10 @@ func (s *Server) ApplyResourceChange(ctx context.Context, req *ApplyResourceChan
 		deleteReq := &DeleteResourceRequest{
 			PlannedPrivate: req.PlannedPrivate,
 			PriorState:     req.PriorState,
+			// MAINTAINER NOTE: There isn't a separate data field for prior identity, like there is with prior_state and planned_state.
+			// Here the planned_identity field will contain what would be considered the prior identity (since the final identity value
+			// after deleting will be null).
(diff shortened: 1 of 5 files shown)


internal/fwserver/server_deleteresource.go
austinvalle, line 1: Not sure why this file got a merge conflict, but it's resolved!

internal/fwserver/server_applyresourcechange.go
```diff
@@ -76,6 +76,10 @@ func (s *Server) ApplyResourceChange(ctx context.Context, req *ApplyResourceChan
 		deleteReq := &DeleteResourceRequest{
 			PlannedPrivate: req.PlannedPrivate,
 			PriorState:     req.PriorState,
+			// MAINTAINER NOTE: There isn't a separate data field for prior identity, like there is with prior_state and planned_state.
```
rainkwan, line 79: Taking a mental note to mention this in the documentation

rainkwan (approved):
LGTM 🚀

internal/fwserver/server_applyresourcechange.go
austinvalle: I'm not sure it'll be super useful information for provider developers to know exactly where the source is coming from (since it's exposed to them as just `Identity`, but if we find a good time to mention it we definitely can!

PR #1133, opened 2025-04-22
chore: Fix identity TODOs and add documentation for interfaces

austinvalle:
This PR just goes through and cleans up all the `TODO:ResourceIdentity` comments, of which I decided that we don't need to handle any of them ATM. We can always do some refactoring internally if we decide it's worth it later.

I also fleshed out the documentation on the `resource.ResourceWithIdentity` interface itself.

### DIFF
--- a/resource/resource.go
+++ b/resource/resource.go
@@ -19,6 +19,9 @@ import (
 //   - Plan Modification: Schema-based or entire plan
 //     via ResourceWithModifyPlan.
 //   - State Upgrades: ResourceWithUpgradeState
+//   - Identity: Define an identity schema with ResourceWithIdentity to enable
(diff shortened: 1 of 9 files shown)


resource/resource.go
```diff
@@ -199,7 +202,21 @@ type ResourceWithValidateConfig interface {
 
 // ResourceWithIdentity is an interface type that extends Resource to implement managed resource identity.
 //
-// TODO:ResourceIdentity: Add more documentation here to describe what identity is used for.
+// Managed resources can optionally define an identity schema, which is a separate object stored in state
```
austinvalle, line 205:
```suggestion
// Managed resources can optionally define an identity schema, which represents a separate object stored in state
```

bbasata (approved):
✨

PR #1134, opened 2025-04-24
Resource Identity: Add a new import helper that can pass-through an identity attribute and an import ID.

austinvalle:
**Context:** https://hashicorp.atlassian.net/browse/TFECO-9285

This PR introduces a new import helper that ensures the scenario of an "id-only refresh" can work with a single identity attribute or an import identifier. If we find there are other useful import helpers we can always introduce them later!


ansgarm (approved):

ansgarm:
Nice work!

PR #1135, opened 2025-04-28
Resource Identity: Add the UpgradeRPC for resource identity

rainkwan:
Follow-Up PR for https://github.com/hashicorp/terraform-plugin-framework/pull/1112
This PR adds the UpgradePRC for Resource Identity to Framework.

### DIFF
--- a/internal/fwserver/server_upgraderesourceidentity.go
+++ b/internal/fwserver/server_upgraderesourceidentity.go
@@ -69,51 +69,16 @@ func (s *Server) UpgradeResourceIdentity(ctx context.Context, req *UpgradeResour
 		},
 	}
 
-	// TODO: Maybe throw error if the schemas are the same, it is a bug in core
-
-	// Terraform CLI can call UpgradeResourceIdentity even if the stored Identity
-	// version matches the current schema. Presumably this is to account for
-	// the previous terraform-plugin-sdk implementation, which handled some
-	// Identity fixups on behalf of Terraform CLI. When this happens, we do not
-	// want to return errors for a missing ResourceWithUpgradeIdentity
-	// implementation or an undefined version within an existing
-	// ResourceWithUpgradeIdentity implementation as that would be confusing
-	// detail for provider developers. Instead, the framework will attempt to
-	// roundtrip the prior RawState to a Identity matching the current Schema.
-	//
-	// TODO: To prevent provider developers from accidentally implementing
-	// ResourceWithUpgradeIdentity with a version matching the current schema
-	// version which would never get called, the framework can introduce a
-	// unit test helper.
-	// Reference: https://github.com/hashicorp/terraform-plugin-framework/issues/113
-	//
-	// UnmarshalWithOpts allows optionally ignoring instances in which elements being
-	// do not have a corresponding attribute within the schema.
-	/*	if req.Version == req.IdentitySchema.GetVersion() {
-		logging.FrameworkTrace(ctx, "UpgradeResourceIdentity request version matches current Schema version, using framework defined passthrough implementation")
-
-		identitySchemaType := req.IdentitySchema.Type().TerraformType(ctx)
-
-		rawStateValue, err := req.RawState.UnmarshalWithOpts(identitySchemaType, unmarshalOpts)
-
-		if err != nil {
-			resp.Diagnostics.AddError(
-				"Unable to Read Previously Saved Identity for UpgradeResourceIdentity",
-				"There was an error reading the saved resource Identity using the current resource schema.\n\n"+
-					"If this resource Identity was last refreshed with Terraform CLI 0.11 and earlier, it must be refreshed or applied with an older provider version first. "+
-					"If you manually modified the resource Identity, you will need to manually modify it to match the current resource schema. "+
-					"Otherwise, please report this to the provider developer:\n\n"+err.Error(),
-			)
-			return
-		}
-
-		resp.UpgradedIdentity = &tfsdk.ResourceIdentity{
-			Schema: req.IdentitySchema,
-			Raw:    rawStateValue,
-		}
-
+	if req.Version == req.IdentitySchema.GetVersion() {
+		resp.Diagnostics.AddError(
+			"Unable to Read Previously Saved Identity for UpgradeResourceIdentity",
+			"There was an error reading the saved resource identity using the current resource identity schema.\n\n"+
+				"If this resource Identity was last refreshed with Terraform CLI 0.11 and earlier, it must be refreshed or applied with an older provider version first. "+
+				"If you manually modified the resource Identity, you will need to manually modify it to match the current resource identity schema. "+
+				"Otherwise, please report this to the provider developer:\n\n",
+		)
 		return
-	}*/
+	}
 
 	if resourceWithConfigure, ok := req.Resource.(resource.ResourceWithConfigure); ok {
 		logging.FrameworkTrace(ctx, "Resource implements ResourceWithConfigure")
@@ -178,7 +143,7 @@ func (s *Server) UpgradeResourceIdentity(ctx context.Context, req *UpgradeResour
 
 		priorSchemaType := resourceIdentityUpgrader.PriorSchema.Type().TerraformType(ctx)
 
-		_, err := req.RawState.UnmarshalWithOpts(priorSchemaType, unmarshalOpts)
+		rawIdentityValue, err := req.RawState.UnmarshalWithOpts(priorSchemaType, unmarshalOpts)
 
 		if err != nil {
 			resp.Diagnostics.AddError(
@@ -189,12 +154,15 @@ func (s *Server) UpgradeResourceIdentity(ctx context.Context, req *UpgradeResour
 			return
 		}
 
-		upgradeResourceIdentityRequest.RawState = req.RawState
+		upgradeResourceIdentityRequest.Identity = &tfsdk.ResourceIdentity{
+			Raw:    rawIdentityValue, // from the output of req.RawState.UnmarshalWithOpts
+			Schema: *resourceIdentityUpgrader.PriorSchema,
+		}
 
 	}
 
 	upgradeResourceIdentityResponse := resource.UpgradeResourceIdentityResponse{
-		UpgradedIdentity: &tfsdk.ResourceIdentity{
+		Identity: &tfsdk.ResourceIdentity{
 			Schema: req.IdentitySchema,
 			// Raw is intentionally not set.
 		},
@@ -215,7 +183,7 @@ func (s *Server) UpgradeResourceIdentity(ctx context.Context, req *UpgradeResour
 		return
 	}
 
-	if upgradeResourceIdentityResponse.UpgradedIdentity.Raw.Type() == nil || upgradeResourceIdentityResponse.UpgradedIdentity.Raw.IsNull() {
+	if upgradeResourceIdentityResponse.Identity.Raw.Type() == nil || upgradeResourceIdentityResponse.Identity.Raw.IsNull() {
 		resp.Diagnostics.AddError(
 			"Missing Upgraded Resource Identity",
 			fmt.Sprintf("After attempting a resource Identity upgrade to version %d, the provider did not return any Identity data. ", req.Version)+
@@ -225,16 +193,16 @@ func (s *Server) UpgradeResourceIdentity(ctx context.Context, req *UpgradeResour
 		return
 	}
 
-	if upgradeResourceIdentityResponse.UpgradedIdentity != nil {
-		logging.FrameworkTrace(ctx, "UpgradeResourceIdentityResponse Raw State set, overriding State")
+	if upgradeResourceIdentityResponse.Identity.Raw.Type() != nil {
+		logging.FrameworkTrace(ctx, "UpgradeResourceIdentityResponse RawState set, overriding State")
 
 		resp.UpgradedIdentity = &tfsdk.ResourceIdentity{
 			Schema: req.IdentitySchema,
-			Raw:    upgradeResourceIdentityResponse.UpgradedIdentity.Raw,
+			Raw:    upgradeResourceIdentityResponse.Identity.Raw,
 		}
 
 		return
 	}
 
-	resp.UpgradedIdentity = upgradeResourceIdentityResponse.UpgradedIdentity
+	resp.UpgradedIdentity = upgradeResourceIdentityResponse.Identity
 }
--- a/internal/fwserver/server_upgraderesourceidentity.go
+++ b/internal/fwserver/server_upgraderesourceidentity.go
@@ -16,9 +16,9 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
 )
 
-// UpgradeIdentityRequest is the framework server request for the
-// UpgradeIdentity RPC.
-type UpgradeIdentityRequest struct {
+// UpgradeResourceIdentityRequest is the framework server request for the
+// UpgradeResourceIdentity RPC.
+type UpgradeResourceIdentityRequest struct {
 	Resource       resource.Resource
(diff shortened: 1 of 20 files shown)


internal/fwserver/server_upgraderesourceidentity.go
```diff
@@ -0,0 +1,240 @@
...
+				"Unable to Read Previously Saved Identity for UpgradeResourceIdentity",
+				fmt.Sprintf("There was an error reading the saved resource Identity using the prior resource schema defined for version %d upgrade.\n\n", req.Version)+
+					"Please report this to the provider developer:\n\n"+err.Error(),
+			)
+			return
+		}
+
+		upgradeResourceIdentityRequest.RawState = req.RawState
+
+	}
```
austinvalle, line 194: You have the right idea here with using a schema to decode the raw state, but we just need to use the actual output from `UnmarshalWithOpts`, since as this code is written `upgradeResourceIdentityRequest.RawState == req.RawState`, so nothing really happens here but validation that the state unmarshals.


What we'll want is a field on the `resource.UpgradeResourceIdentityRequest` struct for us to put a resource identity, like:
```go
type UpgradeResourceIdentityRequest struct {
	// .. other fields

	// Previous identity of the resource if the wrapping IdentityUpgrader
	// type PriorSchema field was present. When available, this allows for
	// easier data handling such as calling Get() or GetAttribute().
	Identity *tfsdk.ResourceIdentity
}
```

Then you can do something like the [existing resource upgrade logic](https://github.com/hashicorp/terraform-plugin-framework/blob/17f1faf8787c6987ade7b6d58e297f256473ab43/internal/fwserver/server_upgraderesourcestate.go#L186-L189), but using a `tfsdk.ResourceIdentity` struct:
```go
upgradeResourceIdentityRequest.Identity = &tfsdk.ResourceIdentity{
	Raw:    rawIdentityValue, // from the output of req.RawState.UnmarshalWithOpts
	Schema: *resourceIdentityUpgrader.PriorSchema,
}
```

internal/fwserver/server_upgraderesourceidentity.go
```diff
@@ -0,0 +1,240 @@
...
+	// Define options to be used when unmarshalling raw state.
+	// IgnoreUndefinedAttributes will silently skip over fields in the JSON
+	// that do not have a matching entry in the schema.
+	unmarshalOpts := tfprotov6.UnmarshalOpts{
+		ValueFromJSONOpts: tftypes.ValueFromJSONOpts{
+			IgnoreUndefinedAttributes: true,
+		},
+	}
+
+	// TODO: Maybe throw error if the schemas are the same, it is a bug in core
```
austinvalle, line 72: I think this sounds like a good idea to do 🙂 

All of the commented code can go as well 👍🏻

internal/proto5server/server_upgraderesourceidentity.go
```diff
@@ -5,11 +5,51 @@ package proto5server
...
+	"github.com/hashicorp/terraform-plugin-framework/internal/logging"
+	"github.com/hashicorp/terraform-plugin-framework/internal/toproto5"
 
 	"github.com/hashicorp/terraform-plugin-go/tfprotov5"
 )
 
 // UpgradeResourceIdentity satisfies the tfprotov5.ProviderServer interface.
 func (s *Server) UpgradeResourceIdentity(ctx context.Context, proto5Req *tfprotov5.UpgradeResourceIdentityRequest) (*tfprotov5.UpgradeResourceIdentityResponse, error) {
-	panic("unimplemented") // TODO:ResourceIdentity: implement
+	// panic("unimplemented") // TODO:ResourceIdentity: implement
```
austinvalle, line 18:
```suggestion
```

internal/proto6server/server_upgraderesourceidentity.go
```diff
@@ -5,11 +5,51 @@ package proto6server
...
+	"github.com/hashicorp/terraform-plugin-framework/internal/logging"
+	"github.com/hashicorp/terraform-plugin-framework/internal/toproto6"
 
 	"github.com/hashicorp/terraform-plugin-go/tfprotov6"
 )
 
 // UpgradeResourceIdentity satisfies the tfprotov6.ProviderServer interface.
 func (s *Server) UpgradeResourceIdentity(ctx context.Context, proto6Req *tfprotov6.UpgradeResourceIdentityRequest) (*tfprotov6.UpgradeResourceIdentityResponse, error) {
-	panic("unimplemented") // TODO:ResourceIdentity: implement
+	// panic("unimplemented") // TODO:ResourceIdentity: implement
```
austinvalle, line 18:
```suggestion
```

internal/fwserver/server_upgraderesourceidentity.go
```diff
@@ -0,0 +1,240 @@
...
+	if upgradeResourceIdentityResponse.UpgradedIdentity != nil {
+		logging.FrameworkTrace(ctx, "UpgradeResourceIdentityResponse Raw State set, overriding State")
+
+		resp.UpgradedIdentity = &tfsdk.ResourceIdentity{
+			Schema: req.IdentitySchema,
+			Raw:    upgradeResourceIdentityResponse.UpgradedIdentity.Raw,
+		}
+
+		return
+	}
```
austinvalle, line 237: I'm not 100% sure what this is looking to do, but I'm assuming it was brought over from the resource upgrade state implementation?

We should revisit what the original intention is here since as of right now, `upgradeResourceIdentityResponse.UpgradedIdentity == nil` will never be true. I guess I would expect an error if it was `nil`?

austinvalle (commented):
Left some initial comments!

internal/fwserver/server_upgraderesourceidentity.go
rainkwan: Yes, I believe the error is in the if prior to this line. The intention I thought was just to set the RawState of the identity but do you think it would be better to just keep the code and remove the if-statement?

austinvalle: Yeah I think we can actually just remove the if statement and all the code inside, since the last line in this function actually does the same thing. I'll make a specific suggestion in my follow-up review 👍🏻

resource/upgrade_identity.go
```diff
@@ -0,0 +1,42 @@
...
+	// Previous identity of the resource if the wrapping IdentityUpgrader
+	// type PriorSchema field was present. When available, this allows for
+	// easier data handling such as calling Get() or GetAttribute().
+	Identity *tfsdk.ResourceIdentity
+}
+
+// Response information for the provider logic to update a resource identity
+// from a prior resource identity version to the current identity version.
+type UpgradeResourceIdentityResponse struct {
+	Identity *tfsdk.ResourceIdentity
```
austinvalle, line 36: This is an external package, so this field needs documentation, see https://github.com/hashicorp/terraform-plugin-framework/blob/17f1faf8787c6987ade7b6d58e297f256473ab43/resource/upgrade_state.go#L64-L73 for an example

resource/upgrade_identity.go
```diff
@@ -0,0 +1,42 @@
...
+	// TypeName is the type of resource that Terraform needs to upgrade the
+	// identity state for.
+	TypeName string
+
+	// Version is the version of the identity state the resource currently has.
+	Version int64
+
+	// RawIdentity is the identity state as Terraform sees it right now in JSON. See the
+	// documentation for `RawIdentity` for information on how to work with the
+	// data it contains.
```
austinvalle, line 24: Might want to mention that this is meant for providers wanting to skip redeclaring old schemas and that it's considered advanced, see: https://github.com/hashicorp/terraform-plugin-framework/blob/17f1faf8787c6987ade7b6d58e297f256473ab43/resource/upgrade_state.go#L17-L26

Also this comment mentions looking at other documentation, but I'm not sure what that meant, perhaps left over from elsewhere?

resource/upgrade_identity.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// Response information for the provider logic to update a resource identity
+// from a prior resource identity version to the current identity version.
+type UpgradeResourceIdentityResponse struct {
+	Identity *tfsdk.ResourceIdentity
+
+	// Diagnostics report errors or warnings related to retrieving the resource
+	// identity schema. An empty slice indicates success, with no warnings
+	// or errors generated.
```
austinvalle, line 40:
```suggestion
	// Diagnostics report errors or warnings related to upgrading the resource
	// identity state. An empty slice indicates success, with no warnings
	// or errors generated.
```

resource/identity_upgrader.go
```diff
@@ -0,0 +1,47 @@
...
+	// UpgradeResourceIdentityRequest type RawIdentity field.
+	PriorSchema *identityschema.Schema
+
+	// Version is the version schema that this Upgrader will handle, converting
+	// it to Version+1.
+	Version int64
+
+	// Type describes the schema that this function can upgrade.
+	Type tftypes.Type
+
```
austinvalle, line 33: I don't see either of these used anywhere, are they needed here? (version should be defined in the key of the map of identity upgraders)

```suggestion
```

austinvalle (commented):
Left some more comments! Mostly just about package comments/naming/diags/extra fields

resource/upgrade_identity.go
```diff
@@ -0,0 +1,42 @@
...
+
+	// Previous identity of the resource if the wrapping IdentityUpgrader
+	// type PriorSchema field was present. When available, this allows for
+	// easier data handling such as calling Get() or GetAttribute().
+	Identity *tfsdk.ResourceIdentity
+}
+
+// Response information for the provider logic to update a resource identity
+// from a prior resource identity version to the current identity version.
+type UpgradeResourceIdentityResponse struct {
```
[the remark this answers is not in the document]
rainkwan: I changed this, but now the code is all mix matched with `plugin-go` still being [UpgradeResourceIdentityResponse](https://github.com/hashicorp/terraform-plugin-go/blob/7df6e42af9849860e368350ffcd71d285bdb8950/tfprotov5/internal/fromproto/resource.go#L39)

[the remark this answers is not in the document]
austinvalle: I think we're mismatched in what I originally was asking for.

Anything that is in the exported `resource` package can be named with just the `UpgradeIdentity` prefix, the reason being that we already have the resource package naming, provider developers are using this in the context of a resource already, etc.

For everywhere internal, the proto5server requests, provider server implementation, etc... that can all stay with `UpgradeResourceIdentity` prefix as it was before.

resource/resource.go
```diff
@@ -223,3 +223,18 @@ type ResourceWithIdentity interface {
...
+	// A mapping of the prior identity version to current identity upgrade
+	// implementation. Only the specified identity upgrader for the prior identity
+	// version is called, rather than each version in between, so it must
+	// encapsulate all logic to convert the prior identity to the current identity schema
+	// version.
+	//
+	// Version keys begin at 0, which is the default schema version when
+	// undefined. The framework will return an error diagnostic should the
+	// requested identity version not be implemented.
+	UpgradeResourceIdentity(context.Context) map[int64]IdentityUpgrader
```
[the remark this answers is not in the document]
austinvalle: Looks like this and the interface above are still missing a rename: 
```suggestion
	UpgradeIdentity(context.Context) map[int64]IdentityUpgrader
```

resource/resource.go
```diff
@@ -216,10 +216,25 @@ type ResourceWithValidateConfig interface {
...
-//     is an upgrade of the identity data after a schema change, via the ResourceWithUpgradeIdentity interface.
+//     is an upgrade of the identity data after a schema change, via the ResourceWithUpgradeResourceIdentity interface.
 type ResourceWithIdentity interface {
 	Resource
 
 	// IdentitySchema should return the identity schema for this resource.
 	IdentitySchema(context.Context, IdentitySchemaRequest, *IdentitySchemaResponse)
 }
+
+type ResourceWithUpgradeResourceIdentity interface {
```
austinvalle, line 227:
```suggestion
type ResourceWithUpgradeIdentity interface {
```

resource/resource.go
```diff
@@ -216,10 +216,25 @@ type ResourceWithValidateConfig interface {
 //     by the identity object via the ResourceWithImportState interface.
 //   - The resource identity should not change during the lifecycle of the remote object. That is, from the
 //     creation of the remote object in the remote system until its destruction. An exception to this rule
-//     is an upgrade of the identity data after a schema change, via the ResourceWithUpgradeIdentity interface.
+//     is an upgrade of the identity data after a schema change, via the ResourceWithUpgradeResourceIdentity interface.
```
austinvalle, line 219:
```suggestion
//     is an upgrade of the identity data after a schema change, via the ResourceWithUpgradeIdentity interface.
```

resource/identityschema/schema.go
```diff
@@ -29,7 +29,7 @@ type Schema struct {
 
 	// Version indicates the current version of the resource identity schema. Resource
 	// identity schema versioning enables identity state upgrades in conjunction with the
-	// [resource.ResourceWithUpgradeIdentity] interface. Versioning is only
+	// [resource.ResourceWithUpgradeResourceIdentity] interface. Versioning is only
```
austinvalle, line 32:
```suggestion
	// [resource.ResourceWithUpgradeIdentity] interface. Versioning is only
```

resource/identity_upgrader.go
```diff
@@ -0,0 +1,39 @@
...
+	// identity version to the current schema version.
+	//
+	// The context.Context parameter contains framework-defined loggers and
+	// supports request cancellation.
+	//
+	// The UpgradeResourceIdentityRequest parameter contains the prior identity data.
+	// If PriorSchema was set, the Identity field will be available. Otherwise,
+	// the RawIdentity must be used.
+	//
+	// The UpgradeResourceIdentityResponse parameter should contain the upgraded
```
austinvalle, line 36:
```suggestion
	// The UpgradeIdentityResponse parameter should contain the upgraded
```

resource/identity_upgrader.go
```diff
@@ -0,0 +1,39 @@
...
+	// UpgradeResourceIdentityRequest type RawIdentity field.
+	PriorSchema *identityschema.Schema
+
+	// Provider defined logic for upgrading a resource identity from the prior
+	// identity version to the current schema version.
+	//
+	// The context.Context parameter contains framework-defined loggers and
+	// supports request cancellation.
+	//
+	// The UpgradeResourceIdentityRequest parameter contains the prior identity data.
```
austinvalle, line 32:
```suggestion
	// The UpgradeIdentityRequest parameter contains the prior identity data.
```

resource/identity_upgrader.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+// Implementation handler for an UpgradeResourceIdentity operation.
+//
+// This is used to encapsulate all upgrade logic from a prior identity to the
+// current version when a Resource implements the
+// ResourceWithUpgradeResourceIdentity interface.
+type IdentityUpgrader struct {
+	// Schema information for the prior identity version. While not required,
+	// setting this will populate the UpgradeResourceIdentityRequest type Identity
```
austinvalle, line 18:
```suggestion
	// setting this will populate the UpgradeIdentityRequest type Identity
```

resource/identity_upgrader.go
```diff
@@ -0,0 +1,39 @@
...
+// SPDX-License-Identifier: MPL-2.0
+
+package resource
+
+import (
+	"context"
+	"github.com/hashicorp/terraform-plugin-framework/resource/identityschema"
+)
+
+// Implementation handler for an UpgradeResourceIdentity operation.
```
austinvalle, line 11:
```suggestion
// Implementation handler for an UpgradeIdentity operation.
```

austinvalle (approved):
🚀

PR #1137, opened 2025-05-09
ResourceIdentity: Validate that identities do not change after Terraform stores it

austinvalle:
This PR re-introduces the validation that previously lived in Terraform core (removed in https://github.com/hashicorp/terraform/pull/36989), which ensures that resource identities do not change after Terraform core stores them in state.

The one tricky part of this PR is `ReadResource`, which the protocol currently does not have sufficient information for us to determine if the state we are refreshing has already been stored (typical use-case) or if we are importing the resource. Rather than loosening how strict the validation is, I added a temporary key to a framework reserved `private` field, which Terraform passes between `ImportResourceState -> ReadResource`. Once we've read that field, we can skip validation of the identity, clear the `private` field, then following refreshes will validate as normal.

As we have the immutable validation in the SDK now, we can allow resources (such as RDS instances) to easily opt-out of that validation when necessary.

There are some corner tests over here which display the immutable error messaging: https://github.com/hashicorp/terraform-provider-corner/pull/340

### DIFF
--- a/internal/fwserver/server_planresourcechange_test.go
+++ b/internal/fwserver/server_planresourcechange_test.go
@@ -4077,7 +4077,7 @@ func TestServerPlanResourceChange(t *testing.T) {
 						"Unexpected Identity Change",
 						"During the planning operation, the Terraform Provider unexpectedly returned a different identity then the previously stored one.\n\n"+
 							"This is always a problem with the provider and should be reported to the provider developer.\n\n"+
-							"Prior Identity: tftypes.Object[\"test_id\":tftypes.String]<\"test_id\":tftypes.String<\"id-123\">>\n"+
+							"Prior Identity: tftypes.Object[\"test_id\":tftypes.String]<\"test_id\":tftypes.String<\"id-123\">>\n\n"+
 							"Planned Identity: tftypes.Object[\"test_id\":tftypes.String]<\"test_id\":tftypes.String<\"new-id-123\">>",
 					),
 				},
@@ -6728,7 +6728,7 @@ func TestServerPlanResourceChange(t *testing.T) {
 						"Unexpected Identity Change",
 						"During the planning operation, the Terraform Provider unexpectedly returned a different identity then the previously stored one.\n\n"+
 							"This is always a problem with the provider and should be reported to the provider developer.\n\n"+
-							"Prior Identity: tftypes.Object[\"test_id\":tftypes.String]<\"test_id\":tftypes.String<\"id-123\">>\n"+
+							"Prior Identity: tftypes.Object[\"test_id\":tftypes.String]<\"test_id\":tftypes.String<\"id-123\">>\n\n"+
 							"Planned Identity: tftypes.Object[\"test_id\":tftypes.String]<\"test_id\":tftypes.String<\"new-id-123\">>",
 					),
 				},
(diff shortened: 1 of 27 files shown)


rainkwan (approved):
Makes sense to handle the problem of changing identities this way and have the SDK help out with validating the identity/or not when we need to.

internal/fwserver/server_updateresource.go
```diff
@@ -172,14 +174,29 @@ func (s *Server) UpdateResource(ctx context.Context, req *UpdateResourceRequest,
...
+				"An unexpected error was encountered when creating the apply response. New identity data was returned by the provider update operation, but the resource does not indicate identity support.\n\n"+
+					"This is always a problem with the provider and should be reported to the provider developer.",
+			)
 
-		return
+			return
+		}
+
+		// If we already have an identity stored, validate that the new identity hasn't changing
+		if !req.ResourceBehavior.MutableIdentity && !req.PlannedIdentity.Raw.IsNull() && !req.PlannedIdentity.Raw.Equal(resp.NewIdentity.Raw) {
```
ansgarm, line 189: Not sure if I'm missing something, but in case of creation (prior identity == nil) this should not error, right? Seems like that case is missing?

austinvalle: Correct! That code path is actually handled by this method `(Server).UpdateResource` not being called at all, since there is some basic checking in the `(Server).ApplyResourceChange` method that would route a null prior state to `(Server).CreateResource`: https://github.com/hashicorp/terraform-plugin-framework/blob/e1d72e4112536aad26a23964555f747e09964ce6/internal/fwserver/server_applyresourcechange.go#L41-L122

ansgarm: Ah, perfect!

ansgarm (approved):

internal/fwserver/server_planresourcechange_test.go
```diff
@@ -3534,66 +3533,7 @@ func TestServerPlanResourceChange(t *testing.T) {
...
-				ResourceSchema: testSchema,
-				Resource: &testprovider.ResourceWithIdentityAndModifyPlan{
-					ModifyPlanMethod: func(ctx context.Context, req resource.ModifyPlanRequest, resp *resource.ModifyPlanResponse) {
-						var data testIdentitySchemaData
-						resp.Diagnostics.Append(req.Identity.Get(ctx, &data)...)
-
-						data.TestID = types.StringValue("new-id-123")
-
-						resp.Diagnostics.Append(req.Identity.Set(ctx, &data)...)
+						resp.Diagnostics.Append(resp.Identity.Set(ctx, &data)...)
```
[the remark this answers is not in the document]
austinvalle: Oh wow, great catch. I completely forgot about this little quirk. The reason I had to change this was because I originally fat fingered this test, however, this should have never actually passed, because the request data should not be the same as the response data (it should be a copy).

I forgot to actually fix this though (I think I originally brought this over similar to plan, which gets copied during the `stateToPlan` function)

Solved this in 90ad13b2715351ce3a411f200f6e1dbfa914c4b9 ⚡

ansgarm (approved):

austinvalle:
> ... I don't know if we want to add a note in the private state documentation about the reserved key, since it's highly unlikely that someone would accidentally use that key name. Maybe if we do something like this again, we can add a section about reserved key names.

We're actually already covered because the framework already validates that the `.` prefix is off-limits 😌 
https://developer.hashicorp.com/terraform/plugin/framework/resources/private-state#reserved-keys

PR #1151, opened 2025-05-28
fwserver: add list resources to `GetMetadata`

bbasata:
## Description

This pull request adds list resource metadata to the protocol-independent framework server implementation of `GetMetadata`.

## Rollback Plan

- [x] If a change needs to be reverted, we will roll out an update to the code within 7 days.

## Changes to Security Controls

No

PR #1152, opened 2025-05-28
fwserver: add list resources to `GetProviderSchema`

bbasata:
## Description

To merge after https://github.com/hashicorp/terraform-plugin-framework/pull/1151. This pull request adds list resource metadata to the protocol-independent framework server implementation of `GetProviderSchema`.

For a focused start, I have defined `list.Schema.StringAttribute` and no other attribute types or block types. Then, the remaining schema definition can follow in another focused pull request.

## Rollback Plan

None


internal/fwserver/server_listresources.go
```diff
@@ -98,3 +99,31 @@ func (s *Server) ListResourceMetadatas(ctx context.Context) ([]ListResourceMetad
...
+
+		logging.FrameworkTrace(ctx, "Calling provider defined ListResource Schemas", map[string]interface{}{logging.KeyListResourceType: typeName})
+		listResource.ListResourceConfigSchema(ctx, schemaReq, &schemaResp)
+		logging.FrameworkTrace(ctx, "Called provider defined ListResource Schemas", map[string]interface{}{logging.KeyListResourceType: typeName})
+
+		diags.Append(schemaResp.Diagnostics...)
+		if schemaResp.Diagnostics.HasError() {
+			continue
+		}
+
```
bbasata, line 124: TODO: validation

internal/fwserver/server_listresources.go
```diff
@@ -0,0 +1,135 @@
...
+	return resourceMetadatas, diags
+}
+
+// ListResourceSchemas returns a map of ListResource Schemas for the
+// GetProviderSchema RPC without caching since not all schemas are guaranteed to
+// be necessary for later provider operations. The schema implementations are
+// also validated.
+func (s *Server) ListResourceSchemas(ctx context.Context) (map[string]fwschema.Schema, diag.Diagnostics) {
+	listResourceSchemas := make(map[string]fwschema.Schema)
+	listResourceFuncs, diags := s.ListResourceFuncs(ctx)
```
austinvalle, line 109: This will give us the list resource implementations actually defined in `ListResources()`, but couldn't there also be implementations in `Resources()` as well? Potentially conflicting 👀 

> same thought for metadata

internal/fwserver/server_listresources.go
```diff
@@ -0,0 +1,135 @@
...
+			s.listResourceFuncsDiags.AddError(
+				"Duplicate ListResource Type Defined",
+				fmt.Sprintf("The %s ListResource type name was returned for multiple list resources. ", typeName)+
+					"ListResource type names must be unique. "+
+					"This is always an issue with the provider and should be reported to the provider developers.",
+			)
+			continue
+		}
+
+		if _, ok := s.resourceFuncs[typeName]; !ok {
```
austinvalle, line 72: Is `s.resourceFuncs` guaranteed to be populated at this point?

internal/fwserver/server_listresources.go
austinvalle, line 109: This will give us the list resource implementations actually defined in `ListResources()`, but couldn't there also be implementations in `Resources()` as well? Potentially conflicting 👀 

> same thought for metadata

list/schema/schema.go
```diff
@@ -3,5 +3,195 @@
...
+
+	// Version indicates the current version of the resource schema. Resource
+	// schema versioning enables state upgrades in conjunction with the
+	// [resource.ResourceWithStateUpgrades] interface. Versioning is only
+	// required if there is a breaking change involving existing state data,
+	// such as changing an attribute or block type in a manner that is
+	// incompatible with the Terraform type.
+	//
+	// Versions are conventionally only incremented by one each release.
+	Version int64
```
austinvalle, line 71: Are these the final attribute fields or are they carryover from resource/schema? Like I'm not sure what `Version` would be used for if there is no state representing this schema.

list/schema/schema.go
```diff
@@ -3,5 +3,195 @@
...
+func (s Schema) ValidateImplementation(ctx context.Context) diag.Diagnostics {
+	var diags diag.Diagnostics
+
+	for attributeName, attribute := range s.GetAttributes() {
+		req := fwschema.ValidateImplementationRequest{
+			Name: attributeName,
+			Path: path.Root(attributeName),
+		}
+
+		diags.Append(fwschema.IsReservedResourceAttributeName(req.Name, req.Path)...)
```
austinvalle, line 160: Just a reminder (since it's also relevant for state storage), we should double check if there are any additional fields that should be protect from shadowing that aren't the same as managed resources.

For example, `state_store` blocks will only need to ensure there is no `provider` attribute in the schema.

austinvalle (approved):
LGTM as a first go, left some considerations for future PRs

internal/fwserver/server_listresources.go
bbasata: My current thinking on this is: Framework only needs to look at `ListResources()`.

If a provider type implements both `resource.Resource` and `list.ListResource`, then they add a `NewMyResource` function in two places: `Resources()` and `ListResources()`.

Example: https://github.com/hashicorp/terraform-provider-corner/pull/358/files#diff-8e71dd37ab963ab4edd62bcbc12cc75b0d4cee4e084f515bf239385c4d16b7e2

There's room for improvement...

internal/fwserver/server_listresources.go
bbasata: Maybe we change this to `s.ResourceFuncs[typeName]` ...

PR #1157, opened 2025-06-05
Add ListResource RPC

bbasata:
## Description

This change adds the ListResource RPC.

This is ~80% ready, so I feel okay moving it out of draft mode. I plan to make a round of code review comments, as a guided walkthrough.

## Rollback Plan

None


internal/fwserver/server_listresources.go
```diff
@@ -87,17 +121,32 @@ func (s *Server) ListResourceFuncs(ctx context.Context) (map[string]func() list.
 // ListResourceMetadatas returns a slice of ListResourceMetadata for the GetMetadata
 // RPC.
 func (s *Server) ListResourceMetadatas(ctx context.Context) ([]ListResourceMetadata, diag.Diagnostics) {
-	resourceFuncs, diags := s.ListResourceFuncs(ctx)
+	listResourceFuncs, diags := s.ListResourceFuncs(ctx)
```
bbasata, line 124: 😃

internal/fwserver/server_listresources.go
```diff
@@ -87,17 +121,32 @@ func (s *Server) ListResourceFuncs(ctx context.Context) (map[string]func() list.
...
+		listResourceMetadatas = append(listResourceMetadatas, ListResourceMetadata{
 			TypeName: typeName,
 		})
 	}
 
-	return resourceMetadatas, diags
+	return listResourceMetadatas, diags
+}
+
+func (s *Server) ListResourceSchema(ctx context.Context, typeName string) (fwschema.Schema, diag.Diagnostics) {
```
bbasata, line 137: This method needs a "proper" re-do with locking etc, see `Server.ResourceSchema`.

internal/proto5server/server_validatelistresourceconfig.go
```diff
@@ -0,0 +1,14 @@
...
+
+package proto5server
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-go/tfprotov5"
+)
+
+func (s *Server) ValidateListResourceConfig(ctx context.Context, request *tfprotov5.ValidateListResourceConfigRequest) (*tfprotov5.ValidateListResourceConfigResponse, error) {
```
bbasata, line 12: This is a placeholder to satisfy the interface. Not implementing the real thing in this PR.

internal/proto6server/server_validatelistresourceconfig.go
```diff
@@ -0,0 +1,14 @@
...
+
+package proto6server
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-go/tfprotov6"
+)
+
+func (s *Server) ValidateListResourceConfig(ctx context.Context, request *tfprotov6.ValidateListResourceConfigRequest) (*tfprotov6.ValidateListResourceConfigResponse, error) {
```
bbasata, line 12: This is a placeholder to satisfy the interface. Not implementing the real thing in this PR.

internal/toproto5/list_resource_result.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-framework/internal/fwserver"
+	"github.com/hashicorp/terraform-plugin-go/tfprotov5"
+)
+
+func ListResourceResult(ctx context.Context, result *fwserver.ListResult) tfprotov5.ListResourceResult {
+	diags := result.Diagnostics
```
bbasata, line 14:
```suggestion
	allDiags := result.Diagnostics
```

internal/toproto5/list_resource_result.go
```diff
@@ -0,0 +1,51 @@
...
+	if diags.HasError() {
+		return tfprotov5.ListResourceResult{
+			Diagnostics: Diagnostics(ctx, diags),
+		}
+	}
+
+	resourceIdentity, d := ResourceIdentity(ctx, result.Identity)
```
bbasata, line 21:
```suggestion
	resourceIdentity, diags := ResourceIdentity(ctx, result.Identity)
```

list/list_resource.go
```diff
@@ -97,6 +98,9 @@ type ListRequest struct {
 	// IncludeResource indicates whether the provider should populate the
 	// [ListResult.Resource] field.
 	IncludeResource bool
+
+	ResourceSchema         fwschema.Schema
+	ResourceIdentitySchema fwschema.Schema
```
austinvalle, line 103: Interesting... I'm not really sure what to do about this. We've never just returned the schema in the request like this before, they are typically used to pre-populate a response object and in the case of them potentially not existing (like during upgrade), we just don't populate the response object.

Here, the results stream won't be pre-populated, but provider developers are required to pass along the same schema/identity schema with each result field. Maybe that implementation detail can be placed somewhere in framework?

list/list_resource.go
```diff
@@ -113,6 +117,9 @@ type ListResultsStream struct {
 	// To indicate a fatal processing error, push a [ListResult] that contains
 	// a [diag.ErrorDiagnostic].
 	Results iter.Seq[ListResult]
+
+	// Diagnostics report errors or warnings related to the list operation.
```
austinvalle, line 121: Are these like global errors? When would a provider developer want to use this field vs. the diagnostics in `ListResult`

list/list_resource.go
austinvalle, line 103: Seperate note, if we do return these schemas, should we return the actual exported schema types? From `identityschema` and `resource/schema` (not sure we can, cycle import?)

tfsdk/resource.go
```diff
@@ -31,23 +31,39 @@ func (c Resource) Get(ctx context.Context, target interface{}) diag.Diagnostics
...
 // from matching, the parent path is returned rather than no match to prevent
 // false positives.
-func (c Resource) PathMatches(ctx context.Context, pathExpr path.Expression) (path.Paths, diag.Diagnostics) {
-	return c.data().PathMatches(ctx, pathExpr)
+func (r Resource) PathMatches(ctx context.Context, pathExpr path.Expression) (path.Paths, diag.Diagnostics) {
+	return r.data().PathMatches(ctx, pathExpr)
 }
 
-func (c Resource) data() fwschemadata.Data {
+// Set populates the entire identity using the supplied Go value. The value `val`
```
austinvalle, line 47:
```suggestion
// Set populates the entire resource using the supplied Go value. The value `val`
```

list/tosdk.go
austinvalle, line 1: This is probably the area we'd want the most feedback on from provider developers right? One thought might be to store the identity/resource schemas on `ListResultsStream` rather then on the request (also do they need to be exported? 🤔 )

Would a variant of this for multiple list results also we worth doing?

austinvalle (approved):
Overall this LGTM, I left some general comments and thoughts

list/list_resource.go
bbasata: Good catch – this is obsolete & removed in an upcoming commit.

list/list_resource.go
bbasata: Yes, I consider this good-enough for this PR, while still under construction. When we look at handling non-Framework schema types, I think that adds new design information that will be useful.

bbasata: To be determined, when we get to the point of handling non-Framework schema types.

On June 17, I'm thinking it's possible that `fwschema.Schema` is the interface to rule them all. Related 💭: https://github.com/hashicorp/terraform-plugin-framework/pull/1165

go.mod
```diff
@@ -6,7 +6,7 @@ toolchain go1.23.7
 
 require (
 	github.com/google/go-cmp v0.7.0
-	github.com/hashicorp/terraform-plugin-go v0.28.0
+	github.com/hashicorp/terraform-plugin-go v0.28.1-0.20250612194609-92d41e5e5c0a
```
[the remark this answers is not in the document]
bbasata: Updated minutes ago in https://github.com/hashicorp/terraform-plugin-framework/pull/1157/commits/9590b11bb23a04f2e90523015a8f3d4fee653c34 ✅

internal/fwserver/server_listresource.go
```diff
@@ -0,0 +1,144 @@
...
+	Diagnostics diag.Diagnostics
+}
+
+// ListResource implements the framework server ListResource RPC.
+func (s *Server) ListResource(ctx context.Context, fwReq *ListRequest, fwStream *ListResultsStream) error {
+	listResource := fwReq.ListResource
+
+	if fwReq.Config == nil {
+		return errors.New("Invalid ListResource request: Config cannot be nil")
+	}
```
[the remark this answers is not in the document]
bbasata: There's no functional reason. I was on the fence re: `error` or diagnostics. I learned that Terraform can work with either one, so `error` struck me as simpler for a request-level nil check.

Now that we have a 👍 on "push a single event with an error diagnostic," we could totally use diagnostics instead. I'm open to that. Any preference?

internal/fwserver/server_listresource.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// processListResult validates the content of a list.ListResult and returns a
+// ListResult
+func processListResult(req list.ListRequest, result list.ListResult) ListResult {
+	if result.Diagnostics.HasError() {
+		return ListResult(result)
+	}
+
+	if result.Identity == nil { // TODO: is result.Identity.Raw.IsNull() a practical concern?
```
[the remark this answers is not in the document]
bbasata: > does Terraform core care if the provider returns a `null` identity? I would say we probably shouldn't return a `null` identity unless core says that's acceptable.

cc: @dbanck @dsa0x for input

internal/fwserver/server_listresource_test.go
```diff
@@ -0,0 +1,257 @@
...
+
+	testIdentityValue2 := tftypes.NewValue(testIdentityType, map[string]tftypes.Value{
+		"test_id": tftypes.NewValue(tftypes.String, "new-id-456"),
+	})
+
+	testCases := map[string]struct {
+		server               *fwserver.Server
+		request              *fwserver.ListRequest
+		expectedStreamEvents []fwserver.ListResult
+	}{
```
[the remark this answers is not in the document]
bbasata: Added in https://github.com/hashicorp/terraform-plugin-framework/pull/1157/commits/7e41955d4fcc6b7357b8558b2cf2f9877d75524a ✨

list/tosdk.go
bbasata: ✅ https://github.com/hashicorp/terraform-plugin-framework/pull/1157#discussion_r2154895086

bbasata:
Ready for re-review and auto-merge.

TODO items for follow-up:

> Perhaps future PR, but if this is a warning that practitioners are likely to see often, maybe we should tailor it specifically to them (i.e. not using the Go types since they have no reference of what these fields mean and won't be reporting to provider developers)

> // TODO: is result.Resource.Raw.IsNull() a practical concern? / does Terraform core care if the provider returns a null identity? I would say we probably shouldn't return a null identity unless core says that's acceptable.

> Now that we have a 👍 on "push a single event with an error diagnostic," we could totally use diagnostics instead. I'm open to that. Any preference?

> Interesting... I'm not really sure what to do about this. We've never just returned the schema in the request like this before, they are typically used to pre-populate a response object and in the case of them potentially not existing (like during upgrade), we just don't populate the response object.

> Here, the results stream won't be pre-populated, but provider developers are required to pass along the same schema/identity schema with each result field. Maybe that implementation detail can be placed somewhere in framework?

> Yes, I consider this good-enough for this PR, while still under construction. When we look at handling non-Framework schema types, I think that adds new design information that will be useful.

internal/fwserver/server_listresource.go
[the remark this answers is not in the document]
austinvalle: Most of framework (like the type system for example) uses diagnostics, which are just a summary/detail, so I'd say we should stick to diagnostics unless we can't for whatever reason 👍🏻

austinvalle (approved):
LGTM 🚀

PR #1161, opened 2025-06-12
all: Adjust `UseStateForUnknown` plan modifiers to preserve known null state values

### LINKED ISSUE #1117: `UseStateForUnknown` doesn't preserve known `null` prior state values
### Module version
```
v1.14.1 (latest)
```

```go
"computed_attr": schema.StringAttribute{
	Computed: true,
	},
},
```

### Expected Behavior
If the prior state value for `computed_attr` is `null`, that should be preserved and stored in the plan (if the attribute is unconfigured) because `null` is a known value.

What I believe the current logic is attempting to do, is prevent this plan modifier from running when the resource is creating (i.e. the entire state is null), which it is achieving, but a little too aggressively since it checks the specific attribute's state value instead: https://github.com/hashicorp/terraform-plugin-framework/blob/15bac5c1758c2f94b904e9a1003cb426a5269830/resource/schema/stringplanmodifier/use_state_for_unknown.go#L39-L42

We should consider updating this logic to:
```go
func (m useStateForUnknownModifier) PlanModifyString(ctx context.Context, req planmodifier.StringRequest, resp *planmodifier.StringResponse) {
	// NOTE: This is the bug fix, do nothing if we are creating
	if req.State.Raw.IsNull() {
		return
	}

	// Do nothing if there is a known planned value.
	if !req.PlanValue.IsUnknown() {
		return
	}

	// Do nothing if there is an unknown configuration value, otherwise interpolation gets messed up.
	if req.ConfigValue.IsUnknown() {
		return
	}

	resp.PlanValue = req.StateValue
}
```

### Actual Behavior
If the prior state value for `computed_attr` is `null`, an unknown value is being planned during update.

austinvalle:
## Related Issue

Closes #1117 

## Description

This PR adjusts all of the `UseStateForUnknown` plan modifiers to also preserve null state values (which are technically known). The original logic was excluding all null state values although we believe the intended behavior was to skip when creating resources (which should be checking the resource instance state for null, not the specific state value).

### Note
One of the impacts of fixing this bug would be any attributes in existing resource implementations that are relying on null values being set to unknown during update plans, which would now start to preserve the prior state (keeping null), so if the provider is not preserving those prior state values during apply, it could cause data consistency issues if that unknown marking was being relied on.

Off the top of my head, I'm not sure why you'd use `UseStateForUnknown` in this way, since the goal of this plan modifier has always been to denote an attribute that doesn't change during apply on updates. Fixing this bug doesn't change that goal, it just makes it more precise.

## Rollback Plan

No

### DIFF
--- a/internal/fwserver/attribute_plan_modification_test.go
+++ b/internal/fwserver/attribute_plan_modification_test.go
@@ -2949,6 +2949,44 @@ func TestAttributeModifyPlan(t *testing.T) {
 						},
 					},
 				),
+				State: tfsdk.State{
+					Raw: tftypes.NewValue(
+						tftypes.Object{
+							AttributeTypes: map[string]tftypes.Type{
+								"test": tftypes.Map{
+									ElementType: tftypes.Object{
+										AttributeTypes: map[string]tftypes.Type{
+											"nested_computed": tftypes.String,
+										},
+									},
+								},
+							},
+						},
+						map[string]tftypes.Value{
+							"test": tftypes.NewValue(
+								tftypes.Map{
+									ElementType: tftypes.Object{
+										AttributeTypes: map[string]tftypes.Type{
+											"nested_computed": tftypes.String,
+										},
+									},
+								},
+								map[string]tftypes.Value{
+									"key1": tftypes.NewValue(
+										tftypes.Object{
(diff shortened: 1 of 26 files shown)


bbasata (approved):
:shipit: :shipit: :shipit:

PR #1171, opened 2025-06-26
list: populate and return `ListResourceSchemas` in `GetProviderSchema` for protov5 and protov6

stephybun:
## Description

Populates and return `ListResourceSchemas` in the `GetProviderSchemaResponse` method for the `GetProviderSchema` RPC for both protov5 and protov6. 

Performed some alphabetical reorderings in the expected objects as well..

## Rollback Plan

None


austinvalle (approved):
Oh nice catch! Not sure how we got so far without actually returning the schemas in the protocol 😅 

On this same note, the `GetMetadata` RPC implementations have the same problem (fwserver is implemented, but `toproto` isn't actually sending anything back)

https://github.com/hashicorp/terraform-plugin-framework/blob/099347e41721a774e128f1107d1d8470cc92df5d/internal/toproto5/getmetadata.go#L29-L43

PR #1180, opened 2025-07-09
list: add dynamic attribute type

stephybun:
## Description

Follow up PR for #1177 adding `DynamicAttribute`

## Rollback Plan

Nerp


austinvalle (dismissed):
Just one suggestion! 🚀

austinvalle (approved):
🚀

PR #1181, opened 2025-07-09
action: Initial package implementation and new RPC support

austinvalle:
## Related Issue

N/A

## Description

This PR implements a skeleton of the new provider server RPCs defined in https://github.com/hashicorp/terraform-plugin-go/pull/534 .

All the implementation currently does is map enough information needed for unlinked resources (to be implemented in a future PR). We also ignore deferred action/client capabilities for now, since the interfaces they will eventually be used with don't exist yet. 

This PR also implements some plumbing necessary to retrieve/store schemas and actions on the framework provider server itself, which will be fully fleshed out and tested once a future PR hooks up `GetProviderSchema` and `GetMetadata`.

-----

Using the `TODO:Actions` tag to keep up with future unit testing and decisions to be made.

## Rollback Plan

No

### DIFF
--- a/action/action.go
+++ b/action/action.go
@@ -1,0 +1,16 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package action
+
+import "context"
+
+type Action interface {
+	// Schema should return the schema for this action.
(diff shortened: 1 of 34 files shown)


bbasata (approved):
:shipit: :shipit: :shipit:

PR #1183, opened 2025-07-11
action: Add initial schema interface, attributes, and unlinked schema

austinvalle:
## Related Issue

N/A

## Description

This PR hooks up the `GetProviderSchema` and `GetMetadata` RPCs and implements the unlinked action schema (along with a couple attributes).

### Notes
- I only implemented a couple attributes, essentially enough to test the full schema (two primitives, one collection, one nested attribute, one nested block). The rest can come in a follow-up PR
- In the `proto5server` and `proto6server` tests, I deleted a ton of duplicate tests. Those tests don't add a ton of value on their own and the schema ones in particular are annoying to update 😆 (perhaps more cleanup/consolidation can come in the future)

## Rollback Plan

No


SBGoods (approved):
LGTM! I'm very much in favor on cutting down on the amount of unit tests in the framework

PR #1185, opened 2025-07-15
action: Implement planning and invoking of actions

austinvalle:
## Related Issue

N/A

## Description

This PR introduces the implementation of the `PlanAction` and `InvokeAction` RPCs. Currently planning actions has one hook `ModifyPlan` (the same as `ModifyPlan` in managed resources, so I named it as such), and invoking has one hook `Invoke` (which always ends with a completed event)

The actual streaming mechanism of the protocol is hidden here since the provider always needs to respond with a single completed event, with multiple (optional) progress events in-between. To keep the interface of `Invoke` simple, we are just providing a function (`InvokeResponse.SendProgress`) for providers to call whenever they want to stream a progress event back to Terraform. Once the invoke method has finished, and the provider has set the proper data on the response, then framework automatically sends the completed event and returns.

## Rollback Plan

No


SBGoods (approved):
LGTM!

PR #1186, opened 2025-07-15
action: Add the remaining attributes and nested blocks to the schema package

austinvalle:
## Related Issue

N/A

## Description

Follow-up to #1183, this PR implements all of the remaining attributes (primitive, collection, structural, nested) and nested blocks to the `action/schema` package.

## Rollback Plan

No

### DIFF
--- a/action/schema/set_attribute_test.go
+++ b/action/schema/set_attribute_test.go
@@ -228,12 +228,12 @@ func TestSetAttributeGetType(t *testing.T) {
 			attribute: schema.SetAttribute{ElementType: types.StringType},
 			expected:  types.SetType{ElemType: types.StringType},
 		},
-		// "custom-type": {
-		// 	attribute: schema.SetAttribute{
-		// 		CustomType: testtypes.SetType{},
-		// 	},
-		// 	expected: testtypes.SetType{},
-		// },
+		"custom-type": {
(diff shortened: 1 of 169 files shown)


action/schema/set_attribute_test.go
```diff
@@ -0,0 +1,532 @@
...
+		"base": {
+			attribute: schema.SetAttribute{ElementType: types.StringType},
+			expected:  types.SetType{ElemType: types.StringType},
+		},
+		// "custom-type": {
+		// 	attribute: schema.SetAttribute{
+		// 		CustomType: testtypes.SetType{},
+		// 	},
+		// 	expected: testtypes.SetType{},
+		// },
```
rainkwan, line 236: Just leaving this for just in case later I pessume?

rainkwan (approved):
Looks reasonable to me

action/schema/set_attribute_test.go
austinvalle: Ah, good catch, that was not on purpose 😆 

We fixed these tests a while ago (as originally written, they produce a weird diff in the test), but must not have fixed them in all schemas. I just went ahead and fixed all the schemas that still had this test commented out in 1503be57c52e11f92efe5af767d3a06a629246c0 ⚡

PR #1188, opened 2025-07-17
action: Add standard validation methods/interfaces for `ValidateActionConfig` RPCs

austinvalle:
## Related Issue

N/A

## Description

This PR introduces the standard methods / interfaces / schema fields to allow actions to define offline validation that will run during `terraform validate`. All of these are used in the implementation of the new `ValidateActionConfig` RPC that was introduced in https://github.com/hashicorp/terraform-plugin-go/pull/539.

There may be some future enhancements made to allow actions to validate linked resources as well, but we'll cross that bridge when it materializes 👻 .

## Rollback Plan

No


action/config_validator.go
```diff
@@ -0,0 +1,30 @@
...
+	// descriptions by external tooling.
+	MarkdownDescription(context.Context) string
+
+	// ValidateAction performs the validation.
+	//
+	// This method name is separate from the datasource.ConfigValidator
+	// interface ValidateDataSource method name, provider.ConfigValidator
+	// interface ValidateProvider method name, ephemeral.ConfigValidator
+	// interface ValidateEphemeralResource method name, and resource.ConfigValidator
+	// interface ValidateResource method name to allow generic validators.
```
stephybun, line 28: Super minor, probably unnecessary and may be subject to change so this does not need to be resolved now, but since we're listing all other existing `ValidateSomething` methods should this comment include `ValidateListResourceConfig` for the `list.ConfigValidator`

stephybun (approved):
LGTM 🍇

action/config_validator.go
austinvalle: Good catch, honestly not really worth keeping comments like this up-to-date, so I'll just make it more generic like the `list` package does. Fixed all the existing ones as well in f90b4f12fc0bb899fbf6e52c5868661a5f036306 ⚡

PR #1193, opened 2025-07-24
Disallow null resource identities

bbasata:
## Related Issue

Fixes https://github.com/hashicorp/terraform-plugin-framework/issues/1184

## Description

🚧 work in progress 🚧 

## Rollback Plan

None.

### DIFF
--- a/internal/fwserver/server_updateresource_test.go
+++ b/internal/fwserver/server_updateresource_test.go
@@ -927,6 +927,58 @@ func TestServerUpdateResource(t *testing.T) {
 				Private: testEmptyPrivate,
 			},
 		},
+		"response-new-identity-null": {
+			server: &fwserver.Server{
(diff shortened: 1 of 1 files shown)


austinvalle (commented):
Code LGTM, as you mentioned, should probably point to the GH ref of `terraform-plugin-go` after that is merged.

Also looks like one of the corner resource's needs to be updated to [actually set identity](https://github.com/hashicorp/terraform-provider-corner/blob/72a8f88c9f729b5e227134d6b34a3f539b095fc3/internal/framework5provider/move_state_resource.go#L66-L76), a good catch! 👍🏻  (Also looks like there is an [extra log statement](https://github.com/hashicorp/terraform-provider-corner/blob/72a8f88c9f729b5e227134d6b34a3f539b095fc3/internal/framework5provider/provider.go#L152) in there 😆 )

PR #1198, opened 2025-08-05
list: framework changes to support list with sdkv2 resources

stephybun:
## Description

Inspired by

https://github.com/hashicorp/terraform-plugin-framework/pull/1176
https://github.com/hashicorp/terraform-plugin-mux/pull/316
https://github.com/hashicorp/terraform-plugin-framework/tree/av/linked-resources

### DIFF
--- a/internal/fwserver/server_getmetadata_test.go
+++ b/internal/fwserver/server_getmetadata_test.go
@@ -839,7 +839,7 @@ func TestServerGetMetadata(t *testing.T) {
 					diag.NewErrorDiagnostic(
 						"ListResource Type Defined without a Matching Managed Resource Type",
 						"The test_resource_1 ListResource type name was returned, but no matching managed Resource type was defined. "+
-							"If the matching managed Resource type not a framework resource either ProtoV5Schema and ProtoV5IdentitySchema must be specified in the RawV5Schemas method, "+
+							"If the matching managed Resource type is not a framework resource either ProtoV5Schema and ProtoV5IdentitySchema must be specified in the RawV5Schemas method, "+
 							"or ProtoV6Schema and ProtoV6IdentitySchema must be specified in the RawV6Schemas method. "+
 							"This is always an issue with the provider and should be reported to the provider developers.",
 					),
--- a/internal/fwserver/server_listresources.go
+++ b/internal/fwserver/server_listresources.go
@@ -101,7 +101,7 @@ func (s *Server) ListResourceFuncs(ctx context.Context) (map[string]func() list.
 				s.listResourceFuncsDiags.AddError(
 					"ListResource Type Defined without a Matching Managed Resource Type",
 					fmt.Sprintf("The %s ListResource type name was returned, but no matching managed Resource type was defined. ", typeName)+
-						"If the matching managed Resource type not a framework resource either ProtoV5Schema and ProtoV5IdentitySchema must be specified in the RawV5Schemas method, "+
+						"If the matching managed Resource type is not a framework resource either ProtoV5Schema and ProtoV5IdentitySchema must be specified in the RawV5Schemas method, "+
 						"or ProtoV6Schema and ProtoV6IdentitySchema must be specified in the RawV6Schemas method. "+
 						"This is always an issue with the provider and should be reported to the provider developers.",
 				)
--- a/list/list_resource.go
+++ b/list/list_resource.go
@@ -45,7 +45,7 @@ type ListResource interface {
 
 // ListResourceWithRawV5Schemas is an interface type that extends ListResource to include a method
 // which allows provider developers to supply the ProtoV5 representations of resource and resource identity
-// schemas. This is necessary if list functionality is being used with a legacy resource.
+// schemas. This is necessary if list functionality is being used with a resource that is not defined with Framework.
 type ListResourceWithRawV5Schemas interface {
 	ListResource
 
@@ -54,12 +54,12 @@ type ListResourceWithRawV5Schemas interface {
 }
 
 // ListResourceWithRawV6Schemas is an interface type that extends ListResource to include a method
-// which allows provider developers to supply the ProtoV5 representations of resource and resource identity
-// schemas. This is necessary if list functionality is being used with a legacy resource.
+// which allows provider developers to supply the ProtoV6 representations of resource and resource identity
+// schemas. This is necessary if list functionality is being used with a resource that is not defined with Framework.
 type ListResourceWithRawV6Schemas interface {
 	ListResource
 
-	// RawV6Schemas is called to provide the ProtoV5 representations of the resource and resource identity schemas.
+	// RawV6Schemas is called to provide the ProtoV6 representations of the resource and resource identity schemas.
 	RawV6Schemas(context.Context, RawV6SchemaRequest, *RawV6SchemaResponse)
 }
 
@@ -209,7 +209,7 @@ type ListResult struct {
 // to the ListResource type RawV5Schemas method.
 type RawV5SchemaRequest struct{}
 
-// RawV5SchemaResponse represents a response that is populated by the Schemas method
+// RawV5SchemaResponse represents a response that is populated by the RawV5Schemas method
 // and is used to pass along the ProtoV5 representations of the resource and resource identity schemas.
 type RawV5SchemaResponse struct {
 	// ProtoV5IdentitySchema is the ProtoV5 representation of the resource identity
@@ -224,19 +224,19 @@ type RawV5SchemaResponse struct {
 }
 
 // RawV6SchemaRequest represents a request for the ListResource to return the
-// ProtoV5 schemas. An instance of this request struct is supplied as an argument
-// to the ListResource type RawV5Schemas method.
+// ProtoV6 schemas. An instance of this request struct is supplied as an argument
+// to the ListResource type RawV6Schemas method.
 type RawV6SchemaRequest struct{}
 
-// RawV6SchemaResponse represents a response that is populated by the Schemas method
-// and is used to pass along the ProtoV5 representations of the resource and resource identity schemas.
+// RawV6SchemaResponse represents a response that is populated by the RawV6Schemas method
+// and is used to pass along the ProtoV6 representations of the resource and resource identity schemas.
 type RawV6SchemaResponse struct {
-	// ProtoV5IdentitySchema is the ProtoV5 representation of the resource identity
+	// ProtoV6IdentitySchema is the ProtoV6 representation of the resource identity
 	// schema. This should only be supplied if framework functionality is being used
 	// with a legacy resource. Currently, this only applies to list.
 	ProtoV6IdentitySchema *tfprotov6.ResourceIdentitySchema
 
-	// ProtoV5Schema is the ProtoV5 representation of the resource schema
+	// ProtoV6Schema is the ProtoV6 representation of the resource schema
 	// This should only be supplied if framework functionality is being used
 	// with a legacy resource. Currently, this only applies to list.
 	ProtoV6Schema *tfprotov6.Schema


internal/fwschemadata/data_set.go
```diff
@@ -10,11 +10,31 @@ import (
...
+	var diags diag.Diagnostics
+
+	if v, ok := val.(tftypes.Value); ok {
+		objType := d.Schema.Type().TerraformType(ctx)
+
+		if !objType.Equal(v.Type()) {
+			diags.AddError(
+				d.Description.Title()+" Write Error",
+				"An unexpected error was encountered trying to write the "+d.Description.String()+". This is always an error in the provider. Please report the following to the provider developer:\n\n"+
+					fmt.Sprintf("Error: Type mismatch between provided value and type of %s", d.Description.String()),
```
austinvalle, line 28: Might be useful to print out the expected type (`objType`) and the type we got (`v.Type()`) to make that easier to debug

internal/fwschemadata/data_set_at_path.go
```diff
@@ -28,9 +28,52 @@ import (
...
+		}
+
+		attrType := atPath.GetType().TerraformType(ctx)
+
+		if !attrType.Equal(v.Type()) {
+			diags.AddAttributeError(
+				path,
+				d.Description.Title()+" Write Error",
+				"An unexpected error was encountered trying to write the "+d.Description.String()+". This is always an error in the provider. Please report the following to the provider developer:\n\n"+
+					fmt.Sprintf("Error: Type of provided value does not match type of %s", path.String()),
```
austinvalle, line 49: Same comment as `data_set`, we should consider adding the expected + actual types to the diagnostic

internal/fwschemadata/data_set_at_path.go
```diff
@@ -28,9 +28,52 @@ import (
 // Lists can only have the next element added according to the current length.
 func (d *Data) SetAtPath(ctx context.Context, path path.Path, val interface{}) diag.Diagnostics {
 	var diags diag.Diagnostics
-
 	ctx = logging.FrameworkWithAttributePath(ctx, path.String())
 
+	if v, ok := val.(tftypes.Value); ok {
```
austinvalle, line 33: I'm not sure the exact usage of this will encounter this problem, but one thing missing from this is validating that the value can be used with the actual type in the framework schema. That is logic used by custom types though so if the developer is always building the schema from a different SDK, it'll never have a custom type.

Perhaps a future thing we might run into if someone, for whatever reason, tries to use this reflection code with a real framework schema with a custom type + `tftypes.Value` (they should just be using `attr.Value` in that case)

-----

TLDR: it's probably fine that we skip validation

internal/fwschemadata/data_set_at_path_test.go
```diff
@@ -2924,6 +2924,39 @@ func TestDataSetAtPath(t *testing.T) {
 				"other": tftypes.NewValue(tftypes.DynamicPseudoType, nil),
 			}),
 		},
+		"write-tftypes-value": {
```
austinvalle, line 2927: We should add a test for the error scenario of the `tftypes.Value` being the wrong type (like a number targeting a string)

internal/fwschemadata/data_set_test.go
```diff
@@ -139,6 +139,39 @@ func TestDataSet(t *testing.T) {
 				),
 			}),
 		},
+		"write-tftypes-values": {
```
austinvalle, line 142: Same as above, we should add a test for the error scenario of the `tftypes.Value` being the wrong type (like an object missing one of the attributes)

list/list_resource.go
```diff
@@ -41,6 +42,16 @@ type ListResource interface {
 	List(context.Context, ListRequest, *ListResultsStream)
 }
 
+// ListResourceWithProtoSchemas is an interface type that extends ListResource to include a method
+// which allows provider developers to supply the ProtoV5 representations of resource and resource identity
+// schemas. This is necessary if list functionality is being used with a legacy resource.
```
austinvalle, line 47: What about protocol v6? It is technically possible with this design to have a resource defined in `terraform-plugin-go` that would be with protocol v6.

austinvalle (commented):
Good stuff @stephybun! Left some initial thoughts

internal/fwschemadata/data_set.go
stephybun: I umm'd and ahh'd over whether to print out the expected and received types for `Set`, particularly for schemas that have a lot of attributes + nesting I find the output of that practically unreadable/unusable unless it's manually manipulated into a more human readable format. 

But also maybe it's just worth providing the info upfront and side-step having to retrieve it some other way..

internal/fwserver/server_listresource.go
```diff
@@ -156,15 +156,17 @@ func (s *Server) ListResource(ctx context.Context, fwReq *ListRequest, fwStream
...
 	}
 
-	fwStream.Results = processListResults(req, stream.Results, diagsStream.Results)
+	if stream.Results == nil {
+		fwStream.Results = processListResults(req, list.NoListResults, diagsStream.Results)
+	}
+
+	if stream.Results != nil {
+		fwStream.Results = processListResults(req, stream.Results, diagsStream.Results)
+	}
```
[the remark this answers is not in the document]
stephybun: Given the way diagnostics are handled for a stream both checks are actually needed to support the scenario where no list results are returned but a warning diagnostic has been set, as well as all other cases. But I agree this looks weird and is not clear in what its purpose is. I added a test case for this scenario and also updated these conditions.

stephybun:
Test coverage might be a bit lacking, but I'll look into this in a separate PR

list/list_resource.go
```diff
@@ -41,6 +43,26 @@ type ListResource interface {
...
+type ListResourceWithRawV5Schemas interface {
+	ListResource
+
+	// RawV5Schemas is called to provide the ProtoV5 representations of the resource and resource identity schemas.
+	RawV5Schemas(context.Context, RawV5SchemaRequest, *RawV5SchemaResponse)
+}
+
+// ListResourceWithRawV6Schemas is an interface type that extends ListResource to include a method
+// which allows provider developers to supply the ProtoV5 representations of resource and resource identity
```
austinvalle, line 57:
```suggestion
// which allows provider developers to supply the ProtoV6 representations of resource and resource identity
```

list/list_resource.go
```diff
@@ -41,6 +43,26 @@ type ListResource interface {
...
+// schemas. This is necessary if list functionality is being used with a legacy resource.
+type ListResourceWithRawV6Schemas interface {
+	ListResource
+
+	// RawV6Schemas is called to provide the ProtoV5 representations of the resource and resource identity schemas.
```
austinvalle, line 62:
```suggestion
	// RawV6Schemas is called to provide the ProtoV6 representations of the resource and resource identity schemas.
```

list/list_resource.go
```diff
@@ -182,6 +204,44 @@ type ListResult struct {
...
+
+// RawV6SchemaRequest represents a request for the ListResource to return the
+// ProtoV5 schemas. An instance of this request struct is supplied as an argument
+// to the ListResource type RawV5Schemas method.
+type RawV6SchemaRequest struct{}
+
+// RawV6SchemaResponse represents a response that is populated by the Schemas method
+// and is used to pass along the ProtoV5 representations of the resource and resource identity schemas.
+type RawV6SchemaResponse struct {
+	// ProtoV5IdentitySchema is the ProtoV5 representation of the resource identity
```
austinvalle, line 234:
```suggestion
	// ProtoV6IdentitySchema is the ProtoV6 representation of the resource identity
```

list/list_resource.go
```diff
@@ -182,6 +204,44 @@ type ListResult struct {
...
+	// schema. This should only be supplied if framework functionality is being used
+	// with a legacy resource. Currently, this only applies to list.
+	ProtoV6IdentitySchema *tfprotov6.ResourceIdentitySchema
+
+	// ProtoV5Schema is the ProtoV5 representation of the resource schema
```
austinvalle, line 239:
```suggestion
	// ProtoV6Schema is the ProtoV6 representation of the resource schema
```

internal/fwserver/server_getmetadata_test.go
```diff
@@ -839,6 +839,8 @@ func TestServerGetMetadata(t *testing.T) {
 					diag.NewErrorDiagnostic(
 						"ListResource Type Defined without a Matching Managed Resource Type",
 						"The test_resource_1 ListResource type name was returned, but no matching managed Resource type was defined. "+
+							"If the matching managed Resource type not a framework resource either ProtoV5Schema and ProtoV5IdentitySchema must be specified in the RawV5Schemas method, "+
```
austinvalle, line 842:
```suggestion
							"If the matching managed Resource type is not a framework resource either ProtoV5Schema and ProtoV5IdentitySchema must be specified in the RawV5Schemas method, "+
```

internal/fwserver/server_listresources.go
```diff
@@ -85,14 +85,28 @@ func (s *Server) ListResourceFuncs(ctx context.Context) (map[string]func() list.
...
-				"ListResource Type Defined without a Matching Managed Resource Type",
-				fmt.Sprintf("The %s ListResource type name was returned, but no matching managed Resource type was defined. ", typeName)+
-					"This is always an issue with the provider and should be reported to the provider developers.",
-			)
-			continue
+			if (rawV5SchemasResp.ProtoV5Schema == nil || rawV5SchemasResp.ProtoV5IdentitySchema == nil) && (rawV6SchemasResp.ProtoV6Schema == nil || rawV6SchemasResp.ProtoV6IdentitySchema == nil) {
+				s.listResourceFuncsDiags.AddError(
+					"ListResource Type Defined without a Matching Managed Resource Type",
+					fmt.Sprintf("The %s ListResource type name was returned, but no matching managed Resource type was defined. ", typeName)+
+						"If the matching managed Resource type not a framework resource either ProtoV5Schema and ProtoV5IdentitySchema must be specified in the RawV5Schemas method, "+
```
austinvalle, line 104:
```suggestion
						"If the matching managed Resource type is not a framework resource either ProtoV5Schema and ProtoV5IdentitySchema must be specified in the RawV5Schemas method, "+
```

list/list_resource.go
```diff
@@ -41,6 +43,26 @@ type ListResource interface {
 	List(context.Context, ListRequest, *ListResultsStream)
 }
 
+// ListResourceWithRawV5Schemas is an interface type that extends ListResource to include a method
+// which allows provider developers to supply the ProtoV5 representations of resource and resource identity
+// schemas. This is necessary if list functionality is being used with a legacy resource.
```
austinvalle, line 48: nit: Should we be more specific about what a legacy resource is? Or is that something we think provider developers will understand by reading this  

Perhaps just "a resource that is not defined in Framework"?🤔

list/list_resource.go
```diff
@@ -182,6 +204,44 @@ type ListResult struct {
 	Diagnostics diag.Diagnostics
 }
 
+// RawV5SchemaRequest represents a request for the ListResource to return the
+// ProtoV5 schemas. An instance of this request struct is supplied as an argument
+// to the ListResource type RawV5Schemas method.
+type RawV5SchemaRequest struct{}
+
+// RawV5SchemaResponse represents a response that is populated by the Schemas method
```
austinvalle, line 212:
```suggestion
// RawV5SchemaResponse represents a response that is populated by the RawV5Schemas method
```

list/list_resource.go
```diff
@@ -182,6 +204,44 @@ type ListResult struct {
...
+
+	// ProtoV5Schema is the ProtoV5 representation of the resource schema
+	// This should only be supplied if framework functionality is being used
+	// with a legacy resource. Currently, this only applies to list.
+	ProtoV5Schema *tfprotov5.Schema
+}
+
+// RawV6SchemaRequest represents a request for the ListResource to return the
+// ProtoV5 schemas. An instance of this request struct is supplied as an argument
+// to the ListResource type RawV5Schemas method.
```
austinvalle, line 228:
```suggestion
// ProtoV6 schemas. An instance of this request struct is supplied as an argument
// to the ListResource type RawV6Schemas method.
```

austinvalle (approved):
LGTM 🚀  - all of the comments are pkg doc related / nits and can be addressed later if preferred

list/list_resource.go
stephybun: The latter suggestion is clearer 👍

PR #1199, opened 2025-08-06
actions: Implement lifecycle actions and linked resources

austinvalle:
## Related Issue

N/A

## Description

This PR implements the lifecycle action type and linked resources, which can be used by both lifecycle actions and linked actions (not yet implemented). I also wrote out most of the package docs that had TODOs on them, and left a new TODO to implement validation on the new lifecycle action type (this PR was getting too long 😆).

Any thoughts on the error messaging, documentation, etc. would be appreciated!

## Rollback Plan

No


internal/fwserver/server_planaction.go
```diff
@@ -94,5 +145,24 @@ func (s *Server) PlanAction(ctx context.Context, req *PlanActionRequest, resp *P
...
+						len(modifyPlanResp.LinkedResources),
+						len(resp.LinkedResources),
+					)+
+					"This is always a problem with the provider and should be reported to the provider developer.",
+			)
+			return
+		}
+
+		for i, newLinkedResource := range modifyPlanResp.LinkedResources {
+			resp.LinkedResources[i].PlannedState = planToState(newLinkedResource.Plan)
```
austinvalle, line 164: This applies to `InvokeAction` as well, but I'm thinking we might want to consider also:
- Running write-only nullification? (probably both)
- Checking mutability of identities and if they change?
- Running semantic equality logic (probably just in Invoke actually)

action/invoke.go
```diff
@@ -14,7 +14,33 @@ type InvokeRequest struct {
...
+	//   - If the action schema type is Linked, this field will be one or more linked resources, which
+	//     will be in the same order as the linked resource schemas are defined in the action schema.
+	//
+	// For Lifecycle actions, the provider may only change computed-only attributes of the linked resources.
+	// For Linked actions, the provider may change any attributes of the linked resources.
+	LinkedResources []InvokeRequestLinkedResource
+}
+
+// InvokeRequestLinkedResource represents linked resource data before the action is invoked.
+type InvokeRequestLinkedResource struct {
```
austinvalle, line 30: The only thing not represented in this request that is normally in the plan, is private data.... Should that be in the protocol?

austinvalle:
I've moved this PR to draft for now as we wait for some design confirmations and moved some of the mergable changes over to #1200

PR #1201, opened 2025-08-12
list: add map, list and single nested blocks and attributes

stephybun:
## Description

Adds map, list and single nested blocks and attributes. Deliberately left out support for sets but we could look into adding these if/when there is a specific request for them.

## Rollback Plan

None


austinvalle (approved):
I left a thread comment continuing some of our slack thread, but the code itself all looks good 👍🏻

PR #1202, opened 2025-08-13
list: add and set list resource configure data

stephybun:
## Description

* Extends `ConfigureResponse` to support `ListResourceData`
* Copies `ListResourceData` over in the `ConfigureProvider` RPC call

## Rollback Plan

Nerp


austinvalle (approved):
Changes look good 🚀 , one nit is we should probably add a quick test (can be in a future PR if you prefer)

https://github.com/hashicorp/terraform-plugin-framework/blob/e3cf75ae63ec459e49b960bd68a816c976c4e129/internal/fwserver/server_configureprovider_test.go#L195-L208

PR #1207, opened 2025-08-19
list: set empty objects in `NewListResult`

stephybun:
## Description

- Sets an empty object into `ResourceIdentity` and `Resource` in the `NewListResult` method

## Rollback Plan

None

### DIFF
--- a/internal/proto5server/server_listresource_test.go
+++ b/internal/proto5server/server_listresource_test.go
@@ -105,7 +105,7 @@ func TestServerListResource(t *testing.T) {
 						continue
 					}
 
-					result := req.NewListResult()
+					result := req.NewListResult(ctx)
 					result.DisplayName = name
 
 					diags = result.Identity.Set(ctx, resources[name].ThingResourceIdentity)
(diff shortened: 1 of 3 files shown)


internal/fwschemadata/data_set.go
```diff
@@ -10,11 +10,17 @@ import (
...
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
 )
 
 // Set replaces the entire value. The value should be a struct whose fields
 // have one of the attr.Value types. Each field must have the tfsdk field tag.
 func (d *Data) Set(ctx context.Context, val any) diag.Diagnostics {
+	if v, ok := val.(tftypes.Value); ok {
+		d.TerraformValue = v
+		return nil
+	}
```
stephybun, line 22: We probably want to check that the supplied `val` actually corresponds to the schema of the object we're setting this too?

stephybun: This has been moved over into https://github.com/hashicorp/terraform-plugin-framework/pull/1198 since it's related to that piece of work

austinvalle (approved):

PR #1216, opened 2025-09-06
Implement `UseStateForUnknownIf` plan modifier

### LINKED ISSUE #1214: Plan Modifier `UseStateForUnknownIf`
Sometimes, the value of an attribute depends on whether another attribute's value is updated. It would be nice to have a `UseStateForUnknownIf(...)` function, similar to [`RequiresReplaceIf(...)`](https://developer.hashicorp.com/terraform/plugin/framework/resources/plan-modification#requiresreplaceif), to mark an attribute as unknown during an update operation if the attribute it depends on has been changed; otherwise, marking it as known.

```go
func (r *MyResource) Schema(ctx context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
	resp.Schema = schema.Schema{
			"a": schema.StringAttribute{
				Required: true,
			},
			"b_depending_on_a": schema.StringAttribute{
				Optional: true,
					stringplanmodifier.UseStateForUnknownIf(
						func(ctx context.Context, sr planmodifier.StringRequest, rrifr *stringplanmodifier.UseStateForUnknownIfFuncResponse) {
							// ...
						},
					),,
				},
			},
		},
	}
}
```

ChihweiLHBird:
## Related Issue

Fixes #1214

## Description

The `UseStateForUnknownIf` plan modifier can copy the state value into plan when the condition is met.

There are some common cases to use this plan modifier, for example:
1. A computed attribute is only changed during an update operation when another attribute is being changed.
2. ...


ChihweiLHBird:
Hi @rainkwan, any chance we may get this change in? I think it will be useful for many downstream provider work. Let me know if you have any feedback.

PR #1223, opened 2025-09-17
Add context setup to `proto{5,6}server` `ListResource()`

bbasata:
## Description

I noticed something was missing when I ran `TF_LOG=trace terraform query` today:
```
2025-09-16T14:23:38.070-0400 [TRACE] provider.terraform-provider-cloudlite: Checking ResourceTypes lock: @caller=/Users/baraa.basata/go/pkg/mod
/github.com/hashicorp/terraform-plugin-framework@v1.16.0-beta.1.0.20250911064719-2607d454c868/internal/logging/framework.go:29 @module=sdk.framework new_logger_warning="This log was generated by an SDK subsystem logger that wasn't created before being used. Use tflog.NewSubsystem to create this logger before it is used." timestamp=2025-09-16T14:23:36.190-0400
```

This pull request adds common Go context setup to the Framework implementations of `ListResource()` and `ValidateListResourceConfig()`. These two lines result in more correct behavior re: context cancellation and logging.

## Rollback Plan

None.

### DIFF
--- a/internal/proto5server/server_listresource.go
+++ b/internal/proto5server/server_listresource.go
@@ -9,6 +9,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/internal/fromproto5"
 	"github.com/hashicorp/terraform-plugin-framework/internal/fwserver"
+	"github.com/hashicorp/terraform-plugin-framework/internal/logging"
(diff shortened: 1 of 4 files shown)


austinvalle (approved):
LGTM, nice catch! 🚀 CI is related to us not having a TF core version that supports the recent protocol changes we made, will be resolved when bumping 1.14 from alpha to beta 👍🏻

PR #1227, opened 2025-09-18
Allow `RequiresReplace` plan modifier to run for write-only attributes

### LINKED ISSUE #1226: Write only PlanModifier RequiresReplace is never executed
### Module version

```
github.com/hashicorp/terraform-plugin-framework v1.16.0
```

```go
    resp.Schema = schema.Schema{
			"secret_string": schema.StringAttribute{
				Optional:      true,
				WriteOnly:     true,
				PlanModifiers: []planmodifier.String{stringplanmodifier.RequiresReplace()},
			},
		},
	}
```

```hcl
terraform {
  required_providers {
    provs = {
      source = "edu/provs"
    }
  }
}

provider "provs" {
  path = "/var/tmp/custom_tf_provider"
}

resource "provs_secret" "test" {
  description = "test secret"
  secret_string = "val"
}
```

### Expected Behavior
To satisfy this specification from the [docs](https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/write-only-arguments):
Image

### Actual Behavior
When `PlanModifiers` are specified for write only attributes, even if there is a configuration for the attribute, the replace of the resource does not happen which contradicts with the public docs.

#### Provider setup
1. Clone the testing provider from [here](https://github.com/yottta/terraform-provider-provs/tree/requiresreplace).
2. Run `go run ./... -debug=true`
3. Save the returned TF_REATTACH_PROVIDERS to use it later

#### Configuration setup
With the configuration shared above, do the following:

1. Run `TF_REATTACH_PROVIDERS='...' terraform apply` (to create the resource)
2. Run `TF_REATTACH_PROVIDERS='...' terraform apply` (to see if the replace is executed which it is not)

yottta:
## Related Issue

Fixes #1226

## Description

Since the request to plan modifiers contains no information about the attribute configuration, I added the WriteOnly attribute one into each type request struct, later used in `requiresReplaceIfModifier` to check if the modifier should be executed or not.
Note that a write only attribute can trigger a replacement only when the value of that particular field is present into the configuration.

## Rollback Plan

- [ ] If a change needs to be reverted, we will roll out an update to the code within 7 days.

Are there any changes to security controls (access controls, encryption, logging) in this pull request? If so, explain: **No**

### DIFF
--- a/internal/fwserver/attribute_plan_modification.go
+++ b/internal/fwserver/attribute_plan_modification.go
@@ -786,6 +786,7 @@ func AttributePlanModifyBool(ctx context.Context, attribute fwxschema.AttributeW
 		Private:        req.Private,
 		State:          req.State,
 		StateValue:     stateValue,
+		WriteOnly:      attribute.IsWriteOnly(),
 	}
 
 	for _, planModifier := range attribute.BoolPlanModifiers() {
@@ -946,6 +947,7 @@ func AttributePlanModifyFloat32(ctx context.Context, attribute fwxschema.Attribu
 		Private:        req.Private,
 		State:          req.State,
 		StateValue:     stateValue,
+		WriteOnly:      attribute.IsWriteOnly(),
 	}
 
 	for _, planModifier := range attribute.Float32PlanModifiers() {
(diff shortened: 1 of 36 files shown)


yottta:
Not needed. See details in #1226

PR #1229, opened 2025-09-22
identity: Update change validation to ensure empty identities (all null attributes) are not validated

### LINKED ISSUE #1228: identity: Allow identities with all null attributes to change
### Module version
```
v1.16.0
```

### Context

Similar to an SDKv2 issue that was raised and fixed: https://github.com/hashicorp/terraform-plugin-sdk/pull/1527

While we want to prevent identities from being stored with all null values or null (#1184), it currently is possible for this to be in state via older versions of Terraform / SDKv2 / Framework (or custom terraform-plugin-go implementations). An identity with all null attributes is equivalent to an identity that is null, so we should prevent change validation from running in that scenario (the latter we already ignore during change validation).

Adjacent but different to: https://github.com/hashicorp/terraform-plugin-framework/pull/1193

austinvalle:
## Related Issue

Closes #1228
Similar fix to https://github.com/hashicorp/terraform-plugin-sdk/pull/1513

## Description

There are some scenarios where an identity may be stored with all null values (a scenario that #1193 will eventually prevent from occurring), which should not be validated, since an identity with all null values is invalid.

The only real possibility of this happening ATM is with:
- A migrated resource from SDKv2 which was storing all null attributes inadvertently 
- A framework resource that was storing all null attributes explicitly
- A migrated resource from `terraform-plugin-go` which was storing all null attributes (there is no validation in the lower-level SDK)

## Rollback Plan

No

### DIFF
--- a/internal/fwserver/server_applyresourcechange_test.go
+++ b/internal/fwserver/server_applyresourcechange_test.go
@@ -59,6 +59,24 @@ func TestServerApplyResourceChange(t *testing.T) {
 		},
 	}
 
+	type testMultiIdentitySchemaData struct {
+		TestAttrA types.String `tfsdk:"test_attr_a"`
+		TestAttrB types.Int64  `tfsdk:"test_attr_b"`
+	}
+
+	testMultiAttrIdentitySchema := identityschema.Schema{
+		Attributes: map[string]identityschema.Attribute{
+			"test_attr_a": identityschema.StringAttribute{
+				RequiredForImport: true,
+			},
+			"test_attr_b": identityschema.Int64Attribute{
+				OptionalForImport: true,
+			},
+		},
+	}
+
+	testMultiAttrIdentityType := testMultiAttrIdentitySchema.Type().TerraformType(context.Background())
+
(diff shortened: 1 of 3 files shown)


internal/fwserver/server_readresource.go
```diff
@@ -185,7 +185,7 @@ func (s *Server) ReadResource(ctx context.Context, req *ReadResourceRequest, res
 		}
 
 		// If we're refreshing the resource state (excluding a recently imported resource), validate that the new identity isn't changing
-		if !req.ResourceBehavior.MutableIdentity && !readFollowingImport && !req.CurrentIdentity.Raw.IsNull() && !req.CurrentIdentity.Raw.Equal(resp.NewIdentity.Raw) {
+		if !req.ResourceBehavior.MutableIdentity && !readFollowingImport && !req.CurrentIdentity.Raw.IsFullyNull() && !req.CurrentIdentity.Raw.Equal(resp.NewIdentity.Raw) {
```
rainkwan, line 188: FullyNull is an interesting way to say it

rainkwan (approved):

PR #1230, opened 2025-10-01
list: update list result validation to check if identity is fully null

stephybun:
## Description

Semi-related to validation added in:

https://github.com/hashicorp/terraform-plugin-framework/pull/1193
https://github.com/hashicorp/terraform-plugin-sdk/pull/1513

Framework allows list results with identities containing all null values, meaning providers may inadvertently forget or incorrectly set identity data. This PR changes the validation performed to also check that the values of the list result's identity are not all null.

## Rollback Plan

None

### DIFF
--- a/.changes/unreleased/BUG FIXES-20251001-152300.yaml
+++ b/.changes/unreleased/BUG FIXES-20251001-152300.yaml
@@ -1,0 +1,5 @@
+kind: BUG FIXES
+body: 'fwserver: update validation list result validation to check if an identity''s values are all null'
+time: 2025-10-01T15:23:00.976476+02:00
+custom:
+    Issue: "1230"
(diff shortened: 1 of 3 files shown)


.changes/unreleased/BUG FIXES-20251001-152300.yaml
```diff
@@ -0,0 +1,5 @@
+kind: BUG FIXES
+body: 'fwserver: update validation list result validation to check if an identity''s values are all null'
```
SBGoods, line 2: nit: typo
```suggestion
body: 'fwserver: update validation list result validation to check if an identity's values are all null'
```

SBGoods (approved):
Just one typo in the changelog entry but otherwise, LGTM

.changes/unreleased/BUG FIXES-20251001-152300.yaml
SBGoods: Actually, ignore this suggestion. The extra `'` is acting as an escape character 😅

PR #1235, opened 2025-11-03
Pluggable State Store Initial Read Implementation

rainkwan:
## Description

Continues PSS implementation from https://github.com/hashicorp/terraform-plugin-go/pull/563

## Rollback Plan

No

### DIFF
--- a/internal/fwserver/server_statestore.go
+++ b/internal/fwserver/server_statestore.go
@@ -23,7 +23,7 @@ func (s *Server) StateStore(ctx context.Context, statestoreType string) (statest
 	if !ok {
 		diags.AddError(
 			"StateStore Type Not Found",
-			fmt.Sprintf("No state store type named %q was found in the provider.", statestoreType),
+			fmt.Sprintf("No statestore type named %q was found in the provider.", statestoreType),
 		)
 
 		return nil, diags


internal/fwserver/server.go
```diff
@@ -250,6 +257,29 @@ type Server struct {
...
+	// Provider.Actions() method.
+	statestoreFuncs map[string]func() statestore.StateStore
+
+	// statestoreFuncsDiags is the cached Diagnostics obtained while populating
+	// statestoreFuncs. This is to ensure any warnings or errors are also
+	// returned appropriately when fetching statestoreFuncs.
+	statestoreFuncsDiags diag.Diagnostics
+
+	// statestoreFuncsMutex is a mutex to protect concurrent statestoreFuncs
+	// access from race conditions.
```
austinvalle:
```suggestion
	// statestoreSchemas is the cached StateStore Schemas for RPCs that need to
	// convert configuration data from the protocol. If not found, it will be
	// fetched from the StateStore.Schema() method.
	statestoreSchemas map[string]statestoreschema.Schema

	// statestoreSchemasMutex is a mutex to protect concurrent statestoreSchemas
	// access from race conditions.
	statestoreSchemasMutex sync.RWMutex

	// statestoreFuncs is the cached StateStore functions for RPCs that need to
	// access statestores. If not found, it will be fetched from the
	// Provider.StateStore() method.
	statestoreFuncs map[string]func() statestore.StateStore

	// statestoreFuncsDiags is the cached Diagnostics obtained while populating
	// statestoreFuncs. This is to ensure any warnings or errors are also
	// returned appropriately when fetching statestoreFuncs.
	statestoreFuncsDiags diag.Diagnostics

	// statestoreFuncsMutex is a mutex to protect concurrent statestoreFuncs
	// access from race conditions.
```

internal/fwserver/server_readstatebytes_test.go
```diff
@@ -0,0 +1,135 @@
...
+	for name, testCase := range testCases {
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			response := &fwserver.ReadStateBytesResponse{}
+			testCase.server.ReadStateBytes(context.Background(), testCase.request, response)
+
+			// For now, just verify the response doesn't panic
+			// The actual streaming behavior would need more complex testing
+			_ = response
```
austinvalle: This test doesn't actually run make any assertions on the data 🤔 - The underlying  `(fwserver).ReadStateBytes` is essentially just mapping data between two data structures, so we should compare them at the very least

internal/proto6server/server_readstatebytes.go
```diff
@@ -0,0 +1,119 @@
...
+				return
+			}
+
+			// TODO: Get chunk size from statestore.ConfiguredChunkSize() when available
+			// For now, use 8MB default
+			chunkSize := 8 << 20 // 8 MB
+
+			reader := bytes.NewReader(fwResp.Bytes)
+			totalLength := reader.Size()
+			rangeStart := 0
```
austinvalle: Since all of this logic is synchronous, I think putting it outside of this iterator would make more sense.

internal/proto6server/server_readstatebytes.go
```diff
@@ -0,0 +1,119 @@
...
+
+			for {
+				readBytes := make([]byte, chunkSize)
+				byteCount, err := reader.Read(readBytes)
+
+				if byteCount == 0 {
+					break
+				}
```
austinvalle: We should check the error first, since the default value of an int is `0`

internal/proto6server/server_readstatebytes_test.go
```diff
@@ -0,0 +1,135 @@
+// Copyright (c) HashiCorp, Inc.
```
austinvalle: These tests are a good start! Maybe adding in some varying lengths (would be nice to test the actual chunking, and you need to hook up the configuring to do that), and definitely some diagnostic testing.

Maybe a test where we create an invalid byte stream to see how the diagnostics are handled in the chunk method?

statestore/configure.go
```diff
@@ -0,0 +1,40 @@
...
+package statestore
+
+import (
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
+)
+
+type StateStoreClientCapabilities struct {
+	DeferralAllowed bool
+}
```
austinvalle: Similar note as I'm not sure if this was meant to be implemented in this PR or not, `DeferralAllowed` isn't a valid client capability here, it's just `ChunkSize`

Also, the response will also need a server capability to respond with the determined chunk size (which we can copy over by default from the client capability)

statestore/schema/doc.go
austinvalle: If we're going to introduce every schema attribute in this PR (which I don't think we really need to), I'd recommend comparing them to the protocol RFC: https://docs.google.com/document/d/1lIK1E3mFj3dVya2SG2tVmq8awfO7c6KpvlzssVLchBA/edit?tab=t.0#heading=h.fko2wos7pa15

For example, nested blocks aren't valid and there are no write only fields

austinvalle (commented):
Left some initial feedback! As of right now the provider server still doesn't implement the `tfprotov6.StateServer` interface, so it won't actually work ATM. I also mostly skipped over the changes for RPCs that are not `ReadStateBytes`, so any missing/incorrect mapping or tests we'll need to remember to come back and fix 👍🏻 (type name is never needed in the request since that's just a routing mechanism, I see some requests have provider type name which I'm not sure where that's coming from, some requests/responses have the wrong data in them, etc.) 

Personally, I would be fine with those additional files (GetStates, DeleteState, Validate, Write) just being deleted if we aren't implementing them fully 😃 

-----

I left some comments on the `ReadStateBytes` implementation, but you might want to compare it to the `terraform-plugin-testing` implementation, as the ordering/diagnostic handling is going to result in some subtle bugs: https://github.com/hashicorp/terraform-plugin-testing/blob/17bfda9c447c2e0556bd0cb0c786d153f7213814/internal/testing/testsdk/providerserver/providerserver.go#L1433-L1506


-----

Also side note, the copyright header on these is going to need to match the new ones https://github.com/hashicorp/terraform-plugin-framework/pull/1246/files

statestore/schema/doc.go
austinvalle: There are a number of other things that we still need to add here to support schema/metadata fully, so I just extracted the commit/code out of this PR, made a couple small additions, and opened a new one: https://github.com/hashicorp/terraform-plugin-framework/pull/1259

Once we review/merge that PR and fix the merge conflict here, we should be able to get this PR more focused on it's main goal -> implementing `ReadStateBytes` 👍🏻

austinvalle:
Superseded by #1265

Issue #1211, opened 2025-08-25
v1.15.1 causing computed attrs to throw "produced an unexpected new value" on child attributes of an attribute that was null on create and filled in on update


mawasile:
### Module version

```
1.15.1
```

### Relevant provider source code

https://github.com/microsoft/terraform-provider-power-platform/blob/d33ddd7dead09a245466a7bb456d0c4ee7cd0fcb/internal/services/environment/resource_environment.go#L653
https://github.com/microsoft/terraform-provider-power-platform/blob/d33ddd7dead09a245466a7bb456d0c4ee7cd0fcb/internal/services/environment/models.go#L240

```go
	attrTypesDataverseObject := map[string]attr.Type{
		"url":                          types.StringType,
		"domain":                       types.StringType,
		"organization_id":              types.StringType,
		"security_group_id":            types.StringType,
		"language_code":                types.Int64Type,
		"version":                      types.StringType,
		"linked_app_type":              types.StringType,
		"linked_app_id":                types.StringType,
		"linked_app_url":               types.StringType,
		"currency_code":                types.StringType,
		"templates":                    types.ListType{ElemType: types.StringType},
		"template_metadata":            types.StringType,
		"administration_mode_enabled":  types.BoolType,
		"background_operation_enabled": types.BoolType,
		"unique_name":                  types.StringType,
	}

	attrValuesProductProperties := map[string]attr.Value{}
	model.Dataverse = types.ObjectNull(attrTypesDataverseObject)
```

### Terraform Configuration Files

please look at unit test 
https://github.com/microsoft/terraform-provider-power-platform/blob/d33ddd7dead09a245466a7bb456d0c4ee7cd0fcb/internal/services/environment/resource_environment_test.go#L1920

or run it using from our repository using 
```
make unittest TEST=EnvironmentsResource_Validate_Create_Environment_And_Dataverse
```

### Debug Output

2025-08-25T10:48:46.928Z [ERROR] sdk.helper_resource: Unexpected error: test_step_number=2 test_name=TestUnitEnvironmentsResource_Validate_Create_Environment_And_Dataverse test_terraform_path=/home/runtimeuser/tfenv/bin/terraform test_working_directory=/tmp/plugintest3209803560
  error=
  | Error running apply: exit status 1
  | 
  | Error: Provider produced inconsistent result after apply
  | 
  | When applying changes to powerplatform_environment.development, provider
[...]

PR #1242, opened 2025-11-21
all: Add `UseNonNullStateForUnknown` plan modifier to assist with child nested attribute use-cases

### LINKED ISSUE #1197: Recent update to `UseStateForUnknown` causes errors when adding to nested objects
When we try to upgrade to 1.15.1, our terraform provider has a regression - it now outputs errors (`Provider produced inconsistent result after apply`) when users add a new entry to a nested attribute object. Each nested object has several attributes that use the `UseStateForUnknown` plan modifier because they're either computed fields (like uuid) or optional fields that we'll preserve existing values for if they aren't set in the terraform config.

### Module version

```
v1.15.1
```

```go
func (r *PipelineResource) Schema(ctx context.Context, req resource.SchemaRequest, resp *resource.SchemaResponse) {
	resp.Schema = schema.Schema{
		MarkdownDescription: "Artie Pipeline resource. This represents a pipeline that syncs data from a single source (e.g., Postgres) to a single destination (e.g., Snowflake).",
			"uuid":                        schema.StringAttribute{Computed: true, PlanModifiers: []planmodifier.String{stringplanmodifier.UseStateForUnknown()}},
			"name":                        schema.StringAttribute{Required: true, MarkdownDescription: "The human-readable name of the pipeline. This is used only as a label and can contain any characters."},
			"tables": schema.MapNestedAttribute{
				Required:            true,
				MarkdownDescription: "A map of tables from the source database that you want to replicate to the destination.",
						"uuid":                   schema.StringAttribute{Computed: true, PlanModifiers: []planmodifier.String{stringplanmodifier.UseStateForUnknown()}},
						"name":                   schema.StringAttribute{Required: true, MarkdownDescription: "The name of the table in the source database."},
						"schema":                 schema.StringAttribute{Optional: true, Computed: true, PlanModifiers: []planmodifier.String{stringplanmodifier.UseStateForUnknown()}, MarkdownDescription: "The name of the schema the table belongs to in the source database. This must be specified if your source database uses schemas (such as PostgreSQL), e.g. `public`."},
						"enable_history_mode":    schema.BoolAttribute{Optional: true, Computed: true, Default: booldefault.StaticBool(false), PlanModifiers: []planmodifier.Bool{boolplanmodifier.UseStateForUnknown()}, MarkdownDescription: "If set to true, we will create an additional table in the destination (suffixed with `__history`) to store all changes to the source table over time."},
						"is_partitioned":         schema.BoolAttribute{Optional: true, Computed: true, Default: booldefault.StaticBool(false), PlanModifiers: []planmodifier.Bool{boolplanmodifier.UseStateForUnknown()}, MarkdownDescription: "If the source table is partitioned, set this to true and we will ingest data from all of its partitions."},
						"alias":                  schema.StringAttribute{Optional: true, Computed: true, PlanModifiers: []planmodifier.String{stringplanmodifier.UseStateForUnknown()}, MarkdownDescription: "An optional alias for the table. If set, this will be the name of the destination table."},
						"columns_to_exclude":     schema.ListAttribute{Optional: true, Computed: true, ElementType: types.StringType, PlanModifiers: []planmodifier.List{listplanmodifier.UseStateForUnknown()}, MarkdownDescription: "An optional list of columns to exclude from syncing to the destination."},
						"columns_to_include":     schema.ListAttribute{Optional: true, Computed: true, ElementType: types.StringType, PlanModifiers: []planmodifier.List{listplanmodifier.UseStateForUnknown()}, MarkdownDescription: "An optional list of columns to include in replication."},
						"columns_to_hash":        schema.ListAttribute{Optional: true, Computed: true, ElementType: types.StringType, PlanModifiers: []planmodifier.List{listplanmodifier.UseStateForUnknown()}, MarkdownDescription: "An optional list of columns to hash in the destination."},
						"skip_deletes":           schema.BoolAttribute{Optional: true, Computed: true, PlanModifiers: []planmodifier.Bool{boolplanmodifier.UseStateForUnknown()}, MarkdownDescription: "If set to true, we will skip delete events for this table and only process insert and update events."},
...
```

```hcl
resource "artie_pipeline" "dev_postgres_to_snowflake" {
  name = "Postgres (dev db) to Snowflake"
  tables = {
    "public.account" = {
       name   = "account"
       schema = "public"
    },
    "public.company" = {
       name   = "company"
       schema = "public"
    },
  }
}
```

Adding a new table to the `tables` object is a common use case for our provider (when a customer wants to start replicating an additional database table in their data pipeline). With the above config for an existing pipeline, imagine adding another table:
```hcl
"public.api_key" = {
   name   = "api_key"
   schema = "public"
}
```

### Behavior as of v1.15.0

The plan output looked like this:
```hcl
# artie_pipeline.dev_postgres_to_snowflake will be updated in-place
  ~ resource "artie_pipeline" "dev_postgres_to_snowflake" {
        name                               = "Postgres (dev db) to Snowflake"
      ~ tables                             = {
          + "public.api_key" = {
              + alias                  = (known after apply)
              + columns_to_exclude     = (known after apply)
              + columns_to_hash        = (known after apply)
              + columns_to_include     = (known after apply)
              + enable_history_mode    = false
              + is_partitioned         = false
              + name                   = "api_key"
              + schema                 = "public"
              + skip_deletes           = (known after apply)
              + uuid                   = (known after apply)
            },
            # (3 unchanged elements hidden)
        }
        # (10 unchanged attributes hidden)
    }
```

And the update succeeded with no errors.

### Behavior as of v1.15.1

Now the plan output looks like this:

```hcl
# artie_pipeline.dev_postgres_to_snowflake will be updated in-place
  ~ resource "artie_pipeline" "dev_postgres_to_snowflake" {
        name                               = "Postgres (dev db) to Snowflake"
      ~ tables                             = {
          + "public.api_key" = {
              + enable_history_mode = false
              + is_partitioned      = false
              + name                = "api_key"
              + schema              = "public"
            },
            # (4 unchanged elements hidden)
        }
        # (10 unchanged attributes hidden)
    }
```

And after the update is applied, these errors show up for every field that uses the `UseStateForUnknown` plan modifier and didn't have a value specified in the config:

```hcl
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to artie_pipeline.dev_postgres_to_snowflake, provider "provider[\"registry.terraform.io/artie-labs/artie\"]" produced an unexpected new value: .tables["public.api_key"].uuid: was null, but now cty.StringVal("e84b4aae-516f-477e-8529-546ac8116ea7").
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to artie_pipeline.dev_postgres_to_snowflake, provider "provider[\"registry.terraform.io/artie-labs/artie\"]" produced an unexpected new value: .tables["public.api_key"].alias: was null, but now cty.StringVal("").
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to artie_pipeline.dev_postgres_to_snowflake, provider "provider[\"registry.terraform.io/artie-labs/artie\"]" produced an unexpected new value: .tables["public.api_key"].skip_deletes: was null, but now cty.False.
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to artie_pipeline.dev_postgres_to_snowflake, provider "provider[\"registry.terraform.io/artie-labs/artie\"]" produced an unexpected new value: .tables["public.api_key"].columns_to_hash: was null, but now cty.ListValEmpty(cty.String).
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to artie_pipeline.dev_postgres_to_snowflake, provider "provider[\"registry.terraform.io/artie-labs/artie\"]" produced an unexpected new value: .tables["public.api_key"].columns_to_exclude: was null, but now cty.ListValEmpty(cty.String).
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
╷
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to artie_pipeline.dev_postgres_to_snowflake, provider "provider[\"registry.terraform.io/artie-labs/artie\"]" produced an unexpected new value: .tables["public.api_key"].columns_to_include: was null, but now cty.ListValEmpty(cty.String).
│ 
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

### References
- https://github.com/hashicorp/terraform-plugin-framework/pull/1161
 
This seems to be the cause of the behavior change. The note in the PR description acknowledges that there could be use cases where this would cause a regression, and it looks like this is one 😅 The change correctly handles resources being created, but in this case the resource already exists and a nested object inside of it is what's being created.

Would it make sense to add an option to the `UseStateForUnknown` plan modifiers so that we can configure how they should treat null values?

Thank you for reviewing!

austinvalle:
## Related Issue

Closes #1197
Ref #1211, #1161, #1117 

## Description

This PR introduces a new plan modifier that implements the same logic as pre-1.15.1 `UseStateForUnknown`, so it will not preserve `null` state values, allowing them to stay marked as unknown.

The primary use-case for a plan modifier like this is for use on child-attributes in a nested attribute which will only be preserved once the nested object exists (regardless of whether the resource has been created or not). See https://github.com/hashicorp/terraform-plugin-framework/issues/1197#issuecomment-3559979490 for a detailed description of the rationale.

While this plan modifier solves the problem more broad then needed, it's also the most accessible/understandable solution for nested attributes of requiring this behavior, since detecting if a parent object exists is not an easy operation (see any issue referencing set identity or their planning 😏 ).

---

As there were existing providers that may be expecting pre 1.15.1 behavior on `UseStateForUnknown`, I plan on retroactively updating the 1.15.1 changelog (in this PR) + GH release to reference this behavior change more explicitly.

## Rollback Plan

No

### DIFF
--- a/.changes/1.15.1.md
+++ b/.changes/1.15.1.md
@@ -1,5 +1,9 @@
 ## 1.15.1 (July 31, 2025)
 
+NOTES:
+
+* This release contains a change in behavior for the `UseStateForUnknown` plan modifier, which will now preserve null values from prior state for unconfigured attributes. This updated version can cause plan inconsistency errors when used on child attributes of a nested attribute that expect `UseStateForUnknown` to keep the child attributes on new nested objects as `<unknown>` (known after apply). See [#1197](https://github.com/hashicorp/terraform-plugin-framework/issues/1197) for more details.
+
 BUG FIXES:
 
 * all: Fixed bug with `UseStateForUnknown` where known null state values were not preserved during update plans. ([#1117](https://github.com/hashicorp/terraform-plugin-framework/issues/1117))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,10 @@ BUG FIXES:
 
 ## 1.15.1 (July 31, 2025)
 
+NOTES:
+
+* This release contains a change in behavior for the `UseStateForUnknown` plan modifier, which will now preserve null values from prior state for unconfigured attributes. This updated version can cause plan inconsistency errors when used on child attributes of a nested attribute that expect `UseStateForUnknown` to keep the child attributes on new nested objects as `<unknown>` (known after apply). See [#1197](https://github.com/hashicorp/terraform-plugin-framework/issues/1197) for more details.
+
 BUG FIXES:
 
 * all: Fixed bug with `UseStateForUnknown` where known null state values were not preserved during update plans. ([#1117](https://github.com/hashicorp/terraform-plugin-framework/issues/1117))


.changes/1.15.1.md
austinvalle: Also updated in the GH release (if we want to change the wording I can re-update 👍🏻 )
https://github.com/hashicorp/terraform-plugin-framework/releases/tag/v1.15.1

It felt weird to update the changelog to reference a future release, so I just mentioned the issue 🙂

SBGoods (approved):
LGTM!

PR #1259, opened 2026-01-21
statestore: Add initial RPC skeletons and implement schema/metadata

austinvalle:
## Related Issue

Ref: https://github.com/hashicorp/terraform-plugin-go/pull/563
Pulled from: #1235

## Description

This PR pulled the schema and metadata code from #1235 so it's easier to review each of the pieces individually. Additionally, merging this PR will allow parallel work on the other state store RPCs.

I created the branch from #1235 to retain the original author commits (from @rainkwan ), removed the pieces of code that were unrelated to schema/metadata, then finished the implementation and tests for `GetMetadata`/`GetProviderSchema`. 

Some notes of the differences between state stores and other schemas:
- State stores can only be implemented in protocol v6, so we raise an error if we detect a v5 provider with state stores (to avoid silently hiding them)
- Nested blocks are not allowed (as the provider must be using protocol v5, nested attributes should be used instead)
- `Computed`, `WriteOnly`, and `Sensitive` are not valid options for an attribute.

## Rollback Plan

No

### DIFF
--- a/.changes/unreleased/ENHANCEMENTS-20260127-122835.yaml
+++ b/.changes/unreleased/ENHANCEMENTS-20260127-122835.yaml
@@ -1,0 +1,5 @@
+kind: ENHANCEMENTS
+body: 'types/basetypes: Added `Length()` method for the `ListValue` type.'
+time: 2026-01-27T12:28:35.393455-05:00
+custom:
+    Issue: "1067"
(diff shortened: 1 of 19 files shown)


go.mod
```diff
@@ -2,33 +2,34 @@ module github.com/hashicorp/terraform-plugin-framework
 
 go 1.24.0
 
-toolchain go1.24.4
+toolchain go1.24.10
 
 require (
 	github.com/google/go-cmp v0.7.0
-	github.com/hashicorp/terraform-plugin-go v0.29.0
+	github.com/hashicorp/terraform-plugin-go v0.29.1-0.20260120220013-b2f99cb946ea
```
austinvalle: TODO: Based on https://github.com/hashicorp/terraform-plugin-go/pull/594, update to point at main before merging 👍🏻

PR #1270, opened 2026-02-04
statestore: Implement `Write` for state stores

austinvalle:
## Related Issue

N/A

## Description

This PR implements the `WriteStateBytes` RPC for `statestore.StateStore` as a simple request/response method, hiding the complexities of the underlying streaming details of the protocol. Generally speaking `internal/proto6server/server_writestatebytes.go` does most of the heavy lifting here, which is where you'll find most of the tests as well 👍🏻 .

## Rollback Plan

No

### DIFF
--- a/internal/proto6server/server_writestatebytes.go
+++ b/internal/proto6server/server_writestatebytes.go
@@ -31,7 +31,7 @@ func (s *Server) WriteStateBytes(ctx context.Context, proto6Req *tfprotov6.Write
 	if configuredChunkSize <= 0 {
 		fwResp.Diagnostics.AddError(
 			"Error Writing State",
-			"The provider server does not have a chunk size configured. This is a bug in either Terraform or terraform-plugin-framework and should be reported.",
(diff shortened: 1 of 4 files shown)


internal/proto6server/server_writestatebytes.go
```diff
@@ -0,0 +1,130 @@
...
+			typeName = chunk.Meta.TypeName
+			stateID = chunk.Meta.StateID
+		}
+
+		if chunk.Range.End < chunk.TotalLength-1 {
+			// Ensure each chunk (except the last) exactly match the configured size.
+			if int64(len(chunk.Bytes)) != configuredChunkSize {
+				fwResp.Diagnostics.AddError(
+					"Error Writing State",
+					fmt.Sprintf("Unexpected chunk of size %d was received from Terraform, expected chunk size was %d. This is a bug and should be reported.", len(chunk.Bytes), configuredChunkSize),
```
stephybun: Can we provide guidance on who the bug should be reported to like we do for the error message on line 34?

PR #1292, opened 2026-03-23
generate config new algo

ansgarm:
## What this does

This improves the quality of generated resource config.

It now:
- clears top-level `id` and `timeouts`
- clears computed-only values
- clears empty optional values
- handles validator groups better for `ConflictsWith`, `ExactlyOneOf`, and `AlsoRequires`

For the validator group handling to work, validators need to implement the three new interfaces that expose the paths they validate. The `terraform-plugin-framework-validators` package will be updated to support that, so updating that package will be one way to pick this up.

## Why

The old behavior was what Terraform Core can do on its own.

This change improves the generated config using schema information that the provider has, but Terraform Core does not. That gives us better output during import and makes the generated config closer to something a user can actually keep instead of having to clean up by hand.

## Notes

I kept the normal Framework default behavior instead of writing defaults straight into generated config across the board.

The one special case is `ExactlyOneOf`, where using a default can help generate config that makes sense.

## Rollback Plan

- [ ] If a change needs to be reverted, we will roll out an update to the code within 7 days.

### DIFF
--- a/internal/fwserver/server_generateresourceconfig.go
+++ b/internal/fwserver/server_generateresourceconfig.go
@@ -5,14 +5,16 @@ package fwserver
 
 import (
 	"context"
-	"errors"
+	"sort"
+	"strings"
 
 	"github.com/hashicorp/terraform-plugin-go/tftypes"
 
 	"github.com/hashicorp/terraform-plugin-framework/diag"
-	"github.com/hashicorp/terraform-plugin-framework/internal/fromtftypes"
 	"github.com/hashicorp/terraform-plugin-framework/internal/fwschema"
-	"github.com/hashicorp/terraform-plugin-framework/internal/logging"
+	"github.com/hashicorp/terraform-plugin-framework/internal/fwschema/fwxschema"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
 )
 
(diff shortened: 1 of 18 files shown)


internal/fwserver/server_generateresourceconfig.go
```diff
@@ -55,138 +55,343 @@ func (s *Server) GenerateResourceConfig(ctx context.Context, req *GenerateResour
...
+	tfValue, ok := rawValue.(tftypes.Value)
+
+	return tfValue, ok
 }
 
-// stateToConfig returns a *tfsdk.Config with a copied value from a tfsdk.State.
-func stateToConfig(state tfsdk.State) *tfsdk.Config {
-	return &tfsdk.Config{
-		Raw:    state.Raw.Copy(),
-		Schema: state.Schema,
```
ansgarm, line 190: this was inlined

internal/fwserver/server_generateresourceconfig.go
```diff
@@ -55,138 +55,343 @@ func (s *Server) GenerateResourceConfig(ctx context.Context, req *GenerateResour
...
+	idPath := tftypes.NewAttributePath().WithAttributeName("id")
+
+	// id
+	idAttribute, err := req.State.Schema.AttributeAtTerraformPath(ctx, idPath)
+	if err == nil && idAttribute.IsComputed() && idAttribute.IsOptional() {
+		config = nullValueAtPath(config, idPath)
+	} // else: no id attribute, ignoring
+
+	// timeouts
+	timeoutPath := tftypes.NewAttributePath().WithAttributeName("timeouts")
```
SBGoods, line 78: I think it would be better if steps 1 and 2 were handled inside of the same `tftypes.Transform` func.

internal/fwserver/server_generateresourceconfig.go
```diff
@@ -55,138 +55,343 @@ func (s *Server) GenerateResourceConfig(ctx context.Context, req *GenerateResour
...
+		}
+	}
+
+	return tftypes.Transform(value, func(p *tftypes.AttributePath, v tftypes.Value) (tftypes.Value, error) {
+		if p.Equal(path) {
+			return replaceWith, nil
+		}
+
+		return v, nil
+	})
```
SBGoods, line 396: I don't think these helper methods are necessary, `tfsdk.State` should have a `SetAttribute()` method that you can use to set an attribute value to null at a specific path.

internal/fwserver/validator_groups.go
SBGoods, line 1: This file is very difficult to read/understand conceptually. There are a lot of tiny helper functions that require me to jump around in the code and breaks my reading flow. It might be better to inline the logic of a lot of these functions back into their calling functions.

internal/fwserver/server_generateresourceconfig_test.go
SBGoods, line 1: We should also have test coverage for the different nested block/attribute combinations (list, set, map nested attributes/blocks).

ansgarm:
superseded by #1297 which we're manually fixing :)

